PR #2, opened 2014-07-30
7.0 account chart report

mcassuto:
Add a new simple module to print the chart of accounts.
This feature is often asked by our customers.


account_chart_report_rml/report/chart_of_accounts.py
```diff
@@ -0,0 +1,60 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+from openerp.report import report_sxw
+
+
+class account_char(report_sxw.rml_parse):
```
pedrobaeza: CamelCase for class name

account_chart_report_rml/report/chart_of_accounts.py
```diff
@@ -0,0 +1,60 @@
...
+        super(account_char, self).__init__(cr, uid, name, context=context)
+        self.localcontext.update({
+            "get_lst_account": self._get_lst_account,
+            "cr": cr,
+            "uid": uid,
+            "actual_context": context,
+        })
+
+    def _get_lst_account(self, cr, uid, account_id, context):
+        account_obj = self.pool.get("account.account")
```
pedrobaeza: Better `self.pool[]` instead of `self.pool.get`

account_chart_report_rml/wizard/account_report_chart_of_account.py
```diff
@@ -0,0 +1,47 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import fields, orm
+
+
+class chart_of_account_report(orm.TransientModel):
```
pedrobaeza: CamelCase

pedrobaeza:
Thanks for the module!

I have tried it, and there's two minor things I see to fix:
- The module name, as it was, without _rml, is correct, because RML is the default report engine, so it's redundant. The surname is for reports in another engine, like webkit, and so on. In v8, it would be the appropiate name, because default engine is QWeb.
- The left and right margin are too height, constraining the account names a lot, and making to extend in two lines. Can you please reduce them.

Regards.

mcassuto:
Hello @pedrobaeza ,

first of all sorry for the delay !

I am impressed you took time to test the module, thanks.
- CamelCase fixed
- back to the original name
- self.pool.get replaced
- I fixed the margins; I also agree it is nicer like that ;-)

Regards,
Marc

pedrobaeza:
Hi, Marc, thank you very much for the changes, and don't worry about the delay: you're answering reviews very quick against the average.

One last thing: please remove `es.mo` file from the PR, that it's a subproduct of POEdit.

Regards.

mcassuto:
@pedrobaeza :  es.mo file **is already** removed as part of the last commit:

--------------------- account_financial_report/i18n/es.mo ---------------------
deleted file mode 100644
index eb09639..0000000
Binary files a/account_financial_report/i18n/es.mo and /dev/null differ

pedrobaeza:
Sorry, I see in a bad way (too early in the morning :wink:).

:+1: (code review and test)

bwrsandman:
:+1:

max3903:
:+1:

PR #7, opened 2014-08-14
7.0 account partner aged statement

mcassuto:
# Print & Send Partner Aged Statement by email

This module adds in the system :
- a new mail template "Aged Statement Letter";
- the ability to print the partner aged statement;
- a button "Send by email" in the partner form view which load the template
  and attach the statement to the email.

Comparing to 'Overdue Payment' report provided with OpenERP, this one adds more
informations :
- the summary per period
- the list of invoices per period
- the invoices are also shown with the foreign currency
  
  Both messages, in the PDF and the email are taken from the Overdue Payments
  Message defined on the Company's configuration.


nbessi:
:-1:  Please resubmit with only account_partner_aged_statement change set or change destination branch

nbessi:
Is there a reason to not inherit  account-financial-reporting / account_financial_report_webkit / report / aged_partner_balance.py instead ?

PR #12, opened 2014-08-25
[FIX] Use the migration machinery to initialize default values

lmignon:
backported from https://code.launchpad.net/~acsone-openerp/account-financial-report/7.0-bug-1312732-lmi/+merge/217240

The proposal is valid for the 7.0 since the migration script is no more called on fresh install in 8.0. For 8.0 the migration script can be replaced by an install hook (https://github.com/odoo/odoo/commit/4105b5f0285821d8056c721b5fd50b9cae902988)

### DIFF
--- a/account_financial_report_webkit/account_move_line.py
+++ b/account_financial_report_webkit/account_move_line.py
@@ -29,26 +29,6 @@ class AccountMoveLine(orm.Model):
     account move line"""
     _inherit = 'account.move.line'
 
-    def init(self, cr):
-        # We do not want to catch error as if sql is not run it will give
-        # invalid data
-        cr.execute("UPDATE account_move_line as acm "
-                   " SET last_rec_date ="
-                   "     (SELECT date from account_move_line"
-                   "          WHERE reconcile_id =  acm.reconcile_id"
-                   "              AND reconcile_id IS NOT NULL"
-                   "          ORDER BY date DESC LIMIT 1)"
-                   " WHERE last_rec_date is null;")
-
-        cr.execute("UPDATE account_move_line as acm "
-                   " SET last_rec_date ="
-                   "     (SELECT date from account_move_line"
-                   "          WHERE reconcile_partial_id"
-                   "                             = acm.reconcile_partial_id"
-                   "              AND reconcile_partial_id IS NOT NULL"
-                   "          ORDER BY date DESC LIMIT 1)"
-                   " WHERE last_rec_date is null;")
-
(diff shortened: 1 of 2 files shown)


guewen:
:+1:

nbessi:
:+1:

laetitia-gangloff:
:+1: 

use in our production environment, and it is really good when updating the server !

lepistone:
Hi, thanks for your work. I was about to do the same thing, thanks to @guewen for pointing me here.

About the PR: do we actually need that at all? Being `last_rec_date` a stored function field, shouldn't it be calculated at module installation anyway?

lmignon:
@lepistone  https://bugs.launchpad.net/openobject-server/+bug/1314680/comments/12, an other solution is maybe to improve the stored function... If you want, you can try ;-)

lepistone:
Thanks @lmignon for the link. In fact I could try to improve the query in the function field, but I am left with my initial question: why to we need `init` or migration scripts at all? In theory OpenERP should recompute all stored function fields anyway on module installation.

My proposal is to get rid of `init` without adding any migration script in this PR, and later improve the query as suggested by @KangOl in a new PR. Or am I missing something?

lmignon:
@lepistone The stored function is called by _auto_init. Each upgrade of the module launch the query again and again and the query takes a lot of time. We want to initialize the value when first installed only. (never during an update)

guewen:
@lepistone yes the orm compute the function field when it is created. But if I remember correctly, the `init` was introduced because the computation done by the function field was way too slow, unfortunately I can't find anything in the bzr/git history because the history of the account-financial-report branch is messed up. @nbessi advices could be useful because it was his work.

nbessi:
@lepistone We did not wanted the ORM to do the update work for performance reason.
It was way tooooo slooow on large accounting. But the init approach had a downside the query was run on each update.It is not an optimal behavior. That is why we use the migration functionality in order to have only the query run on addons installation.

pedrobaeza:
I think that with a correct store triggers, we can avoid this, and the calculation is only done when needed. BTW, v8 brings a new way to put init values once, but I don't know which method it is.

lmignon:
This issue is open for a long time, we can perhaps proceed to merge the PR and if anyone has a better solution, it can be done in another PR.

lepistone:
I am still not convinced that this will not recompute the stored field anyway. Still, I agree with @lmignon: this is an improvement over what was there before, and maybe it can be improved further later. :+1: then! Thanks to all!

PR #13, opened 2014-08-29
add account_trial_balance_period_xls

luc-demeyer:
cf. https://github.com/luc-demeyer/account-financial-reporting/blob/7.0/account_trial_balance_period_xls/doc/doc.rst

for functional description.


account_trial_balance_period_xls/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import wizard
+from . import report
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
bwrsandman: Please don't include vim lines

account_trial_balance_period_xls/report/trial_balance_period_xls.py
```diff
@@ -0,0 +1,671 @@
...
+            cr, uid, name, context=context)
+        self.context = context
+        self.localcontext.update({
+            '_': self._,
+        })
+
+    def _(self, src):
+        lang = self.context.get('lang', 'en_US')
+        return translate(
+            self.cr, _ir_translation_name, 'report', lang, src) or src
```
bwrsandman, line 50: What is the purpose of this and the localcontext stuff in __init__(). Could you provide a docstring?

account_trial_balance_period_xls/wizard/trial_balance_period_wizard.py
```diff
@@ -0,0 +1,177 @@
...
+        if period_from.date_start > period_to.date_start:
+            raise orm.except_orm(
+                _('Error'),
+                _('The Start Period must precede the End Period !'))
+        cr.execute(
+            "SELECT id FROM account_period "
+            "WHERE date_start>=%s AND date_stop<=%s "
+            "AND company_id=%s AND NOT special "
+            "ORDER BY date_start",
+            (period_from.date_start, period_to.date_stop, company_id))
```
bwrsandman, line 137: This would be safer with a `self.pool['account.period'].search()`

account_trial_balance_period_xls/report/trial_balance_period_xls.py
luc-demeyer, line 50: I am not sure that a docstring will be more clear than the code : I override the standard "_" method by a more specific one to avoid that the wrong translation terms are returned.

account_trial_balance_period_xls/wizard/trial_balance_period_wizard.py
luc-demeyer, line 137: I do not want to be victim by modules (including the standard OpenERP one) overriding the account.period search method (and doing things with the 'special' flag which I don't want).

account_trial_balance_period_xls/report/trial_balance_period_xls.py
bwrsandman, line 50: It's not what it's doing that would be nice to be in the docstring, but **why** it's overwritten.
As in why does this `_()` definition have to be here and what's wrong with the default `_()`

account_trial_balance_period_xls/wizard/trial_balance_period_wizard.py
bwrsandman, line 137: What's the problem with an overwrite of the search? If it's overwritten it is to blanket change all searches of periods.

This implementation doesn't take into account user group permissions and will also return shallow unlinked (`active = False`) periods. Is this what is indented?

luc-demeyer, line 137: User permissions: if the user has no permission on account.period, than he will already have a problem during the wizard.
active=False: account.period doesn't have an active flag (and shouldn't have one), this is an accounting report, all move lines have to be shown in a period range,

account_trial_balance_period_xls/report/trial_balance_period_xls.py
luc-demeyer, line 50: The default _() returns unreliable translations: 
If you analyse the code,you'll see that it calls:
res = pool.get('ir.translation')._get_source(cr, SUPERUSER_ID, None, ('code','sql_constraint'), lang, source)
the _get_source on its turn just returns a matching term (trad = res and res[0] or u'')

If you want reliable translations whereby you want the report to return the terms that is in the module's i18n file and not defined somewhere else in different context, than an override of the _() is a way to do this.

bwrsandman, line 50: Thank you for the explanation.
It would be nice to have that explained in a docstring.

yvaucher:
Added need fixing for "It would be nice to have that explained in a docstring."

gurneyalex:
could you rebase your PR to get a fresh runbot build?

ovnicraft:
@luc-demeyer will be rebased or can be closed until next update ?

pedrobaeza:
@luc-demeyer are you going to work on this?

pedrobaeza:
Closing as no answer

PR #20, opened 2014-10-10
account_financial_report_webkit: Port to 8.0

anthony-muschang:
I ported the module "account_financial_report_webkit" to 8.0.

Since report_sxw.py use the new API it's not possible anymore to add non-existing fields to a record like in "report/open_invoices.py":

``` python
account.ledger_lines = ledger_lines_memoizer.get(account.id, {})
account.init_balance = init_balance_memoizer.get(account.id, {})
```

So I choice to store those values in the "localcontext" instead.


StefanRijnhart:
Much appreciated! The migration looks very clean, I don't see any unwarranted changes, and the changes that have been made look fine to me, and fit your comment on this PR.

:+1:

pedrobaeza:
Seems OK for me too. :+1:

max3903:
:+1:

PR #23, opened 2014-10-15
account_financial_report_webkit_xls: Port to 8.0

anthony-muschang:
I ported the module "account_financial_report_webkit_xls" to 8.0.

This pull request is based on https://github.com/OCA/reporting-engine/pull/5.

All the modifications in this pull request are the result of the migration of account_financial_report_webkit to v8 (see https://github.com/OCA/account-financial-reporting/pull/20).


pedrobaeza:
You have to fix PEP8 issue with one line length and modify Travis file to download reporting-engine branch for get the dependencies solved.

anthony-muschang:
The pep8 issues are fixed and the travis file is updated.

account_financial_report_webkit_xls/report/open_invoices_xls.py
```diff
@@ -725,10 +725,12 @@ def print_group_cumul_account(self, row_position, row_start_account,
     # selected
     def print_grouped_line_report(self, row_pos, account, _xs, xlwt, _p, data):
 
-        if account.grouped_ledger_lines and account.partners_order:
+        if account.grouped_ledger_lines and \
+           _p['partners_order'][account.id]:
```
pedrobaeza, line 729: Bad indentation. Should be more indented.

account_financial_report_webkit_xls/report/open_invoices_xls.py
```diff
@@ -757,13 +759,16 @@ def print_grouped_line_report(self, row_pos, account, _xs, xlwt, _p, data):
     # export the invoice AR/AP lines
     def print_ledger_lines(self, row_pos, account, _xs, xlwt, _p, data):
 
-        if account.ledger_lines and account.partners_order:
+        if _p['ledger_lines'][account.id] \
+           and _p['partners_order'][account.id]:
```
pedrobaeza, line 763: Bad indentation

pedrobaeza:
Why are you using `_p` instead of account variable?

anthony-muschang:
> Bad indentation

How the line should be indented? There is no pep8 error. Where can I find the guildline on this subject?

> Why are you using `_p` instead of account variable?

Since `account_financial_report_webkit` was migrated to v8, some data like `partners_order` are not stored as account variables anymore, but are instead stored in `_p`. See the pull request #20 for more information.

pedrobaeza:
There is no force in PEP8 standard to do it that way, that's why Travis doesn't complain, but as you can see here, it's advisable to do it, specially when you don't use parenthesis:

http://legacy.python.org/dev/peps/pep-0008/#indentation

But this is not a blocking issue. What I see a problem is to use Luc de Meyer branch in Travis instead of the official one, because we will need to change again when the PR is merged. What we have to do is to speed up https://github.com/OCA/reporting-engine/pull/5, and then, we can put the official one.

I understand the point about _p. Thanks for the explanation.

pedrobaeza:
Thanks for the review on report_xls PR.

I give you my :+1:

We need to wait for at least another one.

guewen:
:+1: 
Thanks

PR #35, opened 2014-11-26
[FIX] afr_webkit_xls: incorrect debit/credit in xls trial balance

sbidoul:
This is a migration bug. cc/ @anthony-muschang


pedrobaeza:
:+1:

nbessi:
:+1: that said `_p` is not  the best variable name...

sbidoul:
@nbessi agreed. I guess we'll have to live with these modules until we know more about the new v9 account reporting engine.

Can we fast-track this PR?

anthony-muschang:
:+1:

PR #37, opened 2014-11-29
[FIX] afr_webkit sorting issue

### LINKED ISSUE #36: v8 account_financial_report_webkit error when print journal
/account_financial_report_webkit/report/print_journal.py", line 129, in set_context
    objects.sort(key=lambda a: (a.journal_id.code, a.period_id.date_start))
AttributeError: 'account.journal.period' object has no attribute 'sort'

sbidoul:
Should fix #36

### DIFF
--- a/account_financial_report_webkit/report/open_invoices.py
+++ b/account_financial_report_webkit/report/open_invoices.py
@@ -91,7 +91,7 @@ class PartnersOpenInvoicesWebkit(report_sxw.rml_parse,
             return
         for part_id, plane_lines in account_br.ledger_lines.items():
             account_br.grouped_ledger_lines[part_id] = []
-            plane_lines.sort(key=itemgetter('currency_code'))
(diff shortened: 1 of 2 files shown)


pedrobaeza:
:+1:

guewen:
I don't know if it is relevant here but in v8 you can call a `sorted()` method on `recordsets`: https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model.sorted
The advantage is that it returns back a `recordset` and not a list.

sbidoul:
@guewen that is definitely relevant. Thanks for the tip. That is only part of the solution, though.

lmignon:
:+1: although I have not found a trace of the error on RunBot

pedrobaeza:
A rebase is needed

sbidoul:
This one is superseeded by #49

PR #38, opened 2014-11-30
[8.0] migrate account_export_csv

sbidoul:


pedrobaeza:
Seems a little repetitive with account_financial_report_xls, doesn't it? Maybe we can use the same mechanism (add a button in the corresponding wizard) instead of creating a new wizard.

sbidoul:
@pedrobaeza unfortunately, account_financial_report_xls exports XLS which is limited to 65k rows.

When we need to export hundreds of thousands of entries, I'm afraid there is no better solution than this one at the moment.

pedrobaeza:
Yeah, good point. And what about the rearragement inside the same wizard that prints/exports to Excel?

sbidoul:
That should be feasible in principle. However implementing the same logic as the account_financial_report_webkit wizards is non trivial and unfortunately we don't have the resource for that at the moment.

pedrobaeza:
OK, I don't bother you more then. Thanks for the migration.

:+1:

lmignon:
:+1:

sbidoul:
Thanks Pedro. The account reporting landscape is indeed quite confusing at the moment and there is no good solution that fits all use cases. I place great hopes in what Odoo is cooking for 9.0.

yvaucher:
:+1:

PR #41, opened 2014-12-17
Add account filter for CSV export

mdietrichc2c:
Added a new many2many on the CSV export wizard, similar to journal filter, in order to only export one or many specific accounts in the "journal entries" export.

### DIFF
--- a/account_export_csv/wizard/account_export_csv.py
+++ b/account_export_csv/wizard/account_export_csv.py
@@ -93,6 +93,13 @@ class AccountCSVExport(orm.TransientModel):
             'journal_id',
             'Journals',
             help='If empty, use all journals, only used for journal entries'),
+        'account_ids': fields.many2many(
+            'account.account',
+            'rel_wizard_account',
+            'wizard_id',
+            'account_id',
+            'Accounts',
+            help='If empty, use all accounts, only used for journal entries'),
         'export_filename': fields.char('Export CSV Filename', size=128),
     }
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Thanks :+1:

rdeheele:
:+1:

lepistone:
:+1: thanks.

sbidoul:
The merge pace tends to go faster and faster... this is good but has downsides too.

Actually, for this one I had two comments:
- I worry the additional argument to the _get_row methods will breaks modules that override it (the module is easy to extend if you want to add additional columns and we have one extension)
- how do we manage forward ports for this project? For account-analytic and account-financial-tools I do the effort of forward-porting with a git merge. We should do the same here (it's the first one that is hard, after which it is easy if done regularly).

lepistone:
@sbidoul 
- for the additional argument, what do you suggest? (we can make a new PR of course)
- for the forward-port, that would be a new PR 8.0 in any case, and we can do that as well.

Thanks for the heads-up!

sbidoul:
Can you pass the additional argument in the context? Not nice but backward compatible.

lepistone:
I'm not sure about this one @sbidoul: the methods starts with `_`. Do you have only one module doing that? If that's case, maybe it is less of an effort to update it that to use the context hack here.

What do the others think?

pedrobaeza:
I also don't think that nobody has inherited that internal method.

PR #45, opened 2014-12-30
[IMP] add supplier invoice number to XLS General Ledger

veloutin:


account_financial_report_webkit_xls/report/general_ledger_xls.py
```diff
@@ -30,10 +30,83 @@
 # import logging
 # _logger = logging.getLogger(__name__)
 
+
+class GeneralLedgerWebkitSupplier(GeneralLedgerWebkit):
+    """ Extends General Ledger Parser to add the supplier invoice
+    number in the move lines """
+
+    def _get_move_line_datas(self, move_line_ids,
```
pedrobaeza: As we have full control over parent module, please include supplier invoice number on account_financial_reporting_webkit instead overwriting here full method, or take the occasion to create some method that act as hooks for preparing the SQL and overwrite that hook.

max3903:
Given result is correct.

veloutin:
@pedrobaeza  I have refactored to avoid duplicating the SQL, as per your suggestion. Thanks!

account_financial_report_webkit/report/common_reports.py
```diff
@@ -534,18 +520,42 @@ def _get_move_line_datas(self, move_line_ids,
...
-    LEFT JOIN res_partner p on (l.partner_id=p.id)
-    LEFT JOIN account_invoice i on (m.id =i.move_id)
-    LEFT JOIN account_period per on (per.id=l.period_id)
-    JOIN account_journal j on (l.journal_id=j.id)
-    WHERE l.id in %s"""
-        monster += (" ORDER BY %s" % (order,))
+        """
+
+    def _get_move_line_datas(self, move_line_ids,
+                             order='per.special DESC, l.date ASC, \
```
pedrobaeza: Please make this also the returning value of a method that can be overriden.

veloutin:
@pedrobaeza Done the order as well. I left the parameter in the function to minimize possible impacts.

pedrobaeza:
:+1: 

Thanks!

yvaucher:
This has 2 approvals already I merge it

PR #46, opened 2015-01-04
german translations with higher coverage

rschnapka:
tried my best to get terms correct - fine tuning and adjustment in wording compared to main-steam modules (standard OpenERP/ODOO-Modules) might need follow-up work on either end. International financials and german apsects differ in some aspects. Wording and terminology are of course crucial here and should be dealt with carefully... collaboration would be greatly appreciated here.


pedrobaeza:
Excellent work you're doing with the translations!

:+1:

rschnapka:
thanks, it's fun and hopefully helpful too :-)

pedrobaeza:
Indeed, I'm not german, but german-speaking people will thank you a lot.

yvaucher:
Great work thanks :+1:

PR #48, opened 2015-01-07
Deactivate constraint on account.account (unnecessary if Webkit reports are used)

mdietrichc2c:
This PR is to deactivate the constraint _check_account_type on account.account; since the initial balance of partner reports is computed on-the-fly, we must allow payables & receivables accounts to use another type than "unreconciled".

### DIFF
--- a/account_financial_report_webkit/account.py
+++ b/account_financial_report_webkit/account.py
@@ -45,3 +45,16 @@ class AccountAccount(orm.Model):
     _defaults = {
         'centralized': False,
     }
+
+    # Related to WebKit reports: this constraint is not relevant anymore,
+    # since the initial balance of the partner reports
+    # (ledger, balance, open invoices) is computed on-the-fly, instead of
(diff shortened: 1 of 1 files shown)


fclementic2c:
:+1: 
Indeed we do not want payables & recevables accounts set as close_method = 'unreconciled' but 'balance' to reduce the size of the opening entry. 
This detailed information is useless with the financial report webkit since reports calculate initial balance itselves (ignoring the opening entries anyway) for partners

sbidoul:
:+1:

pedrobaeza:
But why deactivate it here in this module?

fclementic2c:
If you work with Odoo reports then, it is better to set the deferral method to unreconcile. 
If you work with the webkit reports, you do not need this set up so you have the possibility to dramatically reduce the size of the opening entry. In this case, this constraint is a pain.

pedrobaeza:
OK then :+1:

PR #51, opened 2015-01-23
[FIX] travis wkhtmltopdf download url

yvaucher:
Fix travis build

Example of failed build: https://travis-ci.org/OCA/account-financial-reporting/builds/47941388

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ virtualenv:
 install:
   - git clone https://github.com/OCA/reporting-engine.git ${HOME}/reporting-engine -b ${VERSION}
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
-  - sudo wget http://downloads.sourceforge.net/project/wkhtmltopdf/0.12.1/wkhtmltox-0.12.1_linux-precise-amd64.deb -P /tmp/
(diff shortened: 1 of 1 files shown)


sbidoul:
:+1: 

Do you confirm 0.12.2 does not work?

pedrobaeza:
I think it's more a question of file accesibility, because the URL has changed.

:+1:

sbidoul:
@pedrobaeza not only: 0.12.2 is there but has been reported to not work with Odoo.

pedrobaeza:
I'm using it without problems...

yvaucher:
No idea if 0.12.2 works but build are currently failing due to this. (updated description)

I'll do the same PR on 7.0

Cedric-Pigeon:
Hi,

I tested 012.2 last week and I encountered some problems with footers. They are not printed.
I tested the Odoo Sale Order and Invoice.

@pedrobaeza  Are you sure your reports are well printed ?

pedrobaeza:
I will try again and tell you back.

guewen:
:+1: 
I merge this one to fix the url and if we see that  012.2 works, we can still make another PR

Cedric-Pigeon:
@pedrobaeza issue https://github.com/odoo/odoo/issues/5177 confirm my test about wkhtmltopdf 0.12.2

pedrobaeza:
Yeah, I see. I was not awared of the footer thing, but yesterday I see also problems in one custom report (but not in the standard one), and now I see that it's due to the wkhtmltopdf version.

PR #61, opened 2015-02-11
Fix sum of partner balance in foreign currency.

mvaled:
The bottom line of each partner's group in the ledger only shows the initial balance (if any).  This PR fixes this issue.

There's an outstanding issue:  What to do about different currencies.  Probably we should allow to filter/group by currency.

### DIFF
--- a/account_financial_report_webkit/report/templates/account_report_partners_ledger.mako
+++ b/account_financial_report_webkit/report/templates/account_report_partners_ledger.mako
@@ -169,6 +169,8 @@
                                   <div class="act_as_cell sep_left amount">${formatLang(part_cumul_balance_curr) | amount }</div>
                                   ## curency code
                                   <div class="act_as_cell">${balance_forward_currency}</div>
+				  <!-- TODO:  What happens if there are differents currencies in the same group? -->
+				  <% cumul_balance_curr += line.get('amount_currency') or 0.0 %>
                              %endif
 
                           </div>


mvaled:
Hi @pedrobaeza,

Would you comment on this?

pedrobaeza:
Hi, Manuel, can you put an screenshot of before and after that illustrates the PR?

mvaled:
Hi Pedro,

I will, but now after the a pull from my fix is broken with:

   <% cumul_balance_curr += line.get('amount_currency') or 0.0 %>
   UnboundLocalError: local variable 'line' referenced before assignment

I will fix it.  Nevertheless, I attach a picture of the "wrongness".
You can see that only the "Initial balance" in EUR is afterwards totally
consumed but the "Accumulated Balance" keeps this number but it must be 0.

I will report back when I fix the PR.

Best regards and thanks,
Manuel.

pedrobaeza:
Attachments are not displayed here, sorry. Can you use an inline image or write it via GitHub interface?

mvaled:
before

mvaled:
Hi @pedrobaeza,  Image up.

sbidoul:
Can you rebase?

gurneyalex:
@mvaled rebase needed to get up to date information about the build status and a runbot build. Thanks.

ovnicraft:
@pedrobaeza it better going to close PRs w/o updates, this comes from 2015 :(

mvaled:
@ovnicraft I agree with closing.  Unfortunately I didn't (and still don't) have the time to complete it.

PR #64, opened 2015-03-03
[FIX][7.0] Sum on wrong column

llacroix:


pedrobaeza:
Can you put an example? I suppose that 8.0 version it's already affected, doesn't it?

llacroix:
@pedrobaeza what do you mean by example? like a before/after file? I can check for V8

llacroix:
Yes, the bug seems to affect the v8 too.

pedrobaeza:
Yes, please, a before/after comparison that illustrates it

bwrsandman:
:+1:

llacroix:
@pedrobaeza  here
Before:
screenshot from 2015-03-12 10 44 53

After:
screenshot from 2015-03-12 10 41 11

pedrobaeza:
But I'm seeing the same result in both...

llacroix:
@pedrobaeza I highlighted the column that does the wrong calculation. As you can see, the sum is done on "Cummul. Bal." instead of "Curr. Balance". The sum for Curr Balance in the first picture gives 300 below. (71.9 + 143.8) = 215.70 != 300

sbidoul:
Understood. :+1:

pedrobaeza:
Ah, OK. Thanks for the explanation.

:+1:

llacroix:
By the way, the number on the column that isn't present next to the formula is there, it just disappear when I select the cell. 

Here two more screenshots:

Before:
screenshot from 2015-03-12 11 45 53

After:
screenshot from 2015-03-12 11 45 27

llacroix:
Thank you

PR #66, opened 2015-03-18
in general ledger report, the sum is done on wrong column

laetitia-gangloff:
before :

image

after :

image

The regression was introduced with the PR #45 

I don't know how it is possible to add automated test on xls report, but it can be a good idea...


laetitia-gangloff:
And there is also a problem with initial balance :

image

vrenaville:
Thanks for this fix :+1:

mdietrichc2c:
:+1: thanks for the fix

nbessi:
:+1:

PR #68, opened 2015-03-23
[FIX] wrong position for balance result and initial balance

vrenaville:
The positions of field for initial balance and for the sum of balance field was wrong


guewen:
Is it related to https://github.com/OCA/account-financial-reporting/pull/66?

vrenaville:
@guewen It's seems, I have made the correction more or less in the same time
I close my PR

laetitia-gangloff:
@vrenaville hello, I see there is less line changed in your PR, what is the reason ?

PR #72, opened 2015-04-24
PR for code review

laetitia-gangloff:


mis_builder/__openerp__.py
```diff
@@ -0,0 +1,59 @@
...
+    'version': '0.1',
+    'category': 'Reporting',
+    'description': """
+    Management Information System Builder
+    """,
+    'author': 'ACSONE SA/NV',
+    'website': 'http://acsone.eu',
+    'depends': ['account',
+                'account_partial_balance',
+                'report_xls',
```
laetitia-gangloff, line 36: Specify from which repo comes these modules.

mis_builder/models/aep.py
```diff
@@ -0,0 +1,205 @@
...
+            mode = 'p'
+        elif mode == 's':
+            mode = 'e'
+        if account_codes.startswith('_'):
+            account_codes = account_codes[1:]
+        else:
+            account_codes = account_codes[1:-1]
+        account_codes = [a.strip() for a in account_codes.split(',')]
+        domain = domain or '[]'
+        domain = tuple(eval(domain))  # TODO: safe_eval
```
laetitia-gangloff, line 110: Maybe atleast set an empty dictionnary !

mis_builder/models/aep.py
```diff
@@ -0,0 +1,205 @@
...
+    def _parse_mo(self, mo):
+        """Split a match object corresponding to an accounting variable
+
+        Returns field, mode, [account codes], [domain expression].
+        """
+        field, mode, account_codes, domain = mo.groups()
+        if not mode:
+            mode = 'p'
+        elif mode == 's':
+            mode = 'e'
```
laetitia-gangloff, line 103: mode 's' is undocumented !
And I see, it is possible to use it to transform it in a mode 'e'.

Is it not better to use a migration to avoid this ?

mis_builder/models/mis_builder.py
```diff
@@ -0,0 +1,829 @@
...
+            'registry': self.pool,
+            'sum': sum,
+            'min': min,
+            'max': max,
+            'len': len,
+            'avg': lambda l: sum(l) / float(len(l)),
+        }
+        domain_p = self.compute_period_domain(cr, uid, c, False, False,
+                                              context=context)
+        domain_s = self.compute_period_domain(cr, uid, c, True, False,
```
laetitia-gangloff, line 614: not consistent with defined mode (s instead of e).

What happen if a new mode is added ? Is it possible to get something more generic ?

PR #74, opened 2015-05-02
[FIX] afr_webkit: fix aged partner balance test

### LINKED ISSUE #67: account_financial_report_webkit : broken yml test (aged_trial_balance.yml).
Travis fails because of broken test.

sbidoul:
Even if period_from is not visible in the wizard
it is still necessary for the validation check to pass.

Fixes #67


pedrobaeza:
:+1:

adrienpeiffer:
:+1:

PR #77, opened 2015-05-22
8.0 account financial report horizontal

hbrunn:
This is basically a rewrite to get rid of the inherited 6.0 code. So better ignore the diff and have a look at the code.

I inject a context key to determine which reports to show, so the most probable place where I might have messed up is in https://github.com/hbrunn/account-financial-reporting/blob/8.0-account_financial_report_horizontal/account_financial_report_horizontal/models/account_financial_report.py#L28 and `_get_children_by_order`


pedrobaeza:
@gurneyalex, can you please activate runbot here to try this module?

StefanRijnhart:
Nice! Failing tests are unrelated. The same errors occur in other PRs.

You can remove views/account_report.xml.

In README.rst, please change the description to reflect that this is no longer a 'rewrite of the financial reports from OpenERP 6.0' but instead only a cosmetic change to the existing reports.

Last remark: conventionally, in our region, the sign is reversed in the right column so that it does not usually contain negative amounts. However, this should probably not be changed to preserve consistency with the 'show debit/credit' option, which makes it clear that the amounts are actually the balances of the accounts, with its proper sign.

hbrunn:
@StefanRijnhart thanks!

Here the updated README.rst for convenience: https://github.com/hbrunn/account-financial-reporting/blob/8.0-account_financial_report_horizontal/account_financial_report_horizontal/README.rst

If anybody knows other countries where this reports are customary, please let me know

StefanRijnhart:
Thanks for the changes, :+1: (code review, test)

sbidoul:
Red because of #67

pedrobaeza:
Please rebase to get green results

hbrunn:
@pedrobaeza done

pedrobaeza:
:+1: Tried on runbot

PR #79, opened 2015-05-26
8.0 add modules intrastat_base and intrastat_product (v2)

alexis-via:
My second attempt at moving the 2 modules intrastat_base and intrastat_product from LP:new-report-intrastat to github/OCA


alexis-via:
It seems that this second attempt also failed :-(

PR #82, opened 2015-06-02
[IMP] Fix incompatibility with l10n_es_fiscal_year_closing

AlexandroTapia:
At l10n_es_fiscal_year_closing you need to setup several special periods
at the end of the fiscal year for holding the profit & loss and close
account moves, this was causing that balance report for last period for a
given fiscal year were not using the calculated opening balance.

### DIFF
--- a/account_financial_report_webkit/report/common_reports.py
+++ b/account_financial_report_webkit/report/common_reports.py
@@ -245,15 +245,32 @@ class CommonReportHeaderWebkit(common_report_header):
                                                      ['id', 'in', period_ids]])
 
     def get_included_opening_period(self, period):
-        """Return the opening included in normal period we use the assumption
-        that there is only one opening period per fiscal year"""
+        """Return the opening included in normal period only if
+        normal period is the first period on fiscal year"""
         period_obj = self.pool.get('account.period')
-        return period_obj.search(self.cursor, self.uid,
-                                 [('special', '=', True),
-                                  ('date_start', '>=', period.date_start),
-                                  ('date_stop', '<=', period.date_stop),
-                                  ('company_id', '=', period.company_id.id)],
-                                 limit=1)
+        # Get special period whit date in selected period
+        special_period_id = period_obj.search(
+            self.cursor, self.uid,
+            [('special', '=', True),
+             ('date_start', '>=', period.date_start),
+             ('date_stop', '<=', period.date_stop),
+             ('company_id', '=', period.company_id.id)],
+            limit=1
+        )
+        # Get first not special period from selected period fiscal year
+        first_period_id = period_obj.search(
+            self.cursor, self.uid,
+            [('special', '!=', True),
+             ('fiscalyear_id', '=', period.fiscalyear_id.id),
+             ('company_id', '=', period.company_id.id)],
+            order='date_stop asc', limit=1
+        )
+        # If first period of fiscal year is equal to selected period and
+        # there is a special_period we return special period else return False
+        if period.id in first_period_id and special_period_id:
+            return special_period_id
+
+        return False
 
     def periods_contains_move_lines(self, period_ids):
         if not period_ids:
--- a/account_financial_report_webkit/report/common_reports.py
+++ b/account_financial_report_webkit/report/common_reports.py
@@ -258,7 +258,7 @@ class CommonReportHeaderWebkit(common_report_header):
             limit=1
         )
         # Get first not special period from selected period fiscal year
-        first_period_id = period_obj.search(
+        first_period_ids = period_obj.search(
             self.cursor, self.uid,
             [('special', '!=', True),
              ('fiscalyear_id', '=', period.fiscalyear_id.id),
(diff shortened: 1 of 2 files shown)


account_financial_report_webkit/report/common_reports.py
```diff
@@ -245,15 +245,32 @@ def exclude_opening_periods(self, period_ids):
...
+        special_period_id = period_obj.search(
+            self.cursor, self.uid,
+            [('special', '=', True),
+             ('date_start', '>=', period.date_start),
+             ('date_stop', '<=', period.date_stop),
+             ('company_id', '=', period.company_id.id)],
+            limit=1
+        )
+        # Get first not special period from selected period fiscal year
+        first_period_id = period_obj.search(
```
pedrobaeza: Call it `first_period_ids` to identify that it's a list

pedrobaeza:
A little remark, but seems OK,

Please sign also the CLA.

sbidoul:
Multiple special periods in the same fiscal year looks weird to me. I always understood special periods contain opening balances and nothing else.

pedrobaeza:
Multiple special periods are for closing balances, not opening, because in Spain we need to make two moves at the end of the period: P&L balance move to a special account, that we make it on "PL/X" period, and the regularization of the rest of the accounts, that is made on "C/X" (closing) period.

I haven't examined yet the PR anyways to see its validity.

gurneyalex:
CLA ok

gurneyalex:
My PR fixing the runbot build on the 7.0 branch was recently merged. Can you rebase your PR?

AlexandroTapia:
@gurneyalex  Sorry for the delay on making the requestes change. Rebased now

gurneyalex:
@AlexandroTapia https://github.com/OpenPymeMx/account-financial-reporting/pull/19 shoud fix the travis build.

agb80:
Your PR is merged, waiting for test results.

PD: @gurneyalex unfortunately @AlexandroTapia don't work anymore with OpenPyme, appreciate if you pin me in case any new change is needed.

sbidoul:
Be very careful with multiple special periods. One pitfall, for instance, is that they don't exist anymore in Odoo v9 and the migration process simply drops all moves in special periods. So if you use them for anything else than opening entries your data will be incorrect in v9.

agb80:
Thanks for your remarks @sbidoul we know about the changes on Odoo 9, but if you see the MR is for openerp 7.0, we have been using in production for more than a year now with some customers.

gurneyalex:
:+1:

JordiBForgeFlow:
@sbidoul Regarding special periods, if odoo drops them all during migration to v9, is the initial opening balance of the company also lost? I assume that people use the special period to introduce the opening entries for their company, don't they?

sbidoul:
@jbeficent indeed I'm afraid those who use special period to initialize accounting will have more work when migrating to v9.

Here is  a quote from the Odoo migration team:

> We delete account_move_line if the corresponding account_period has special = True

IMO, the correct way to initialize accounting is to introduce the initial balance at the end of the previous fiscal year and do a normal opening using the standard Odoo feature.

JordiBForgeFlow:
@sbidoul Then probably people migrating to v9 will need to take the opening entries out of the special periods.

pedrobaeza:
So official migration scripts drops data from your DB when migrating?!! :scream:

Spanish fiscal year closing is a bit special, requiring at least three account moves: the loss&profit, the closing one (setting all balances to 0), and the opening one in the next fiscal year, putting again the balances. And this is totally needed by law.

sbidoul:
> So official migration scripts drops data from your DB when migrating?!!

I don't know if openupgrade has addressed that topic yet, but I see no other way given the v9 data model.

> And this is totally needed by law.

Sure, but does the law mandate special periods? Can't you do that as regular miscellaneous operations, potentially in a dedicated journal?

pedrobaeza:
Of course we are doing it that way (the module l10n_es_fiscal_year_closing will do). @jbeficent is driving the account migration script on OpenUpgrade, so I don't know what is the decision taken. 

In fact, thinking about it, Odoo approach seems correct: we now calculate the end of year balances on the fly, so we don't need that intermediate account entries... but I don't know why they didn't think about the law requirements (I remember we expressed that problem when the blueprints were released...).

JordiBForgeFlow:
@pedrobaeza @sbidoul The fiscal year closing procedure in spain generates entries that will be compatible with the v9 model. Because both P&L and BS accounts are closed in previous FY and reopened in subsequent FY. So, if you keep these journal entries in v9, and compute the BS dynamically, from the start of times until the desired time, the result will be formally correct.

IMHO people migrating in Spain to v9 should make a pre-migration adjustment, moving those entries out of the special periods, into normal periods. The migration should then work fine.

For companies using the standard FY closing procedure see here: https://github.com/OCA/OpenUpgrade/issues/622

sbidoul:
> moving those entries out of the special journals, into general journals

@jbeficent you mean "out of the special periods, into normal periods"?

JordiBForgeFlow:
@sbidoul Yes, sorry. Just corrected that.

pedrobaeza:
Closing as old version and no further interest.

PR #88, opened 2015-06-11
[FIX] Impossible to install account_report_webkit on database with 2million move_line objects

fedexin40:
If the account_move_line table is too big, update the last_rec_date field takes many time and it is not possible to install the module

### DIFF
--- a/account_financial_report_webkit/account_move_line.py
+++ b/account_financial_report_webkit/account_move_line.py
@@ -28,6 +28,13 @@ class AccountMoveLine(orm.Model):
     Last rec date is the date of the last reconciliation (full or partial)
     account move line"""
     _inherit = 'account.move.line'
+ 
+    def _auto_init(self, cr, context=None):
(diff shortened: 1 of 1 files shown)


lmignon:
:-1: I don't understand how your change solve your problem. What is the expected result?

fedexin40:
The use case is a database where already exist more than 2million rows on account_move_line table.

When installing account_report_webkit the `_auto_init` fuction try to evaluate the `last_rec_date` for all the rows in table iterarting very row on python. This obviusly makes the server hangs.

The expected result  is that installing the module it don't calculate the 2 millions rows and use the migrate SQL for populate the data.

gurneyalex:
@fedexin40 in that case I'd use an ad-hoc migration locally, and not potentially break the update without warning. 

Therefore I'm :-1: in it's current form, but would accept a patch explaining the problem and proposing your solution in the README.rst file.

feketemihai:
@fedexin40 I will close one since no activity for a long time. Please reopen it if still need it.

PR #86, opened 2015-06-07
[8.0] mis_builder

sbidoul:
This is the mis_builder PR.

Some examples reports are in the test data and in the module mis_builder_demo.
See also http://www.slideshare.net/acsone/misbuilder (slides show at Odoo Experience 2015).

The code is production ready but there are a few things left to do:
- [x] replace Acsone icon
- [x] README.rst 
- [ ] increase test coverage

PR #89, opened 2015-06-15
7.0 mis builder lmi

lmignon:
Back port mis_builder from 8.0 (https://github.com/OCA/account-financial-reporting/pull/86)


lmignon:
wrong target branch

PR #101, opened 2015-07-24
[FIX][7.0] There was a missing parameter in the function formatLang in the Reports

fedexin40:
There was a missing parameter in the function formatLang, monetary=True, to print monetary amounts

If was not definite the thousand separator in the languange
To print a report gave us a error
codec can't decode byte 0xe2 in position 2: ordinal not in range(128)

### DIFF
--- a/account_financial_report_webkit/report/templates/account_report_general_ledger.mako
+++ b/account_financial_report_webkit/report/templates/account_report_general_ledger.mako
@@ -153,14 +153,14 @@
                           ## counterpart
                           <div class="act_as_cell"></div>
                           ## debit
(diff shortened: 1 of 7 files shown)


pedrobaeza:
But which currency symbol is used. Don't you pass it as a parameter or set a context argument?

yvaucher:
`monetary` argument is used only when you don't setup monetary in language, this to select locale `mon_thousands_sep` instead of `thousands_sep`. So it make sens. :+1:

yvaucher:
@fedexin40 the error you encounter might come from what you have in `localeconv()['thousands_sep']`

feketemihai:
@fedexin40 I will close one since no activity for a long time. Please reopen it if still need it.

PR #102, opened 2015-07-27
8.0 Add "Aged Open Invoices" report in account_financial_report_webkit

alexis-via:
and some string updates on aged balance.

### DIFF
--- a/account_financial_report_webkit/report/aged_open_invoices.py
+++ b/account_financial_report_webkit/report/aged_open_invoices.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
+#    Author: Alexis de Lattre
 #    Author: Nicolas Bessi
+#    Copyright 2015 Akretion (www.akretion.com)
 #    Copyright 2014 Camptocamp SA
 #
 #    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


account_financial_report_webkit/report/aged_open_invoices.py
```diff
@@ -0,0 +1,425 @@
...
+
+        res = super(AccountAgedOpenInvoicesWebkit, self).set_context(
+            objects,
+            data,
+            ids,
+            report_type=report_type
+        )
+
+        aged_open_inv = {}
+        # {account1_id: {(0, 30): 0.0,
```
pedrobaeza: Remove this commented block

pedrobaeza:
A little PEP8 remark:

```
./account_financial_report_webkit/report/aged_open_invoices.py:197:9: E265 block comment should start with '# '
```

pedrobaeza:
Tried on runbot and seems OK, except 2 little things, :+1:

alexis-via:
@pedrobaeza It seems I missed your comments because they were made during my summer holidays.

I took into accout your remarks.

I deployed this PR with the "aged open invoices" report in the Odoo system of all my customers who use v8 in production with accounting, and they are happy with it.

pedrobaeza:
flake8 error unrelated, so :+1:

fclementic2c:
Is this not similar to https://github.com/OCA/account-financial-reporting/pull/118 (in v7)?

alexis-via:
Yes, it seems similar. My PR is more ancient and is for v8, but the feature seems to be the same.

pedrobaeza:
I merged this old one that is approved. Maybe the scope of only invoices instead of account move lines is more limited, but it can be extended later.

PR #104, opened 2015-08-03
[FIX] Fixes #95: fix the account_financial_report_webkit_xls.

### LINKED ISSUE #95: Xls bug in report Partners balance.
This  occurs when exported from the wizard to xls, the report does not print correctly balance column. 

Checking the report learn that the formula that calculates the total balance of the line, makes miscalculations because they do not have the initial balance column so goes the "D",  and column takes Code / Ref.

fv-clearcorp:
Formula fails when Initial Balance is not computed as shown
captura de pantalla de 2015-08-03 09 09 23

It occurs when the user selects "Date" filter or "Opening Only" filter
captura de pantalla de 2015-08-03 09 11 59
captura de pantalla de 2015-08-03 09 13 42

### DIFF
--- a/account_financial_report_webkit_xls/report/partners_balance_xls.py
+++ b/account_financial_report_webkit_xls/report/partners_balance_xls.py
@@ -376,7 +376,7 @@ class partners_balance_xls(report_xls):
                              regular_cell_style_decimal),
                         ]
                     else:
-                        debit_col = 3
+                        debit_col = 4
                     c_specs += [
                         ('debit', 1, 0, 'number', partner.get('debit', 0.0),
                          None, regular_cell_style_decimal),
--- a/account_financial_report_webkit_xls/report/partners_balance_xls.py
+++ b/account_financial_report_webkit_xls/report/partners_balance_xls.py
@@ -366,17 +366,15 @@ class partners_balance_xls(report_xls):
                              partner_ref if partner_ref else '')]
                 if _p.comparison_mode == 'no_comparison':
                     bal_formula = ''
+                    debit_col = 4
                     if _p.initial_balance_mode:
                         init_bal_cell = rowcol_to_cell(row_pos, 3)
                         bal_formula = init_bal_cell + '+'
-                        debit_col = 4
                         c_specs += [
                             ('init_bal', 1, 0, 'number', partner.get(
                                 'init_balance', 0.0), None,
                              regular_cell_style_decimal),
                         ]
-                    else:
-                        debit_col = 4
                     c_specs += [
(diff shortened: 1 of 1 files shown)


account_financial_report_webkit_xls/report/partners_balance_xls.py
```diff
@@ -376,7 +376,7 @@ def generate_xls_report(self, _p, _xs, data, objects, wb):
                              regular_cell_style_decimal),
                         ]
                     else:
-                        debit_col = 3
+                        debit_col = 4
```
pedrobaeza: But then debit_col is the same for both parts of the if

fv-clearcorp: Should I move the debit_col variable before the if statement?

pedrobaeza: Yeah, but I wonder why the differentiation and the variable debit_col for at the end, getting the same value.

fv-clearcorp: It must be so because when the initial balance column is not exported, the account / Partner name column takes columns a, b and c and the formula takes the data from the Code / Ref column.

pedrobaeza: OK, then, maybe this is for historical reasons. Make the change we have talked.

yvaucher:
This seems wrong to me :-1: 

In case of 'no_comparison' in comparison mode, the column 'init_bal' is added at col 3 so there is an extra column to take account of. So both value cannot be the same ones.

So your row must be: ['acc_title', 'partner_ref', 'init_bal', 'debit', 'credit', 'bal', ...]
Or ['acc_title', 'partner_ref', 'debit', 'credit', 'bal', ...]

@fv-clearcorp can you add a print screen of the xls with and without the exported initial balance?

ckasimis:
I have the same problem and none of the fixes worked for me.

I found out that problem ocurrs when i filter using date and initial balance is not displayed. If i filter using periods initial balance is displayed and the cells are correct.

Filtered by date
no initial balance

Filtered by period

with initial balance

fv-clearcorp:
Hi @yvaucher, @ckasimis already uploaded the screenshot with the problem I'm trying to solve. What do you think about?

yvaucher:
@fv-clearcorp I was asking the 2 different output, one with initial balance and one without. I think that your fix work for one but break the other.

AaronHForgeFlow:
Closed due to inactivity. Feel free to reopen in case of working on this again.

PR #107, opened 2015-08-05
[FIX] [8.0] Add reference field to account_financial_report_webkit_xls

JordiBForgeFlow:
This fix adds the field 'Reference' associated to the account move line to 'account_financial_report_webkit_xls', so that the resulting layout matches with 'account_financial_report_webkit'.


pedrobaeza:
I would put a little smaller the column. And now you're adding a column in between, so you need to move the totals one column right ("Cumulated Balance on Account" and so on).

seleccion_005

JordiBForgeFlow:
Thanks @pedrobaeza. I fixed.

pedrobaeza:
Thanks :+1:

antespi:
:+1:

ovnicraft:
Ready to merge ?

PR #108, opened 2015-08-14
7.0-Add filter to skip account with a null balance

florian-dacosta:
In the partner ledger report.
Some companies have so much journal items in one fical year that the excel file become too big.
Skipping account with a null balance resolve the problem and make a cleaner report.

### DIFF
--- a/account_financial_report_webkit/report/common_partner_balance_reports.py
+++ b/account_financial_report_webkit/report/common_partner_balance_reports.py
@@ -36,7 +36,8 @@ class CommonPartnerBalanceReportHeaderWebkit(CommonBalanceReportHeaderWebkit,
     def _get_account_partners_details(self, account_by_ids, main_filter,
                                       target_move, start, stop,
                                       initial_balance_mode,
-                                      partner_filter_ids=False):
+                                      partner_filter_ids=False,
+                                      exclude_null_balance=False):
         res = {}
         filter_from = False
         if main_filter in ('filter_period', 'filter_no', 'filter_opening'):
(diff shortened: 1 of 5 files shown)


account_financial_report_webkit/report/common_partner_balance_reports.py
```diff
@@ -75,11 +75,16 @@ def _get_account_partners_details(self, account_by_ids, main_filter,
                             {'init_balance': initial_balances['init_balance']})
 
             # compute balance for the partner
+            partner_to_remove = []
             for partner_id, partner_details in details.iteritems():
                 details[partner_id]['balance'] = details[partner_id].\
                     get('init_balance', 0.0) + \
                     details[partner_id].get('debit', 0.0) - \
                     details[partner_id].get('credit', 0.0)
+                if not round(details[partner_id]['balance'], 4):
```
pedrobaeza: Here exclude_null_balance is not taken into account

pedrobaeza: Use decimal_precission to solve the account accuracy

florian-dacosta: I take exclude_null_balance into account now...
I am not sure how to use decimal_precision since I have no cursor in the method

account_financial_report_webkit/report/common_partner_balance_reports.py
```diff
@@ -75,11 +77,17 @@ def _get_account_partners_details(self, account_by_ids, main_filter,
                             {'init_balance': initial_balances['init_balance']})
 
             # compute balance for the partner
+            partner_to_remove = []
             for partner_id, partner_details in details.iteritems():
                 details[partner_id]['balance'] = details[partner_id].\
                     get('init_balance', 0.0) + \
                     details[partner_id].get('debit', 0.0) - \
-                    details[partner_id].get('credit', 0.0)
+                if exclude_null_balance:
```
pedrobaeza: You can put both conditions in one if sequence if you want

pedrobaeza:
Check syntax, because it's not correct.

PR #110, opened 2015-08-20
[FIX] Subprocess encoding error when printing General Ledger

ghost:
In some Python versions/environments it seems that the `subprocess.call()` function could throw a `TypeError` exception if the arguments list doesn't share the same string type (_ASCII_ or _Unicode_) among all its elements:

> execv() arg 2 must contain only strings.

As there are some arguments that are read from database, and translated by OpenERP/Odoo, we must ensure that them all share the same string type. This patch achieve this by using some encoding fallbacks in case of error and in our tests we have not experienced this issue anymore.

If you agree with this patch it could be propagated to the `8.0` branch.

### DIFF
--- a/account_financial_report_webkit/report/webkit_parser_header_fix.py
+++ b/account_financial_report_webkit/report/webkit_parser_header_fix.py
@@ -149,8 +149,29 @@ class HeaderFooterTextWebKitParser(webkit_report.WebKitParser):
         command.append(out_filename)
         stderr_fd, stderr_path = tempfile.mkstemp(text=True)
         file_to_del.append(stderr_path)
+
+        # The encoding can break subprocess in some Python versions,
+        # ensure that all args are in ASCII or Unicode.
+        encoding_parsers = [
+            lambda s: s,
+            lambda s: s.decode('utf-8-sig') if isinstance(s, str) else s,
+            lambda s: s.encode('utf-8-sig') if isinstance(s, unicode) else s,
+            lambda s: s.decode('utf-8') if isinstance(s, str) else s,
+            lambda s: s.encode('utf-8') if isinstance(s, unicode) else s,
+        ]
+
         try:
-            status = subprocess.call(command, stderr=stderr_fd)
+            for encoding_parser in encoding_parsers:
+                try:
+                    comm = [encoding_parser(c) for c in command]
+                    _logger.debug('Running command: %r', comm)
+                    status = subprocess.call(comm, stderr=stderr_fd)
+                except Exception as e:
+                    status = str(e)
+                else:
+                    _logger.debug('Success!')
+                    break
+
             os.close(stderr_fd)  # ensure flush before reading
             stderr_fd = None  # avoid closing again in finally block
             fobj = open(stderr_path, 'r')


account_financial_report_webkit/report/webkit_parser_header_fix.py
```diff
@@ -149,8 +149,29 @@ def generate_pdf(self, comm_path, report_xml, header, footer, html_list,
...
+        ]
+
         try:
-            status = subprocess.call(command, stderr=stderr_fd)
+            for encoding_parser in encoding_parsers:
+                try:
+                    comm = [encoding_parser(c) for c in command]
+                    _logger.debug('Running command: %r', comm)
+                    status = subprocess.call(comm, stderr=stderr_fd)
+                except Exception as e:
```
pedrobaeza, line 169: Why do you still enclose this in a try block, making silent any possible error?

diagramsoftware, line 169: It isn't silent at all, it returns the latest error (if any) at line no. 187. Maybe the exception block could be changed into this to log any possible error:

``` python
except Exception as e:
    _logger.exception(e)
```

pedrobaeza, line 169: I still prefer to let Odoo shows the error the same way it does now with `execv() arg 2 must contain only strings.`

diagramsoftware, line 169: Well, the problem is that all the elements in the arguments list are always `strings` –as expected– and that this issue appears after some time of use, which makes this bug so difficult to reproduce.

See the next commands list example –it sometimes works, it sometimes doesn't–:

> ['/usr/local/bin/wkhtmltopdf', '--quiet', '--encoding', 'utf-8', '--orientation', 'Landscape', '--page-size', 'A4', '--header-font-name', 'Helvetica', '--footer-font-name', 'Helvetica', '--header-font-size', '10', '--footer-font-size', '6', '--header-left', u'LIBRO MAYOR - My Company, S.L. - EUR', '--header-spacing', '2', '--footer-left', '13/05/2015 15:10:35', '--footer-right', u'P\xe1gina [page] de [topage]', '--footer-line', '/tmp/tmpNJ0AlL0.body.html', '/tmp/webkit.tmp.2mzEzV.pdf']

The only clue is that setting a different user's default language seems to fix it (e.g. from **es** to **en_GB** or **es_CL**) so it may be a translations-related issue with **subprocess**, as the translated terms seems to be in _Unicode_ –see the `u''`– instead of _ASCII_. But –as mentioned– it sometimes works "out of the box"...

Clearing, building and/or overwriting the translation terms doesn't always helps fixing this bug, only forcing the encoding as shown.

PS: If it helps, we're using `Debian GNU/Linux 7.8 (wheezy)` with `es_ES.UTF-8` as the **openerp** user locale, and `Python 2.7.3` with the next code snippet in the `sitecustomize.py`:

```
import sys
sys.getdefaultencoding('utf-8')
```

lepistone:
Thanks for the PR!

I agree with you that what Odoo does with `str` vs `unicode` is bad (i.e. translations return `unicode` when they actually translate, but string literals are written as `str`.

On the other hand, here I read commentaries about `ASCII` that don't make sense to me: a `str` might be encoded with any encoding, including `utf-8`.

I gave a look at the point where the error is generated:

https://github.com/python/cpython/blob/2.7/Modules/posixmodule.c#L3210

from the C-api documentation https://docs.python.org/2/c-api/arg.html#c.PyArg_Parse , it looks like that bytestrings are accepted and assumed to be in the `Py_FileSystemDefaultEncoding`, whatever that is. I have no idea what that is in the cases that break.

I am :-1: to the approach of trying many combinations of conversions until one doesn't break. Are you sure there isn't one that always works (like decoding everything to `unicode`, or encoding everything to `utf-8` or `latin1`) ?

Thanks!

JGarcia-Panach:
I have the same error when I try to print general ledger.
I have applied the changes proposed by @diagramsoftware and it works now.
(Local enviroment with Odoo V8)

PR #113, opened 2015-08-25
import guard for report_xls

lepistone:
The dependency is indeed declared in **openerp**.py and will be
enforced. Odoo evaluates all python files, even for modules which are
not installed, and crashes if an unneeded dependency is needed.

### DIFF
--- a/mis_builder/report/__init__.py
+++ b/mis_builder/report/__init__.py
@@ -22,5 +22,9 @@
 #
 ##############################################################################
 
-from . import mis_builder_xls
+try:
+    from . import mis_builder_xls
+except ImportError:
+    pass  # this module is not installed
+
 from . import report_mis_report_instance


pedrobaeza:
Why not protect `from openerp.addons.report_xls.report_xls import report_xls` in report/mis_builder_xls.py?

lepistone:
@pedrobaeza I tried that before, but it doesn't work. `report_xls` is then undefined, and we get failures later.

pedrobaeza:
I did it this way: https://github.com/OCA/l10n-spain/blob/8.0/account_balance_reporting_xls/report/reporting_xls.py#L30, but it's a bit heavy also...

lepistone:
@pedrobaeza at least in `__init__.py` I don't pollute the real code of the module, so I prefer it that way.

pedrobaeza:
OK, let's do it this way.

:+1:

archetipo:
:+1:

lmignon:
@gurneyalex @lepistone @pedrobaeza This kind of error is too recurring. We must find an other way to avoid this. Our code becomes polluted by all these tricks required to turn around a bug in Odoo itself. https://github.com/odoo/odoo/issues/7507

pedrobaeza:
Yeah, but meanwhile we should workaround them. Do you have a better suggestion?

lepistone:
@lmignon of course I'd love to avoid this misfeature of Odoo evaluating all uninstalled modules. I'm open to a general solution. However I need this workaround now because it breaks existing instances badly.

lmignon:
@lepistone @pedrobaeza I'm not against the fix :smirk: 
@pedrobaeza We should at least ask to Odoo to fix this issue... They doesn't explain why they load all addons from addons path containing static files and  controllers but IMO it's like hacking the system... Even worst, this way of auto loading all addons following this rule introduce some security vulnerabilities.

pedrobaeza:
I know, but previous attempts to Odoo were closed without further explanations.

gurneyalex:
@lmignon they know about it. They depend on this behaviour, and they are not going to change it.

PR #116, opened 2015-08-28
[IMP] mis_builder: update min/max function signatures

sbidoul:
so min(x, y, ...) and min([x, y, ..]) both work as expected.

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -81,16 +81,16 @@ def _avg(l):
     return sum(l) / float(len(l))
 
 
-def _min(l):
+def _min(*l):
     if not l:
         return None
-    return min(l)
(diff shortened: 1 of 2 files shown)


laetitia-gangloff:
:+1:

mis_builder/models/mis_builder.py
```diff
@@ -81,16 +81,16 @@ def _avg(l):
     return sum(l) / float(len(l))
 
 
-def _min(l):
+def _min(*l):
```
pedrobaeza: Why don't call it args for being easily recognisable?

lmignon:
:+1:

pedrobaeza:
:+1:

sbidoul:
This one is good for merging.

PR #117, opened 2015-08-31
silence warnings

lepistone:


pedrobaeza:
:+1:

sbidoul:
I'm still very uneasy with such brutal silencing of import errors. Things will become hard to debug when a real error will lurk in there.

BTW, what do you think about this proposal: https://github.com/odoo/odoo/issues/7507#issuecomment-136434162

lepistone:
@sbidoul I'm not happy of silencing errors of course, but I don't know any other way that works today.

feketemihai:
@lepistone I will close one since no activity for a long time. Please reopen it if still need it.

lepistone:
Absolutely @feketemihai, thanks!

Issue #114, opened 2015-08-25
Aged Partner Balance from invoice date or due date


max3903:
When generating the Aged Partner Balance report, we want to choose the date to compute the age of the invoices. The age could be based on:
- the invoice date
- the due date

Mockup of the wizard:
![customer invoices - openerp 2015-09-02 11-49-55](https://cloud.githubusercontent.com/assets/939952/9636715/f60ab966-516a-11e5-8fcf-d97c7ab15476.png)

pedrobaeza:
Seems to be fixed.

Issue #76, opened 2015-05-21
Aged Partner Balance by invoice


max3903:
Provide an aged partner balance report with invoice details:

| Partner | Invoice | 0 - 30 | 30 - 60 | 60 - 90 | 90+ | Total |
| --- | --- | --- | --- | --- | --- | --- |
| Agrolait |  |  |  |  |  |  |
|  | SAJ/2015/0001 |  |  |  | 1000 $ | 1000 $ |
|  | SAJ/2015/0003 | 2000 $ |  |  |  | 2000 $ |
|  | **Total** | **2000 $** |  |  | **1000 $** | **3000 $** |
| Asustek |  |  |  |  |  |  |
|  | SAJ/2015/0002 |  | 5000 $ |  |  | 5000 $ |
|  | **Total** |  | **5000 $** |  |  | **5000 $** |
| **Total** |  | **2000 $** | **5000 $** |  | **1000 $** | **8000 $** |

Mockup of the wizard:
![customer invoices - openerp 2015-09-02 11-49-55](https://cloud.githubusercontent.com/assets/939952/9636715/f60ab966-516a-11e5-8fcf-d97c7ab15476.png)

max3903:
https://github.com/moturi12/shared_public_addons_v7

pedrobaeza:
This seems to be fixed and any way, it's a very old version. Closing.

PR #118, opened 2015-09-04
[7.0] Improve Aged Partner Balance report

Ehtaga:
- Add a boolean that allow the user to generate an aged partner balance report with invoice details #76 
- Allow the user to choose the aging method : from Due Date (default) or Invoice Date #114 

Pdf with invoice details:
aged_partner_balance


max3903:
Tested on runbot
:+1:

max3903:
ping @jgrandguillaume

fclementic2c:
:+1:

fclementic2c:
@guewen @gurneyalex  any time for a review, thxs ? It is a really usefull feature for accountants.

nmoturi:
The ageing seems to ignoring the transactions prior fiscal year if run with date option, I believe when the date option from and to is selected it needs to consider all the open invoices in that range, Ideally when date option is selected it need to have only one field "as at date"

PR #121, opened 2015-09-11
translation / context in browse

BT-ojossen:

PR #123, opened 2015-09-26
8.0 account_financial_report_webkit fix for multilingual Chart of Acc…

luc-demeyer:
This PR ensures that the user language is correctly passed in the context which is required for users with the "translate=True" on the account.account name field.

### DIFF
--- a/account_financial_report_webkit/report/common_partner_balance_reports.py
+++ b/account_financial_report_webkit/report/common_partner_balance_reports.py
@@ -232,6 +232,8 @@ class CommonPartnerBalanceReportHeaderWebkit(CommonBalanceReportHeaderWebkit,
         return accounts_details_by_ids, comp_params
 
     def compute_partner_balance_data(self, data, filter_report_type=None):
+        lang = self.localcontext.get('lang')
+        lang_ctx = lang and {'lang': lang} or {}
         new_ids = data['form']['account_ids'] or data[
             'form']['chart_account_id']
         max_comparison = self._get_form_param(
@@ -271,7 +273,7 @@ class CommonPartnerBalanceReportHeaderWebkit(CommonBalanceReportHeaderWebkit,
         # get details for each accounts, total of debit / credit / balance
         accounts_by_ids = self._get_account_details(
             account_ids, target_move, fiscalyear, main_filter, start, stop,
-            initial_balance_mode)
+            initial_balance_mode, context=lang_ctx)
 
         partner_details_by_ids = self._get_account_partners_details(
             accounts_by_ids, main_filter, target_move, start, stop,
@@ -293,7 +295,8 @@ class CommonPartnerBalanceReportHeaderWebkit(CommonBalanceReportHeaderWebkit,
                 comp_accounts_by_ids.append(comparison_result)
         objects = self.pool.get('account.account').browse(self.cursor,
                                                           self.uid,
-                                                          account_ids)
+                                                          account_ids,
+                                                          context=lang_ctx)
 
         init_balance_accounts = {}
         comparisons_accounts = {}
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Reviewing code, seems :+1:, but not tested

luc-demeyer:
@pedrobaeza 
Thanks for the fast review.
I have tested partner reports, trial balance, general ledger with and without comparison.

BT-ojossen:
In this file "account_financial_report_webkit/report/common_reports.py" we need also extension with the context. Please have a look in #121.

luc-demeyer:
@BT-ojossen 
I tested all reports but maybe I forgot a specific use case.
Can you let me know which report is affected by not having the context passed in the "common_reports.py" ?

BT-ojossen:
@luc-demeyer 
Hi Luc, I guess you're right, it should be ok like this.

luc-demeyer:
I think this one is ready for merge.

luc-demeyer:
cf. https://github.com/OCA/account-financial-reporting/issues/133#issuecomment-155180329
Can someone commit this PR ?

pedrobaeza:
Needs another review at least. I'm not comfortable enough with only my code review to merge by myself.

account_financial_report_webkit/report/common_partner_balance_reports.py
```diff
@@ -259,7 +261,7 @@ def compute_partner_balance_data(self, data, filter_report_type=None):
         # get details for each accounts, total of debit / credit / balance
         accounts_by_ids = self._get_account_details(
             account_ids, target_move, fiscalyear, main_filter, start, stop,
-            initial_balance_mode)
+            initial_balance_mode, context=lang_ctx)
```
damdam-s, line 276: Why not use `context=data['form']['used_context']` (or a similar form) instead of this limited context which only contains `lang` ?
`data['form']['used_context']` is defined here : https://github.com/odoo/odoo/blob/8.0/addons/account/wizard/account_report_common.py#L183 and `lang` key is already defined.

luc-demeyer, line 276: This is a very minimal change. By only adding context=lang_ctx to those places in the code where there was no context passing before I fix the multi-language problem of the account.account,name field without risk to break something else. 
I tested all reports with this change and everything work fine, hence I prefer that we merge it this way. 
Seen the fact that many methods within the standard account module react differently depending on what is in the context, I think we risk to break a set of reports which work fine for several years now (and we all know how sensitive accountants are when this module would start to return wrong financial data).

damdam-s, line 276: @luc-demeyer I need this (https://github.com/luc-demeyer/account-financial-reporting/pull/1) for a customer project. Can you have a look please ? Thanks

droelneo:
Tested and ok !

pedrobaeza:
@luc-demeyer, are you going to include the patch by @damdam-s?

pedrobaeza:
https://github.com/OCA/account-financial-reporting/pull/121 is a patch for solving the same problem, please judge which one is best.

BT-ojossen:
+1 Please merge

melroy89:
@pedrobaeza 
@sbidoul 
Please merge? Plus sync to all new releases up to and including v11.

pedrobaeza:
This has conflicts. Forward porting this patch should be done by a contributor.

luc-demeyer:
merging this fix should imho not be an issue.
Forward port is indeed another issue.

lasley:
@luc-demeyer Merge is not possible at the moment - there are conflicts (the branch needs to be rebased).

luc-demeyer:
@lasley 
rebased hence can be merged now.

lasley (approved):

PR #127, opened 2015-10-27
[ADD] mis_builder: hooks for analytic filtering

sbidoul:


mis_builder/models/aep.py
```diff
@@ -328,7 +328,7 @@ def get_aml_domain_for_dates(self, date_from, date_to,
         return expression.normalize_domain(domain)
 
     def do_queries(self, date_from, date_to, period_from, period_to,
-                   target_move):
+                   target_move, additional_move_line_filter=[]):
```
pedrobaeza, line 331: Don't initialize to a mutable argument, but to None, to avoid side effects: https://pythonconquerstheuniverse.wordpress.com/category/python-gotchas/

sbidoul:
Hm. Wrong branch... See #128

mis_builder/models/aep.py
sbidoul, line 331: I know that gotcha. But here I don't mutate the argument in the method, so it's safe.

pedrobaeza, line 331: No, but this means that any variable initalized to the same value (`[]`) in any method will be shared, so you must change it.

sbidoul, line 331: > any variable initalized to the same value ([]) in any method will be shared

What do you mean?

pedrobaeza, line 331: In the link is all explained. Multiple calls to the same method will keep in the same variable what you append on each call.

sbidoul, line 331: Sure, I know the risk. But in this case I never append anything in the method so it is completely safe. Anyway, I'll change.

PR #129, opened 2015-10-29
[FIX] account_financial_report_webkit: search of first opening period…

jbaudoux:
search of first opening period cannot be based on simply ordering moves by dates (could be that an old document is entered later in the accounting). Find first fiscal year having moves and take its opening period (as it was documented in the method).

Method Documentation:
||  57         Returns the browse record of the period with the `special` flag, which
||  58         is the special period of the first fiscal year used in the accounting.
||  59 
||  60         i.e. it searches the first fiscal year with at least one journal entry,
||  61         and it returns the id of the first period for which `special` is True
||  62         in this fiscal year.
||  63 
||  64         It is used for example in the partners reports, where we have to
||  65         include the first, and only the first opening period.
||  66 
||  67         :return: browse record of the first special period.

### DIFF
--- a/account_financial_report_webkit/report/common_partner_reports.py
+++ b/account_financial_report_webkit/report/common_partner_reports.py
@@ -66,16 +66,15 @@ class CommonPartnersReportHeaderWebkit(CommonReportHeaderWebkit):
 
         :return: browse record of the first special period.
         """
-        move_line_obj = self.pool.get('account.move.line')
-        first_entry_id = move_line_obj.search(
-            self.cr, self.uid, [], order='date ASC', limit=1)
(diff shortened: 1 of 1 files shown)


luc-demeyer:
@jbaudoux 
The issue and the solution issue is a bit more complex.
Cf. https://github.com/OCA/account-financial-reporting/pull/130

For more details, cf. the history of :
https://code.launchpad.net/~luc-demeyer/account-financial-report/7.0-account_financial_report_webkit-first-special-fix2/+merge/216380

feketemihai:
@jbaudoux I will close one since no activity for a long time. Please reopen it if still need it.

PR #131, opened 2015-11-02
[IMP] mis_builder: _get_additional_query_filter on period model

sbidoul:
This is a better api that is more consistent in case we want to have an analytic filter on the columns.

Thanks @jbeficent for the productive discussion and tests of this new API.


pedrobaeza:
:+1:

JordiBForgeFlow:
:+1: Thanks to for your insights @sbidoul 

This would be the result of use with separate module "mis_builder_analytic_account". I'll propose a separate PR for that:
image

laetitia-gangloff:
:+1:

alexis-via:
:+1:

PR #139, opened 2015-11-25
[IMP]: Allow to export excel reports with more than 65536 rows

hugosantosred:
- This modification will allow to export excel accounting reports with more than 65536 rows by creating new sheets in the workbook.

### DIFF
--- a/account_financial_report_webkit_xls/report/general_ledger_xls.py
+++ b/account_financial_report_webkit_xls/report/general_ledger_xls.py
@@ -73,9 +73,12 @@ COLS_TOT = sum((COLS_COA, COLS_FY, COLS_DF, COLS_AF, COLS_TM, COLS_IB))
 class general_ledger_xls(report_xls):
     column_sizes = [x[1] for x in _column_sizes]
 
-    def generate_xls_report(self, _p, _xs, data, objects, wb):
-
-        ws = wb.add_sheet(_p.report_name[:31])
+    def _new_ws_with_header(self, _p, _xs, data, wb, sheet_index=None):
+        report_name = _p.report_name[:31]
+        if sheet_index:
+            report_name = "%s %s" % (report_name[:29], sheet_index)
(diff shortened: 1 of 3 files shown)


rafaelbn:
This is a limitation in Odoo in all objects. Have you ask Odoo about this limitation?

hugosantosred:
No, I didn't. It's a limitation exporting excel files with xlwt library and using report_xls module from OCA/reporting-engine. Odoo could do something about this?

I made the changes in this repository because the base report_xls module provides only the basic tools to generate your reports, and I think than the better place to fix this is in the report itself.

pedrobaeza:
The limit is in XLS format, not in Odoo nor the library: http://superuser.com/questions/366468/what-is-the-maximum-allowed-rows-in-a-microsoft-excel-xls-or-xlsx

A new report_xlsx that exports to XLSX will solve it, or this patch to split the export through several sheets, but a better patch would be on the report_xls itself, so each report is agnostic to the number of results. Do you dare to try, @hugosantosred?

hugosantosred:
Ok! @pedrobaeza I will try to do it in report_xls, to divide in sheets any xls report that has more than 65536 rows.

luc-demeyer:
I prefer the solution proposed by Pedro: create a report_xlsx module and as a next step adapt the modules depending on report_xls.
Since report_xls is only a few lines of code, I don't think it's a major job to create the report_xlsx alternative. We need to switch from pyxlwt to either xlsxwriter or openpyxl in order to generate xlsx files.
Does anyone has concrete recommendations/experience to help in the choice of openpyxl versus xlsxwriter ?

In the meanwhile we can of course adapt report_xls (and hence have a scaleable solution for people still running of old versions of excell.

damendieta:
I think that the report_xlsx module has been done. May we close this PR? There is any reason to keep it open?.

Daniel.

StefanRijnhart:
@damendieta agreed, closing.

PR #140, opened 2015-11-25
[IMP] Performance improvement on partners ledger with large amounts of data

hugosantosred:
This commit is made to improve performance with large sets of data by requesting the data of all partners in a single postgresql query instead a query by partner. The query result is grouped and transformed in the function and then passed to the report.


feketemihai:
@hugosantosred I will close one since no activity for a long time. Please reopen it if still need it.

PR #146, opened 2015-12-12
[mis_builder] Added column level CSS report

nmoturi:
This will add a column level CSS support.
unnamed

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -41,6 +41,10 @@ from .aep import AccountingExpressionProcessor as AEP
 from .aggregate import _sum, _avg, _min, _max
 
 _logger = logging.getLogger(__name__)
+DATE_LENGTH = len(datetime.date.today().strftime(
+    tools.DEFAULT_SERVER_DATE_FORMAT))
+DATETIME_LENGTH = len(datetime.datetime.now().strftime(
+    tools.DEFAULT_SERVER_DATETIME_FORMAT))
 
 
 class AutoStruct(object):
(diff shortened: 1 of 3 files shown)


mis_builder/models/mis_builder.py
```diff
@@ -41,10 +41,6 @@
 from .aggregate import _sum, _avg, _min, _max
 
 _logger = logging.getLogger(__name__)
-DATE_LENGTH = len(datetime.date.today().strftime(
-    tools.DEFAULT_SERVER_DATE_FORMAT))
-DATETIME_LENGTH = len(datetime.datetime.now().strftime(
-    tools.DEFAULT_SERVER_DATETIME_FORMAT))
```
lmignon, line 47: You must preserve this code and keep the method '_utc_midnight' unchanged

mis_builder/models/mis_builder.py
```diff
@@ -606,11 +600,11 @@ def _fetch_queries(self, cr, uid, c, context):
                     cr, uid, obj_ids, field_names, context=context)
                 s = AutoStruct(count=len(data))
                 if query.aggregate == 'min':
-                    agg = _min
+                    agg = min
```
lmignon: Plz revert this change

mis_builder/models/mis_builder.py
```diff
@@ -606,11 +600,11 @@ def _fetch_queries(self, cr, uid, c, context):
...
                 elif query.aggregate == 'max':
-                    agg = _max
+                    agg = max
```
lmignon: Same here

mis_builder/models/mis_builder.py
```diff
@@ -606,11 +600,11 @@ def _fetch_queries(self, cr, uid, c, context):
...
                 elif query.aggregate == 'avg':
-                    agg = _avg
```
lmignon, line 613: Same here

nmoturi:
Thanks Laurent, I will check if we have latest code and push it again.

laetitia-gangloff:
@nmoturi Why did you revert all your commit ?!
Is this PR always on ?

sbidoul:
Note we are planning an evolution of mis_builder to add a proper notion of styles for rows and cells. The goal is to configure styles in a single place, and to allow preserving styles when exporting to Excel. I'll see if it works to add column styles too.

sbidoul:
Preserved as an enhancement suggestion in https://github.com/OCA/mis-builder/issues/14

PR #147, opened 2015-12-17
[PORT] Port MIS Builder in 9.0

adrienpeiffer:
Enjoy !
- [x] Depends on https://github.com/OCA/reporting-engine/pull/38


JordiBForgeFlow:
In Odoo 9 Enterprise MIS builder fails because 'account.menu_account_reports' does not exist. Please replace the parent of "mis_report_view_menu" to "account.account_management_menu".

Same with menu item mis_report_instance_view_menu. Parent should be "account.menu_finance_reports"

JordiBForgeFlow:
test file 'mis.report.query.csv' fails because "model_account_analytic_balance" does not exist in v9.

gurneyalex:
you may want to use #149 too

adrienpeiffer:
I'll create a new PR on Sorrento Branch

PR #151, opened 2016-01-27
[IMP][mis_builder] MIS Builder widget improvement, add print and export button in the widget.

adrienpeiffer:
In the Qweb file, I simply add a general div causing a large diff. I also added the two buttons.

PR #154, opened 2016-02-05
[WIP][FIX] Flake8 issues

adrienpeiffer:


adrienpeiffer:
Travis is green :smile:

pedrobaeza:
:+1: Fast-tracking then

mis_builder/static/src/xml/mis_widget.xml
```diff
@@ -42,7 +42,7 @@
                                     </a>
                                 </t>
                                 <t t-if="!value_value.drilldown">
-                                    <t t-esc="value_value.val_r"/>
+                                    <t t-esc="value_value.val_r" t-field-options='{"widget": "monetary"}'/>
```
yvaucher, line 45: This doesn't seems linked to pep8. Was it intended ?

yvaucher:
@pedrobaeza Wrong fast tracking

mis_builder/static/src/xml/mis_widget.xml
pedrobaeza, line 45: Sorry. I missed this part in my review before fast-tracking, but I think it doesn't hurt, and most as they're the authors of the module.

pedrobaeza:
Yeah, I see. We can revert if you consider so.

adrienpeiffer:
@yvaucher @pedrobaeza I will correct it

yvaucher:
I guess this change was included in https://github.com/OCA/account-financial-reporting/pull/151

PR #157, opened 2016-02-10
[7.0] mis_builder : in compute do not set val in localdict if val is none

laetitia-gangloff:
Use case is the following :

line total is a + b + c
line a is 5
line b is d + e
line c is 3
line d is 7
line e is 4

When the computation is done, total and b are set in localdict with value None.
(the first error is a NameError, it is normal)
When we try to recompute total (cause of NameError), the error is the following : bad operand type for unary +: 'NoneType' so in this case, we don't add it to recompute_queue.
And total is recomputed before b.

To resolve this case, I propose to not append value None in the localdict.

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -651,6 +651,7 @@ class MisReportInstancePeriod(orm.Model):
                     kpi_val_comment = kpi.name + " = " + kpi.expression
                     kpi_eval_expression = aep.replace_expr(kpi.expression)
                     kpi_val = safe_eval(kpi_eval_expression, localdict)
+                    localdict[kpi.name] = kpi_val
                 except ZeroDivisionError:
                     kpi_val = None
                     kpi_val_rendered = '#DIV/0'
@@ -668,7 +669,6 @@ class MisReportInstancePeriod(orm.Model):
                     kpi_val_rendered = kpi_obj.render(
                         cr, uid, lang_id, kpi, kpi_val, context=context)
 
-                localdict[kpi.name] = kpi_val
                 try:
                     kpi_style = None
                     if kpi.css_style:


sbidoul:
@laetitia-gangloff I don't understand your issue very well. What is the symptom for the user?

If the None value is not in the dict, then would you not get another error, but an error nevertheless?

laetitia-gangloff:
In my example, for the user the value for total is ERR instead of 19.

Do you prefer add kpi in the recompute_queue if the error is unknown ?

laetitia-gangloff:
2016-02-11 12:12:55,892 5667 ERROR odoo-test openerp.tools.safe_eval: Cannot eval u'a+b+c'
  File "/home/odoo-test/instance/release/parts/openerp/openerp/tools/safe_eval.py", line 288, in safe_eval
    return eval(test_expr(expr, _SAFE_OPCODES, mode=mode), globals_dict, locals_dict)
  File "", line 1, in <module>
NameError: name 'a' is not defined
2016-02-11 12:12:55,894 5667 ERROR odoo-test openerp.tools.safe_eval: Cannot eval u'd+e'
  File "/home/odoo-test/instance/release/parts/openerp/openerp/tools/safe_eval.py", line 288, in safe_eval
    return eval(test_expr(expr, _SAFE_OPCODES, mode=mode), globals_dict, locals_dict)
  File "", line 1, in <module>
NameError: name 'd' is not defined
2016-02-11 12:12:55,895 5667 ERROR odoo-test openerp.tools.safe_eval: Cannot eval u'a+b+c'
  File "/home/odoo-test/instance/release/parts/openerp/openerp/tools/safe_eval.py", line 288, in safe_eval
    return eval(test_expr(expr, _SAFE_OPCODES, mode=mode), globals_dict, locals_dict)
  File "", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'

image

mis_builder/models/mis_builder.py
```diff
@@ -668,7 +668,9 @@ def _compute(self, cr, uid, lang_id, c, aep, context=None):
                     kpi_val_rendered = kpi_obj.render(
                         cr, uid, lang_id, kpi, kpi_val, context=context)
 
-                localdict[kpi.name] = kpi_val
+                if kpi_val is not None:
+                    localdict[kpi.name] = kpi_val
```
sbidoul: It makes sense, indeed. I wish I'd understand why it works on 8.0.

Instead of this test, I'd move  `localdict[kpi.name] = kpi_val` before the `except ZeroDivisionError`

sbidoul: or in the `else` block just before the render()

laetitia-gangloff:
@sbidoul  as you suggest I move the line to feed localdict when kpi_val is know (before ZeroDivisionError).

In 8.0, the error is "always" ValueError :
https://github.com/odoo/odoo/blob/8.0/openerp/tools/safe_eval.py#L332

In 7.0, it is the original Exception, that it is send :
https://github.com/odoo/odoo/blob/7.0/openerp/tools/safe_eval.py#L295

sbidoul:
Okay. All is explained now. Thanks!

:+1:

adrienpeiffer:
:+1:

PR #159, opened 2016-02-11
[FIX] check if the line have an amount currency instead of the linked…

vrenaville:
I check if the line have a amount currency instead of the currency of the linked account.

### DIFF
--- a/account_financial_report_webkit_xls/report/open_invoices_xls.py
+++ b/account_financial_report_webkit_xls/report/open_invoices_xls.py
@@ -355,7 +355,7 @@ class open_invoices_xls(report_xls):
                      cumul_balance, style_line_decimal)]
 
         if _p.amount_currency(data):
-            if line.get('amount_currency'):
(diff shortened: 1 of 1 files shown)


vrenaville:
If you create invoice in company currency and foreign currency on the same account, you will not be able to print the currency balance.

pedrobaeza:
Isn't needed to fallback to the currency of the account?

vrenaville:
@luc-demeyer Could you please review this PR
thanks for your help

vrenaville:
@pedrobaeza I don't think it's necessary,(but I can be wrong :) ) if you specified a currency on a account all the related move line must have the fields currency and amount_currency set.

pedrobaeza:
OK, I think so also

:+1:

max3903:
:+1:

luc-demeyer:
@vrenaville 
Thanks for correcting this bug.
:+1: 
I think this PR is ready for merge.

PR #160, opened 2016-02-13
Update open_invoices.py

luc-demeyer:
Fix 'Group Partner by currency' option.

### DIFF
--- a/account_financial_report_webkit/report/open_invoices.py
+++ b/account_financial_report_webkit/report/open_invoices.py
@@ -173,6 +173,8 @@ class PartnersOpenInvoicesWebkit(report_sxw.rml_parse,
             ledger_lines[account.id] = ledger_lines_memoizer.get(account.id,
                                                                  {})
             if group_by_currency:
+                account.ledger_lines = ledger_lines_memoizer.get(account.id,
+                                                                 {})
                 self._group_lines_by_currency(account)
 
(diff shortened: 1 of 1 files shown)


account_financial_report_webkit/report/open_invoices.py
```diff
@@ -173,6 +173,8 @@ def set_context(self, objects, data, ids, report_type=None):
             ledger_lines[account.id] = ledger_lines_memoizer.get(account.id,
                                                                  {})
             if group_by_currency:
+                account.ledger_lines = ledger_lines_memoizer.get(account.id,
```
pedrobaeza, line 176: Why don't you use `ledger_lines[account.id]`?

luc-demeyer:
@pedrobaeza 
Your suggestion would require a bit more code to fix it:
- the set_context method doc says
      """Populate a ledger_lines attribute on each browse record that will
         be used by mako template"""
      My fix just implements what has been the intention of the original author.
- The '_group_lines_by_currency' method expects the ledger_lines attributes on the browse account. Hence we should modify the logic in this method and pass the ledger_lines dict to this method.
  Just adding the attribute to the account_br record (as it was intended) is less work.

pedrobaeza:
Well, I'm saying that because 2 lines above the same value is assigned to that variable, but never mind, it's not so important.

PR #162, opened 2016-02-18
[FIX] account_financial_report_webkit: Don't overpass top level account

pedrobaeza:
With current code, the top level accounts (the 0 one) is ignored in
the population of debit, credit and balance data, provoking that the
report shows a line with all 0's, which leads to confusion to users.

I see no reason for zeroing this data, as it can be significant for
certain account charts, and can be easily ignored for those who not.

### DIFF
--- a/account_financial_report_webkit/report/common_balance_reports.py
+++ b/account_financial_report_webkit/report/common_balance_reports.py
@@ -288,8 +288,6 @@ class CommonBalanceReportHeaderWebkit(CommonReportHeaderWebkit):
         balance_accounts = dict.fromkeys(account_ids, False)
 
         for account in objects:
-            if not account.parent_id:  # hide top level account
-                continue
             if account.type == 'consolidation':
                 to_display_accounts.update(
                     dict([(a.id, False) for a in account.child_consol_ids]))


fclementic2c:
I do not get want you mean by 'provoking that the
report shows a line with all 0's'. 
Do you have a printscreen ?

pedrobaeza:
This is what I mean:

sin nombre

sbidoul:
:+1:

rafaelbn:
:+1:

fclementic2c:
This behaviour is logic in my view. Why would you want to see this line ?

pedrobaeza:
As said, in some chart accounts this data is relevant. For example, in Spain.

fclementic2c:
The top level account (spanish or not) groups P&L account and Balance sheet account so obviously it is zero unless your report is wrong.
I think it is ugly to display it but I will not fight on such a detail.

pedrobaeza:
It depends on the account filter you have applied. Please don't block something that doesn't hurt.

sbidoul:
I think I remember a customer who once asked why debit and credit were 0 on that line. That's why I  voted :+1:

fclementic2c:
...so we agree Stephane !
I like the way the C2C Trial Balance is displayed right now. This line simply does not appear.
image

Sorry but I do not agree to merge this

sbidoul:
@fclementic2c  Indeed it may be better to remove the root line completely.

pedrobaeza:
@fclementic2c, that line is useful as summary when you filter out some accounts and you want to see global balances. Don't restrict to your use cases. These reports are too global to decide yourself if this is good or not.

pedrobaeza:
I prefer to have it the line, but with the balance properly calculated. If it's a full balance sheet, the result will be 0. If not, you will have the real balance.

fclementic2c:
But Pedro, you acan only select view accounts, right ? So if you select the view account 'assets' you already have this total.
It is only if you select several view accounts that it may have some interest. But selecting several accounts on a trial balance takes more time that printing the full report straight.

I am wondering if we should not simply limit the selection of 1 account max (instead of a M2M field in the 2nd tab like now).

Pedro. Merge it we already spent to much time on this (+ in v9 it is another story on top of that)

pedrobaeza:
Thanks for merging and understanding me!

As we say, v9 have to be our focus for discusing how to implement the reports.

PR #163, opened 2016-02-22
8.0 fix overlapping text

jbaudoux:
[FIX] account_financial_report_webkit: in landscape layout, row are split accross pages with half the line displayed on page1 and second half displayed on page2 and overlapping on the row header.
This make data unreadable.
Adding page-break-inside: avoid; on tr solves the issue.
It happens with wkhtmltopdf 0.12.3 (with patched qt)
See https://github.com/wkhtmltopdf/wkhtmltopdf/issues/1524

### DIFF
--- a/account_financial_report_webkit/data/financial_webkit_header.xml
+++ b/account_financial_report_webkit/data/financial_webkit_header.xml
@@ -73,6 +73,7 @@ body, table, td, span, div {
 }
 .act_as_row  {
     display: table-row;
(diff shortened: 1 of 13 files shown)


pedrobaeza:
For 9.0 branch? I think you have pointed to the wrong branch

jbaudoux:
You're right. Closing this one and creating
https://github.com/OCA/account-financial-reporting/pull/164

PR #165, opened 2016-03-04
[IMP] account_financial_report_webkit: Allow to select account level on trial balance

pedrobaeza:
This PR allows to select an account level to filter accounts by this criteria, which is very common here in Spain. The default behaviour is to leave it blank, so there's no different behaviour in this case. If you select one level, all the remaining accounts selected from the rest of the criteria will be filtered out if their account level is higher than the selected one.


jjscarafia:
Same need in Argentina, :+1:  functional review!

elicoidal:
LGTM :+1:

sbidoul:
LGTM :+1:

PR #167, opened 2016-03-06
mis_builder: better layout of mis report settings form on small screens

sbidoul:
image

### DIFF
--- a/mis_builder/views/mis_builder.xml
+++ b/mis_builder/views/mis_builder.xml
@@ -29,7 +29,9 @@
                     <group col="2">
                         <field name="name"/>
                         <field name="description"/>
-                        <field name="query_ids">
+                    </group>
+                    <group string="Queries">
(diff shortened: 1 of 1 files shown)


adrienpeiffer:
:+1: (Functionnal test & Code review)

laetitia-gangloff:
:+1:

sbidoul:
I also improved the layout of the template form so it's not overflowing the sheet anymore.

image

adrienpeiffer:
@sbidoul it seems that the form template still overlaps the sheet. But I don't know if this problem can be solved. So :+1:

sbidoul:
@adrienpeiffer there is not much more we can do indeed. `web_sheet_full_width` helps. And when we'll add more features (styles, sub-kpis) we'll create popups to edit KPI's instead of having an editable list.

But that's for the future. In the meantime, the screen is nicer with this PR anyway.

lepistone:
👍 thanks!

PR #169, opened 2016-03-07
[IMP] mis_builder: display blank cells instead of 0.0 when there is no data

ThomasBinsfeld:
In KPI values computed on accounts where there are no moves during the period, show a blank instead of 0.0. This is nicer for the eye, and sometimes mandatory in some kind of reports.

This is implemented with an AccountingNone singleton which is a null value that "dissolves" in basic arithmetic operations with non-None values.


sbidoul:
:+1: (code review and functional test)

@ThomasBinsfeld can you please fill and send your individual CLA as explained here https://odoo-community.org/page/cla

sbidoul:
@jbeficent @alexis-via you may want to look into this one

ThomasBinsfeld:
@sbidoul ICLA is signed and sent :email:

alexis-via:
yes, better to put 0.0 instead of bank field for those kind of reports which are mainly financial :+1:

sbidoul:
@alexis-via I'm not sure about your last sentence. To be clear:
- before this PR, KPI with no move lines in the period are displayed 0.0
- with this PR: KPI with no move lines in the period are displayed as an empty cell

This is important because:
- it distinguishes clearly KPI with no activity from KPI which have a 0.0 ending balances (for instance)
- in some cases (eg Luxemburg offical reporting) this distinction is mandatory

adrienpeiffer:
:+1: (Functionnal test & Code review)

laetitia-gangloff:
Is it normal than comparison doesn't work with AccountNone value ? (compare with)

Before :
image

After :
image

sbidoul:
@laetitia-gangloff good catch

sbidoul:
@laetitia-gangloff it should be ok, now.

laetitia-gangloff:
It seems not ok in all cases (percentage) :

image

I don't know if it is ok to get +0 if there is no move (difference)?

image

sbidoul:
@laetitia-gangloff p1 vs p3 is blank, that is normal because (-4960 - 0)/abs(0) does not make sense. p3 vs p1 is (0-(-4960))/abs(4960) = 100%. So that part is ok.

For the second part, we can hide the +0 indeed. I'll look into it.

sbidoul:
@laetitia-gangloff 0 comparisons now show blank too.

laetitia-gangloff:
:+1:

sbidoul:
Ok, ready to merge.

adrienpeiffer:
:+1: Again ! Thanks !

pedrobaeza:
What is the added value of using AccountingNone object instead of embedding the logic? I see that at the end, you have to add the clause `if x is AccoutingNone` in some `if` clauses.

sbidoul:
> What is the added value of using AccountingNone object instead of embedding the logic?

@pedrobaeza because the logic would be complex and would need to be repeated in user-entered kpi expressions.

The tests for AccountingNone are at the "boundaries" only (when rendering or returning the value to clients).

pedrobaeza:
OK then

:+1:

PR #179, opened 2016-04-09
[IMP] mis_builder: refactor the compute method

sbidoul:
The main computation entry point is now on mis.report instead of
mis.report.instance.period so it can easily be reused by other
modules which need to evaluate mis.report's without creating
full-fledged persistent mis.report.instance (such as a wizard
that allows the user to select the period for a one-off run).

Beside shuffling code around, there is no change in the logic.

@thomasbinsfeld can you see if https://github.com/acsone/l10n-luxemburg/pull/4 can be simplified with this as we expect?

cc/ @adrienpeiffer

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -412,6 +412,7 @@ class MisReport(models.Model):
         localdict.update(self._fetch_queries(
             date_from, date_to, get_additional_query_filter))
 
+        additional_move_line_filter = None
         if get_additional_move_line_filter:
             additional_move_line_filter = get_additional_move_line_filter()
         aep.do_queries(date_from, date_to,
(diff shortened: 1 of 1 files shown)


adrienpeiffer:
Great ! :+1: (Code review and functional test)

mis_builder/models/mis_builder.py
```diff
@@ -284,6 +284,206 @@ class MisReport(models.Model):
...
+            'len': len,
+            'avg': _avg,
+            'AccountingNone': AccountingNone,
+        }
+
+        localdict.update(self._fetch_queries(
+            date_from, date_to, get_additional_query_filter))
+
+        if get_additional_move_line_filter:
+            additional_move_line_filter = get_additional_move_line_filter()
```
ThomasBinsfeld, line 417: If `get_additional_move_line_filter` is None, `additional_move_line_filter` will be _referenced before assignement_

sbidoul, line 417: Well spotted. Fixed.

sebastienbeau:
tested, ok with the concept

PR #182, opened 2016-04-26
new module to manage financial reports in OCA

lepistone:


lepistone:
We are wondering if it's better to start with an empty one like this or if we want to rename `account_financial_report_webkit`.

Ideas?

pedrobaeza:
I think it's better to start from scratch, and maybe copy/paste some code (crediting authors) if needed. The architecture of account_financial_report_webkit is very complicated for the task per se.

lepistone:
I agree with you pedro: starting with a rename is a can of worms, and we don't get something working quickly. Thanks!

astirpe:
I agree, could we have it merged soon, even if it's just a skeleton? This way we all can start to check it out.

pedrobaeza:
@astirpe, you don't need to merge it. You can clone just right now:

```
git clone https://github.com/damdam-s/account-financial-reporting -b 9.0-sorrento-ledger
```

lepistone:
@pedrobaeza we decided to have a special common branch for the sprint, the target of this PR, and all use this! Not a real merge to 9.0 of course.
This will help us all work together at the same time.

pedrobaeza:
Yeah, of course, you can create a branch and make all code sprint participants member of it, so everyone can push to it. Good choice!

lepistone:
Great! I'm merging this so others pick it up. cc @OpenCode @andrea4ever @CiroBoxHub

PR #185, opened 2016-04-26
more fields

lepistone:


lepistone:
I think we want to merge that quick. Then we'll switch to a shared branch as suggested by @sbidoul.

lepistone:
btw good stuff! 👍

damdam-s:
:+1:

account_financial_report_qweb/wizard/ledger_report_wizard.py
```diff
@@ -11,3 +11,24 @@ class LedgerReportWizard(models.TransientModel):
...
+    date_from = fields.Date()
+    date_to = fields.Date()
+    target_move = fields.Selection([('posted', 'All Posted Entries'),
+                                    ('all', 'All Entries')],
+                                   string='Target Moves',
+                                   required=True,
+                                   default='posted')
+    account_ids = fields.Many2many(
+        comodel_name='account.account',
+        relation='account_account_ledger_rel',
```
pedrobaeza: You don't need to specify relation attribute

lepistone:
merged manually into `oca-sorrento/push-me`.

PR #195, opened 2016-05-25
mis_builder: fix typo, causes crash if report_xslx is missing

lepistone:
ping @sbidoul @lmignon @adrienpeiffer 

Thanks!

### DIFF
--- a/mis_builder/report/mis_report_instance_xlsx.py
+++ b/mis_builder/report/mis_report_instance_xlsx.py
@@ -17,7 +17,7 @@ try:
 except ImportError:
     _logger.debug("report_xlsx not installed, Excel export non functional")
 
-    class ReportXlsx:
+    class ReportXlsx(object):
         pass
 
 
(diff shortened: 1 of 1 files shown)


lepistone:
WIP: still crashes for another reason.

lepistone:
It now probably works, but it is a bit hacky.

I am letting @sbidoul and the others authors of the module choose what you prefer. Thanks!

EDIT: removed a section "other options" that didn't make sense.

sbidoul:
We depend on report_xlsx already, it's just that the module is imported because of the `static` directory.

I'm fine with your approach. 

:+1:

lepistone:
Ah yes @sbidoul, I was making a mess because I didn't have `reporting-engine` available, but still we need to handle `static`. So I agree with you, and my patch is OK like this. Thanks!

sbidoul:
Ok. I think I can fast track this.

PR #198, opened 2016-05-27
[FIX] iter_cell_tuples

ThomasBinsfeld:
In KpiMatrixCol, in iter_cell_tuples : fix the return sentence.


ThomasBinsfeld:
@sbidoul @adrienpeiffer

adrienpeiffer:
:+1: of course

sbidoul:
Yes, this method is not used by mis_builder itself but it's needed for dependent module which reuse the computation algorithm.

:+1:

sbidoul:
Fast tracking as it is not used by mis_builder and it's an obvious typo.

PR #199, opened 2016-05-27
[IMP] mis_builder: Update localsdict in declare_and_compute_period 

ThomasBinsfeld:
Localsdict can be updated when declare_and_compute_period is called.

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -904,14 +904,16 @@ class MisReport(models.Model):
                                             query argument and returns a
                                             domain compatible with the query
                                             underlying model
-        :param locals_dict: personalized localsdict
+        :param locals_dict: personalized locals dictionary used as evaluation
+                            context for the KPI expressions
         """
         self.ensure_one()
 
         # prepare the localsdict
         if locals_dict is None:
             locals_dict = {}
-            locals_dict.update(self.prepare_localsdict())
+
+        locals_dict.update(self.prepare_localsdict())
 
         # fetch non-accounting queries
         locals_dict.update(self._fetch_queries(
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -797,7 +797,7 @@ class MisReport(models.Model):
         aep.done_parsing()
         return aep
 
-    def prepare_localsdict(self):
+    def prepare_locals_dict(self):
         return {
             'sum': _sum,
             'min': _min,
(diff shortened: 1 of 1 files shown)


ThomasBinsfeld:
@sbidoul @adrienpeiffer

sbidoul:
I would prefer to name the argument locals_dict (optional prefilled locals dictionary used in evaluating the kpis).

Then it would go like this:
- if locals_dict is None: locals_dict = {}
- locals_dict.update(self.prepare_locals_dict())

Where prepares_locals_dict returns the current one (with sum, AccountingNone, etc).

adrienpeiffer:
I agree with @sbidoul

ThomasBinsfeld:
More elegant, indeed. I'm on it.

mis_builder/models/mis_report.py
```diff
@@ -892,18 +904,14 @@ def declare_and_compute_period(self, kpi_matrix,
                                             query argument and returns a
                                             domain compatible with the query
                                             underlying model
+        :param locals_dict: personalized localsdict
```
sbidoul: personalized locals dictionary used as evaluation context for the KPI expressions

mis_builder/models/mis_report.py
```diff
@@ -892,18 +904,14 @@ def declare_and_compute_period(self, kpi_matrix,
...
-            'max': _max,
-            'len': len,
-            'avg': _avg,
-            'AccountingNone': AccountingNone,
-            'SimpleArray': SimpleArray,
-        }
+        # prepare the localsdict
+        if locals_dict is None:
+            locals_dict = {}
+            locals_dict.update(self.prepare_localsdict())
```
sbidoul: I would do this update outside of the `if`, as I see no use case for not having sum/min/max/AccountingNone etc in the locals dict.

sbidoul:
:+1:

adrienpeiffer:
Thanks @ThomasBinsfeld :+1:

mis_builder/models/mis_report.py
```diff
@@ -797,6 +797,17 @@ def prepare_aep(self, company):
         aep.done_parsing()
         return aep
 
+    def prepare_localsdict(self):
```
sbidoul: @ThomasBinsfeld Oh, one last small thing: can you rename to `prepare_locals_dict` (with the underscore). Since we work a lot by grepping it's always better to be very consistent when naming things.

sbidoul:
Fixed typo squashed and merged at https://github.com/OCA/account-financial-reporting/commit/9af17587fc872f1f630bcf78bf10c38bdcc55e0d

Issue #200, opened 2016-05-29
8.0 Crash in account_financial_report_webkit "execv() arg 2 must contain only strings"


alexis-via:
I had this crash when clicking on the "Print" button of a Trial balance in a company that has an accentuated character in its name:

```
2016-05-29 23:02:23,672 1981 INFO absamad werkzeug: 192.168.10.201 - - [29/May/2016 23:02:23] "POST /web/dataset/call_kw/account.journal/read HTTP/1.0" 200 -
2016-05-29 23:02:24,022 1981 WARNING absamad openerp.pooler: openerp.pooler.get_db_and_pool() is deprecated.
2016-05-29 23:02:24,070 1981 WARNING absamad openerp.pooler: openerp.pooler.get_db_and_pool() is deprecated.
2016-05-29 23:04:16,417 1981 INFO absamad financial.reports.webkit: DEBUG ALEXIS command=['/usr/local/bin/wkhtmltopdf', '--quiet', '--encoding', 'utf-8', '--margin-top', '17.0', '--margin-bottom', '15.0', '--orientation', 'Portrait', '--page-size', 'A4', '--header-font-name', 'Helvetica', '--footer-font-name', 'Helvetica', '--header-font-size', '10', '--footer-font-size', '6', '--header-left', u'BALANCE DES COMPTES - Monast\xe8re Sainte-Madeleine - EUR', '--header-spacing', '2', '--footer-left', '30/05/2016 01:02:24', '--footer-right', u'Page [page] de [topage]', '--footer-line', '/tmp/tmpRth7M90.body.html', '/tmp/webkit.tmp.QPX1Ic.pdf']
2016-05-29 23:04:16,451 1981 ERROR absamad openerp.service.report: Exception: execv() arg 2 must contain only strings
Traceback (most recent call last):
  File "/home/openerp/erp/odoo/openerp/service/report.py", line 93, in go
    result, format = openerp.report.render_report(cr, uid, ids, object, datas, context)
  File "/home/openerp/erp/odoo/openerp/report/__init__.py", line 40, in render_report
    return registry['ir.actions.report.xml'].render_report(cr, uid, ids, name, data, context)
  File "/home/openerp/erp/odoo/openerp/api.py", line 256, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/openerp/erp/odoo/openerp/addons/base/ir/ir_actions.py", line 156, in render_report
    return new_report.create(cr, uid, res_ids, data, context)
[...]

PR #201, opened 2016-05-29
8.0 Fix bug #200 "execv() arg 2 must contain only strings"

alexis-via:
This is a possible fix for bug #200 Crash in account_financial_report_webkit "execv() arg 2 must contain only strings" 

It is certainly not the best solution to the problem, but at least it fixes the issue.
You need to have the unidecode lib installed


feketemihai:
@alexis-via I will close one since no activity for a long time. Please reopen it if still need it.

PR #203, opened 2016-05-31
add an option in partner balance for non-zero only

lepistone:
This options exists in the partner balance from the account module. It
was not there in the webkit module though. The naming is the same as
from the odoo module.

A very old commit message for the webkit report mentions this
functionality but apparently it is not there.

By default all partners are shown as before. This option works for both
PDF and XLS reports.

### DIFF
--- a/account_financial_report_webkit/report/common_partner_balance_reports.py
+++ b/account_financial_report_webkit/report/common_partner_balance_reports.py
@@ -23,6 +23,8 @@
 from collections import defaultdict
 from operator import add
 
+from openerp.tools.float_utils import float_is_zero
+
 from .common_balance_reports import CommonBalanceReportHeaderWebkit
 from .common_partner_reports import CommonPartnersReportHeaderWebkit
 
@@ -83,9 +85,11 @@ class CommonPartnerBalanceReportHeaderWebkit(CommonBalanceReportHeaderWebkit,
                     details[partner_id].get('credit', 0.0)
 
             if display_partner == 'non-zero_balance':
-                details = {k: v
-                           for k, v in details.iteritems()
-                           if abs(v['balance']) > 0.0001}
+                details = {
+                    k: v
+                    for k, v in details.iteritems()
+                    if not float_is_zero(v['balance'], precision_digits=5)
+                }
             res[account_id] = details
 
         return res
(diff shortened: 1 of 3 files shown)


account_financial_report_webkit/report/common_partner_balance_reports.py
```diff
@@ -80,6 +81,11 @@ def _get_account_partners_details(self, account_by_ids, main_filter,
                     get('init_balance', 0.0) + \
                     details[partner_id].get('debit', 0.0) - \
                     details[partner_id].get('credit', 0.0)
+
+            if display_partner == 'non-zero_balance':
+                details = {k: v
+                           for k, v in details.iteritems()
+                           if abs(v['balance']) > 0.0001}
```
mdietrichc2c: Why not use the `float_is_zero` util command?

account_financial_report_webkit/wizard/partner_balance_wizard.py
```diff
@@ -1,63 +1,51 @@
...
+        "Partner's", required=True, default='customer_supplier')
+    partner_ids = fields.Many2many(
+        'res.partner', string='Filter on partner',
+        help="Only selected partners will be printed. "
+        "Leave empty to print all partners.")
+
+    # same field in the module account
+    display_partner = fields.Selection(
+        [
+            ('non-zero_balance', 'With balance is not equal to 0'),
```
mdietrichc2c: "With non-zero balance"

mdietrichc2c:
A few comments, but :+1: on the rest

lepistone:
Thanks for the review @mdietrichc2c. I addressed your comments.

mdietrichc2c:
:+1:

account_financial_report_webkit/report/common_partner_balance_reports.py
```diff
@@ -80,6 +83,13 @@ def _get_account_partners_details(self, account_by_ids, main_filter,
                     get('init_balance', 0.0) + \
                     details[partner_id].get('debit', 0.0) - \
                     details[partner_id].get('credit', 0.0)
+
+            if display_partner == 'non-zero_balance':
+                details = {
+                    k: v
+                    for k, v in details.iteritems()
+                    if not float_is_zero(v['balance'], precision_digits=5)
```
damdam-s, line 91: Why `precision_digits=5` and not `precision_rounding` according to `Account` precision ?

damdam-s:
A question (non blocking) but :+1:

account_financial_report_webkit/report/common_partner_balance_reports.py
lepistone, line 91: In my mind this is not meant to be the business concept of rounding.

Instead I really want to compare with zero, so that is just a technical workaround for handling floats, and 5 digits is an arbitrarily high precision.

jcoux:
I tested this option for my project and it works correctly : :+1:

gurneyalex:
:+1:

PR #207, opened 2016-06-22
Fix comparison between balances

mdietrichc2c:
Small PR to fix an issue in 8.0 comparison reports : the account's current balance was used instead of the computed one (for example, if you filtered on some periods).


sbidoul:
:+1:

fclementic2c:
:+1:

pedrobaeza:
:+1:

PR #213, opened 2016-08-23
[8.0] Fix account.move.line field last_rec_date

ghost:
In previous implementation of the field with the api 7.0, the field was not always computed at the right moment.
This would cause last_rec_date to be null. The drawback is that the amounts on the aged partner balance may be incorrect.
For example, if you have an invoice on May and a payment on June, the amount would not appear on the partner
aged balance of May.


account_financial_report_webkit/i18n/account_financial_report_webkit.pot
```diff
@@ -1507,7 +1507,7 @@ msgstr ""
```
pedrobaeza: You can remove this file as it's no longer needed.

pedrobaeza:
Thanks for this improvement. Minor nitpick, but :+1:

ghost:
@pedrobaeza I improved the migration script and removed the pot file.

account_financial_report_webkit/__openerp__.py
```diff
@@ -169,8 +169,12 @@
 the header and footer are created as text with arguments passed to
 wkhtmltopdf. The texts are defined inside the report classes.
 """,
-    'version': '8.0.1.1.0',
```
elicoidal, line 172: Please remove the descriptioon key and input the content in README

elicoidal:
small non blocking detail.
:+1:

ghost:
@elicoidal Thanks for the review. I moved the description to README.

Issue #57, opened 2015-02-03
How to handle accented characters in report names?


sbidoul:
Example: when printing the Aged Partner Balance in french, "Balance âgée des tiers" causes an execv() exception or is misprinted. The exact behaviour seems to depend on the locale under which the server runs.

This is due to this title being passed as a --header-left argument to wkhtmltopdf.

Does anyone knows if there is some best practice in this respect?

ghost:
Hi @sbidoul 

Were you able to fix the errors without changing the translation?

sbidoul:
@gs-clearcorp no progress. But I'm pretty sure it's related to the server environment locale.

adrienpeiffer:
Hi @gs-clearcorp,
After several investigation, I think put `LANG="en_US.UTF-8"` as an environment variable should solve the problem. Could you give us your feedback if you test it ?

ghost:
Hi @sbidoul , @adrienpeiffer I tried to fix the error with LANG=en_US.utf-8 and es_CR.utf-8  without success.

adrienpeiffer:
It's strange. 
My scenario :  
I launched my server like this `LANG=en_US ./odoo.py` and the execv() is raised. After that I launched my server like this `LANG=en_US.UTF-8 ./odoo.py` and the initial exception isn't raised.

ghost:
I was able to solve the error

Had to add <i>env LANG=en_US.utf-8</i> to the start-stop daemon script.

luisfqba:
Work for me. I run odoo on Docker Container and when I added env LANG=en_US.utf-8 webkit run fine. Thanks.

antoniocanovas:
Working for me with LANG=es_ES.utf-8
Thanks

alexis-via:
@sbidoul Do you consider the solution above as the "real" fix, or as a sort of hack/workaround ?

When I read this comment in the code:
https://github.com/OCA/account-financial-reporting/blob/8.0/account_financial_report_webkit/report/webkit_parser_header_fix.py#L51

it confirms what I think: it's really dirty to use arguments to a command line to pass this kind of information with accentuated characters.

sbidoul:
@alexis-via Putting .utf-8 in the LANG env var is the best workaround I've seen so far. 

But I agree a better solution should come from somewhere upstream (probably in wkhtmltopdf).

alexis-via:
@sbidoul ok, we have the same point of view.

[...]

PR #214, opened 2016-08-24
[Fix] #57 Fixed accented characters in report names

martini97:


pedrobaeza:
This fix is not valid, as it replaces the accentuated characters. You need to set the correct language on the terminal.

martini97:
As @alexis-via said, using arguments in the command line to fix this is dirty
"""
@sbidoul Do you consider the solution above as the "real" fix, or as a sort of hack/workaround ?

When I read this comment in the code:
https://github.com/OCA/account-financial-reporting/blob/8.0/account_financial_report_webkit/report/webkit_parser_header_fix.py#L51

it confirms what I think: it's really dirty to use arguments to a command line to pass this kind of information with accentuated characters.
"""

sbidoul:
@martini97 I'm not sure everyone talks about the same thing when using the word "command line argument" here.

What is dirty is using arguments to pass such long textual information such as header and footer, I agree. However, this is due to the way wkhtmltopdf works, there is now way around it, as far as I know. So we have to live with.

Configuring the environment variable to have Odoo run in a proper unicode environment is clean and robust. IMO, it is more robust than trying to fiddle and convert encodings in the code.

That's why I'm :-1: for this PR for now.

ovnicraft:
Close this ?

PR #211, opened 2016-07-26
OCA financial reports

jcoux:
PR for OCA financial reports.

It's based on PR https://github.com/OCA/account-financial-reporting/pull/196 which has been started at Sorrento code sprint.

The content description is available at this URL :
http://bit.ly/29XNpec

What was done :
- General Ledger (PDF / XLSX)
- Trial Balance (PDF / XLSX)
- Open Items (PDF / XLSX)
- Aged Parner Balance (PDF / XLSX)

PR #219, opened 2016-09-07
Performance improvement for OCA accounting reports

jcoux:
This PR complete the PR https://github.com/OCA/account-financial-reporting/pull/211 with a performance improvement for OCA accounting reports.
- Deletion of joins between tables with "OR" clauses
- Deletion of useless data generation of General Ledger report for Trial Balance report
- Customization of xlsxwriter options to limit usage of memory for xlsx files generation

### DIFF
--- a/account_financial_report_qweb/report/abstract_report_xlsx.py
+++ b/account_financial_report_qweb/report/abstract_report_xlsx.py
@@ -3,6 +3,8 @@
 # Copyright 2016 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from cStringIO import StringIO
+import xlsxwriter
 from openerp.addons.report_xlsx.report.report_xlsx import ReportXlsx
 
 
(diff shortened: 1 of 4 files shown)


account_financial_report_qweb/report/abstract_report_xlsx.py
```diff
@@ -3,6 +3,8 @@
 # Copyright 2016 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from cStringIO import StringIO
+import xlsxwriter
```
pedrobaeza, line 7: Why this dependency here? It should be in other module

jcoux: Hi @pedrobaeza 

I have added the `cStringIO` and `xlsxwriter` dependencies because I have override the `create_xlslx_report` method for "Customization of xlsxwriter options to limit usage of memory for xlsx files generation".

See lines 38 to 51 of this file.

pedrobaeza: But then the previous PR was incorrect. We can't mix XLSX generation with PDF generation. That should be optional through a module  account_financial_report_qweb_xlsx

jcoux: @pedrobaeza , you're right on this point.
But it's not the subject of this PR.
I will try to do another PR for that, but I don't know when I could do that.

pedrobaeza: OK, thanks

fclementic2c:
2015 Trial balance (PDF) out in 3 min pdf on a 6.2 millions account move lines database... much better

:+1:

fclementic2c:
and...balance calculation validated ! 
kudos @jcoux for your great work !
Now we have financial reports in version 9 : fast and reliable :)

cc @jgrandguillaume

sbidoul:
@fclementic2c @jcoux @jgrandguillaume out of curiosity, would you have a chance to time a trial balance with mis_builder on the same database? Or this method: https://github.com/OCA/account-financial-reporting/blob/f9d637d8bca733ec980b8aebc69ab28dc9302db8/mis_builder/models/aep.py#L413

gurneyalex:
worked with julien on this. :+1:

fclementic2c:
@sbidoul I ll do the test on the same DB and PC and I will tell you.

For now we need this merged so let's do it.

sbidoul:
@fclementic2c sure, I did not intend to slow this PR. Pure curiosity, nothing else.

Thanks for finishing this, BTW!

PR #222, opened 2016-09-14
[FIX] Fix debit formula on partner balance xls report

hugosantosred:
Fixes the debit col position on this report when is computed with the date filter.

In the cases that the date filter is set the initial balance is not shown but the debit and credit columns position are the same because of the span set in partner name column.

In the balance formula the column position changes if initial balance is present or not, so the formula is wrong in this cases, because is using the partner reference column as the start column in the formula instead of the debit column.

### DIFF
--- a/account_financial_report_webkit_xls/report/partners_balance_xls.py
+++ b/account_financial_report_webkit_xls/report/partners_balance_xls.py
@@ -368,17 +368,15 @@ class partners_balance_xls(report_xls):
                              partner_ref if partner_ref else '')]
                 if _p.comparison_mode == 'no_comparison':
                     bal_formula = ''
+                    debit_col = 4
                     if _p.initial_balance_mode:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do you remove the case when there's initial balance?

hugosantosred:
Is the same column in both cases. Because when the initial balance is set the partner name column span is set to 2 in order to include the initial balance column, if the initial balance is not set the partner name column span is set to 3

https://github.com/factorlibre/account-financial-reporting/blob/e45882112e7c0193b94c64f5dc5a96596cd443b8/account_financial_report_webkit_xls/report/partners_balance_xls.py#L359-L366

Because of this the debit column is in the same position in both cases.

pedrobaeza:
OK, thanks for the explanation :+1:

feketemihai:
@hugosantosred I will close one since no activity for a long time. Please reopen it if still need it.

PR #225, opened 2016-09-16
Fix initial balance of general_ledger_xls

ghost:
The initial balance row is not aligned with the rest of the report. It is 1 column off to the left.


liebana:
I think this is duplicated from #222

luc-demeyer:
@liebana 
it's not a duplicate of  #222

@dufresnedavid 
I tested the patch and it works ok.
Hence green light for merge.

feketemihai (approved):
Ok

pedrobaeza (approved):

PR #224, opened 2016-09-16
Qweb: add a counterpart accounts column

ghost:
I wasn't aware of [this](https://github.com/OCA/account-financial-reporting/blob/9.0/account_financial_report_webkit/report/common_reports.py#L558) at first, I don't know what's preferable. I felt like doing it the Python way because I knew about `lru_cache`.

PR #227, opened 2016-09-20
Qweb: honor translated account names

ghost:
The skeleton of the report is available in French and `account.account.name` is a translatable field, so it should be translated as well. We've had issues due to that at Open Net Sàrl before. I based my SQL additions on [the request in ir.translation itself](https://github.com/odoo/odoo/blob/216d328a11f254a3c037eb9e23ab362b984ca0ba/openerp/addons/base/ir/ir_translation.py#L403-L432), and it can fall back on displaying the untranslated value.

NOTE: I haven't added my name and my company's to the README because it would likely conflict with #224 later. In the event that my other pull request was rejected, I would like [lbeopennet:9.0-translated-copyright](https://github.com/open-net-sarl/account-financial-reporting/tree/9.0-translated-copyright) to be integrated in order for the copyright notice to be updated. In any other case, please disregard.

### DIFF
--- a/account_financial_report_qweb/report/general_ledger.py
+++ b/account_financial_report_qweb/report/general_ledger.py
@@ -413,7 +413,7 @@ SELECT
     NOW() AS create_date,
     a.id AS account_id,
     a.code,
-    a.name,
+    COALESCE(tr.value, a.name) AS name,
     COALESCE(i.debit, 0.0) AS initial_debit,
     COALESCE(i.credit, 0.0) AS initial_credit,
     COALESCE(i.balance, 0.0) AS initial_balance,
@@ -427,6 +427,11 @@ LEFT JOIN
     initial_sum_amounts i ON a.id = i.account_id
 LEFT JOIN
     final_sum_amounts f ON a.id = f.account_id
+LEFT JOIN
+    ir_translation tr ON a.id = tr.res_id
+        AND tr.lang = %s
+        AND tr.type = 'model'
+        AND tr.name = 'account.account,name'
 WHERE
     (
         i.debit IS NOT NULL AND i.debit != 0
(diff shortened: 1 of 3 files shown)


account_financial_report_qweb/report/general_ledger.py
```diff
@@ -492,6 +497,7 @@ def _inject_account_values(self):
         query_inject_account_params += (
             self.id,
             self.env.uid,
+            self.env.lang,
```
pedrobaeza, line 500: Shouldn't this be `self.env.context.get('lang', self.env.lang)`?

pedrobaeza:
About the copyright, your change is very little (a bug with its patch) to be considered as a co-author/contributor. If you make more changes or contributions, then it can be considered.

pedrobaeza:
Anyways, your name will appear and counts as contribution in git commit history and in the statistics we are preparing at OCA, so every little thing counts!

account_financial_report_qweb/report/general_ledger.py
ghost: That would be redundant given that [`self.context.lang`](https://github.com/odoo/odoo/blob/9.0/openerp/api.py#L802) is itself nothing more than an alias for `self.context.get('lang')`.

pedrobaeza: Cool! I didn't know about that "alias"

pedrobaeza (approved):

ghost:
Yes you're right, I'm fine with that. Thank you very much!

dcoopennet:
LGTM

damdam-s:
:+1: thanks

Garamotte:
LGTM :+1:

jcoux:
Thanks you !
Bug fix is ok for the General Ledger Report.
But if you can do it also on the Open Items Report (similar SQL request), the bug will be fixed on all reports.
:+1: with that

ghost:
hey thanks @jcoux, didn't think of that. I'll do it today.

ghost:
Done.

ghost:
(while I'm at it)

atchuthan:
code review :+1:

PR #228, opened 2016-09-21
8.0 financial_report_webkit: Better Partner Ledger + Open Invoices report

alexis-via:
I improved Partner Ledger and Open Invoices Report to allow the selection of all account types (not just receivable and/or payable) and I added a filter on accounts.

I also renamed the object account.aged.trial.balance.webkit to aged.trial.balance.webkit to make it shorter.
Postgres limits the length of table and index names to 63 characters cf https://til.hashrocket.com/posts/8f87c65a0a-postgresqls-max-identifier-length-is-63-bytes.
This causes a problem with a M2M field from account.aged.trial.balance.webkit to account.account (for account filter) because the names of the 2 indexes of the M2M table would be:
- account_account_account_aged_trial_balance_webkit_rel_account_aged_trial_balance_webkit_id_index
- account_account_account_aged_trial_balance_webkit_rel_account_account_id_index

The labels are longer than 63 characters and the 63 first caracters are identical:

account_account_account_aged_trial_balance_webkit_rel_account_a (63 chars)

Bad luck: the first different caracters is character 64 !
That's why I decided to reduce the size of the object name 'account.aged.trial.balance.webkit'.

### DIFF
--- a/account_financial_report_webkit/report/aged_partner_balance.py
+++ b/account_financial_report_webkit/report/aged_partner_balance.py
@@ -421,7 +421,7 @@ class AccountAgedTrialBalanceWebkit(PartnersOpenInvoicesWebkit):
         return dict((x[0], x[1]) for x in res)
 
 HeaderFooterTextWebKitParser(
-    'report.account.account_aged_trial_balance_webkit',
+    'report.account.aged_trial_balance_webkit',
     'account.account',
     'addons/account_financial_report_webkit/report/templates/\
                                                     aged_trial_webkit.mako',
--- a/account_financial_report_webkit/report/common_reports.py
+++ b/account_financial_report_webkit/report/common_reports.py
@@ -84,6 +84,8 @@ class CommonReportHeaderWebkit(common_report_header):
             return _('Payable Accounts')
         elif val == 'customer_supplier':
             return _('Receivable and Payable Accounts')
+        elif val == 'all':
+            return _('All Account Types')
         else:
             return val
 
--- a/account_financial_report_webkit/report/open_invoices.py
+++ b/account_financial_report_webkit/report/open_invoices.py
@@ -71,6 +71,7 @@ class PartnersOpenInvoicesWebkit(report_sxw.rml_parse,
             'amount_currency': self._get_amount_currency,
             'display_partner_account': self._get_display_partner_account,
             'display_target_move': self._get_display_target_move,
+            'accounts': self._get_accounts_br,
             'additional_args': [
                 ('--header-font-name', 'Helvetica'),
                 ('--footer-font-name', 'Helvetica'),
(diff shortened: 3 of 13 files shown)


pedrobaeza:
Instead of changing the table name, just use `relation` attribute of the many2many field to set manually the many2many table, because this change in an stable version can't be as there are modules that can inherit the renamed table.

alexis-via:
@pedrobaeza That's what I did first by defining the 'account_ids' M2M field that way in account_financial_report_webkit/wizard/partners_ledger_wizard.py:

```
'account_ids': fields.many2many('account.account', 'partners_ledger_webkit_account_rel',
'wizard_id', 'account_id', string='Filter on accounts')
```

but then I get another technical issue : when I run the "Open Invoices" report (which inherits the partner ledger report), I get this error:

```
2016-09-21 19:30:33,108 15946 ERROR adldemo1 openerp.sql_db: bad query:  INSERT INTO partners_ledger_webkit_account_rel (wizard_id, account_id)
                        (SELECT 1, unnest(ARRAY[433])) EXCEPT (SELECT wizard_id, account_id FROM partners_ledger_webkit_account_rel WHERE wizard_id=1)

  File "/home/odoo/erp/odoo-work/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
IntegrityError: insert or update on table "partners_ledger_webkit_account_rel" violates foreign key constraint "partners_ledger_webkit_account_rel_wizard_id_fkey"
DETAIL:  Key (wizard_id)=(1) is not present in table "partners_ledger_webkit".
```

and I didn't find any way to solve this. This problem doesn't happen when you define the field as I did in my PR:

```
'account_ids': fields.many2many('account.account', string='Filter on accounts')
```

pedrobaeza:
I think you just need to redefine the field `account_ids` with the same technique in the new open invoices model. Have you tried?

alexis-via:
@pedrobaeza No. But I don't need to re-define the field in open.invoices.webkit if I define the field in partners.ledger.webkit with the syntax I use in this PR:

```
'account_ids': fields.many2many('account.account', string='Filter on accounts')
```

and that's how it is done in all the account_financial_report_webkit module. FYI, there are up to 4 levels of inheritance in the financial report wizard ; for example : account.common.partner.report -> partners.ledger.webkit -> open.invoices.webkit -> aged.open.invoices.webkit (and also aged.trial.balance.webkit)
And the fields of account.common.partner.report are hopefully not re-defined at every level of inheritance.

pedrobaeza:
OK, if nobody opposes to this model renaming, go ahead.

CasVissers-360ERP (approved):
Functional test done.

pedrobaeza:
This PR has conflicts.

PR #229, opened 2016-09-23
[FIX][account_financial_report_qweb] Fix ImportError.

yajo:
It was happening when `xlsxwriter` python module was not available.

@Tecnativa

### DIFF
--- a/account_financial_report_qweb/README.rst
+++ b/account_financial_report_qweb/README.rst
@@ -50,6 +50,7 @@ Contributors
 * Francesco Apruzzese <opencode@e-ware.org>
 * Lorenzo Battistini <lorenzo.battistini@agilebg.com>
 * Julien Coux <julien.coux@camptocamp.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
 
 Much of the work in this module was done at a sprint in Sorrento, Italy in
 April 2016.
--- a/account_financial_report_qweb/__openerp__.py
+++ b/account_financial_report_qweb/__openerp__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'category': 'Reporting',
     'summary': 'OCA Financial Reports',
     'author': 'Camptocamp SA,'
@@ -35,8 +35,11 @@
         'report/templates/trial_balance.xml',
         'view/account_view.xml'
     ],
-    'test': [
-    ],
+    'external_dependencies': {
+        "python": [
+            "xlsxwriter",
+        ],
+    },
     'installable': True,
     'application': True,
     'auto_install': False,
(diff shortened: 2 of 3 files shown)


account_financial_report_qweb/README.rst
```diff
@@ -50,6 +50,7 @@ Contributors
 * Francesco Apruzzese <opencode@e-ware.org>
 * Lorenzo Battistini <lorenzo.battistini@agilebg.com>
 * Julien Coux <julien.coux@camptocamp.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza, line 53: This contribution is very little to include you here. In fact, it's a patch.

pedrobaeza (changes requested):

pedrobaeza:
Bump version number

sbidoul:
Please, wrap the import xslxwriter at a lower level. 

Note that I've never had an import error since I'm using pip to properly install the addons I need.

account_financial_report_qweb/README.rst
yajo: Yep, but it's mine :blush:

pedrobaeza: git blame will know that, but here we put only relevant contributors, or the list would be big for all the modules.

yajo: This seems important to me. :stuck_out_tongue:

account_financial_report_qweb/__openerp__.py
```diff
@@ -35,8 +35,11 @@
         'report/templates/trial_balance.xml',
         'view/account_view.xml'
     ],
-    'test': [
-    ],
+    'external_dependencies': {
+        "python": [
+            "xlsxwriter",
+        ],
+    },
```
sbidoul, line 42: AFAICS xslxwriter is not a direct dependency of afr_qweb. It is a dependency of report_xslx, so it should not be declared here.

yajo: I updated the patch and put the try lower in the stack. Please review that, you probably missed that.

sbidoul: @Yajo indeed I had not seen that part https://github.com/OCA/account-financial-reporting/blob/58c559b59fd0d28dc4ee9c824365c2c579b491ff/account_financial_report_qweb/report/abstract_report_xlsx.py#L38

IMO it is not the best way to do it because this override of `create_xlsx_report` does not call super(). @jcoux I think reporting-engine must be extended to support options.

pedrobaeza: And it should be split in 2 modules: one for the QWeb part, and the other for XLSX support extension.

yajo: That's the job for another PR. :smile:

yajo:
This one should be fast tracked IMHO.

yajo:
Ping! :blush:

yajo:
:zzz:

pedrobaeza (approved):

pedrobaeza:
@sbidoul any feedback on this?

sbidoul:
@pedrobaeza @Yajo I created https://github.com/OCA/account-financial-reporting/issues/241. I'd prefer to get that fixed instead of this PR. What do you think?

pedrobaeza:
OK, but if the solution is not still done, I would vote for merging this meanwhile

PR #232, opened 2016-09-27
[IMP] Add refresh button in mis report preview. Widget code changes to allow to add fields in the widget easier

adrienpeiffer:

PR #168, opened 2016-03-06
[8.0][mis_builder][ADD] mis_builder_analytic_filter

sbidoul:
Add analytic filtering on mis_builder reports. The filter is static (in the MIS Report instance configuration form) but also dynamic (in the preview and dashboard view) so the user can change analytic account on the fly.
- [x] Depends on #151 

image

PR #233, opened 2016-09-27
[9.0][mis_builder][ADD] Add mis_builder_analytic_filter

adrienpeiffer:
An idea from https://github.com/OCA/account-financial-reporting/pull/168
- [x] Depends on #232


sbidoul:
@adrienpeiffer can you check lint errors?

PR #235, opened 2016-10-06
[FIX] mis_builder: crash in xslx with one column

### LINKED ISSUE #204: [9.0] mis_builder col_width-related crash in xslx export when there is only one column
when i test with mis_builder (9.0), i found some error at line no. 134 of "mis_report_instance_xslx.py".
show the following.

 data_col_width = min(MAX_COL_WIDTH, max(*col_width.values()))
TypeError: 'int' object is not iterable

when i printed out show "col_width.values() is [10]" and not working.

help me some solution.

Thank you.

sbidoul:
fixes #204


adrienpeiffer:
:+1:

adrienpeiffer (approved):

lmignon (approved):

pedrobaeza (approved):

PR #239, opened 2016-10-11
Migration of mis_builder to 10.0

adrienpeiffer:
- [x] Forward fix/improvement in version 9.0

~- [x] Depends on https://github.com/OCA/web/pull/439~
~- [x] Depends on https://github.com/OCA/reporting-engine/pull/76~


sbidoul:
@adrienpeiffer can you look at the test failure?

adrienpeiffer:
@sbidoul Travis is green.

sbidoul (approved):

sbidoul:
This one is working fine, should be good for merging.

mis_builder/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'MIS Builder',
-    'version': '9.0.2.0.2',
+    'version': '10.0.2.0.2',
```
pedrobaeza, line 7: 10.0.1.0.0

sbidoul: @pedrobaeza since the 10.0 and 9.0 are identical (except very minor changes in the source code) we plan to maintain feature parity between the two versions for some time. And we would like to use the version number to express the feature level of each series. So in this case I would like to have 10.0.2.0.3.

Do you have any objection for that scheme?

pedrobaeza: Well, you will have the same conflicts cherry-picking/merging commits from previous branch, and it's not the regular norm, but you are the maintainer of this module, so you can decide in this.

sbidoul (approved):
Rebased and green. Works fine in production.

alexis-via (approved):

PR #240, opened 2016-10-11
Migration of account_financial_report_qweb to 10.0

adrienpeiffer:
- [x] Depends on https://github.com/OCA/reporting-engine/pull/76

### DIFF
--- a/account_financial_report_qweb/__manifest__.py
+++ b/account_financial_report_qweb/__manifest__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'category': 'Reporting',
     'summary': 'OCA Financial Reports',
     'author': 'Camptocamp SA,'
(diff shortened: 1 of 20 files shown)


pedrobaeza:
We must split first the XLSX export from the main module in 9.0 before migration.

adrienpeiffer:
@pedrobaeza is there some discussion about this split ?

pedrobaeza:
Yeah, I said it in https://github.com/OCA/account-financial-reporting/pull/219#r77776645 (when I tracked the problem)

sbidoul:
If there are no volunteers right now to do this split, I suggest we migrate as is. Splitting can be done in another PR.

yvaucher (approved):

jgrandguillaume (approved):
Hi,

Thanks for the work, as Stéphane, I think we should merge it like this, the split can be done in a second time..

:+1: 

Regards,

Joël

sbidoul:
@adrienpeiffer do you know why travis is red?

alexis-via (approved):

adrienpeiffer:
@sbidoul it's green now :smile:

lmignon (approved):

ovnicraft:
I have a doubt here, in v8 reports allow get general ledger per account without grouping by partner, i cant find the options to obtain this, was removed in v10 ?

ovnicraft:
My fault, its present.

PR #245, opened 2016-10-18
[Fix] #57 Fixed accented characters in report names

robch2342:
I stole this solution to issue #57 from martini97, who doesn't seem to have made a pull request. It solved the issue for our customer.


pedrobaeza:
This is a problem in your system configuration. See https://github.com/OCA/account-financial-reporting/issues/57

PR #247, opened 2016-10-31
[FIX] Full access on mis report style for adviser.

adrienpeiffer:


sbidoul (approved):
Of course. :+1:

pedrobaeza (approved):
Merging now as obvious

PR #248, opened 2016-11-03
9.0 Issue when computing the initial balance with Qweb Report

sebastienbeau:
Hi all
I face an issue when I print my trial balance.

Let's give an example
If you add the following move at the date 2015-12-31              

|  ref   | debit  |  credit |  account |
|---|---|---|---|
|  move_1  |  1000 | 0  |  receivable  |  
| move_1   | 0  |   1000 |  incomming |  

And you revers it the 2016-01-01

If you add the following move at the date 2016-01-01

|  ref   | debit  |  credit |  account |
|---|---|---|---|
|  move_2  |  0 | 1000  |  receivable  |  
| move_2    | 1000  |   0 |  incomming |  

We expect to have for initial balance in the receivable account an amount of 1000 but now the amount is 0 because the initial balance include the first day of the fiscal year to compute it.

If you take a look here : https://github.com/OCA/account-financial-reporting/blob/10.0/account_financial_report_qweb/report/general_ledger.py#L281

You will see that for computing the balance we use an "<=". It's work if we want to compute the ending balance because if we want to compute the balance until the end of 2015 we filter with <= 2015-12-31.
But if we want to the initial balance we need to filter with
< 2015-01-01
or
<= 2014-12-31
but not  <= 2015-01-01 because we include in the initial balance the move of the first day.

I have added a test that check the computation of the initial balance, and an ugly patch to show where is the bug.

Now I see two solution for fixing the bug.
One is filtering with < 2015-01-01 and in this case we need to change the way to generate the sub_subquery_sum_amounts
The other solution is to filter with <= 2014-12-31 and in this case we need to do a date - 1 day when we filter with the self.date_from

What do you think @jcoux @fclementic2c @simahawk

### DIFF
--- a/account_financial_report_qweb/report/general_ledger.py
+++ b/account_financial_report_qweb/report/general_ledger.py
@@ -263,7 +263,9 @@ class GeneralLedgerReportCompute(models.TransientModel):
         # Refresh cache because all data are computed with SQL requests
         self.refresh()
 
-    def _get_account_sub_subquery_sum_amounts(self, include_initial_balance):
+    def _get_account_sub_subquery_sum_amounts(
+            self, include_initial_balance, date_included
+    ):
         """ Return subquery used to compute sum amounts on accounts """
         sub_subquery_sum_amounts = """
             SELECT
@@ -278,9 +280,15 @@ class GeneralLedgerReportCompute(models.TransientModel):
             INNER JOIN
                 account_move_line ml
                     ON a.id = ml.account_id
-                    AND ml.date <= %s
         """
-
+        if date_included:
+            sub_subquery_sum_amounts += """
+                    AND ml.date <= %s
+            """
+        else:
+            sub_subquery_sum_amounts += """
+                    AND ml.date < %s
+            """
         if not include_initial_balance:
             sub_subquery_sum_amounts += """
                 AND at.include_initial_balance != TRUE AND ml.date >= %s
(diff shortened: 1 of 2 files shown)


jcoux:
Hello @sebastienbeau , I will look your PR next week.

jcoux:
Hello @sebastienbeau ,

You're right about this issue.

The first day of the chosen period must not used to compute the initial balance.

But the issue is present in 3 places :
- https://github.com/OCA/account-financial-reporting/blob/9.0/account_financial_report_qweb/report/general_ledger.py#L286
- https://github.com/OCA/account-financial-reporting/blob/9.0/account_financial_report_qweb/report/general_ledger.py#L520
- https://github.com/OCA/account-financial-reporting/blob/9.0/account_financial_report_qweb/report/general_ledger.py#L1199

To fix it, it will be better to not change the sql query (or with a parameter passed to the functions), but to add a variable with the date of the day preceding the chosen first day and to change the variable in the query parameters.

If you want, I can fix that with your PR as base.

sebastienbeau:
@jcoux indeed changing the day to the preceding the chosen first day is perfect.
You can reuse my branch (at least the test) to fix it.

When you PR is ready tell me I will test it

sebastienbeau:
@jcoux do you plan to work on it?

fclementic2c:
@sebastienbeau ok it is planned 1st week of 2017.

sebastienbeau:
@fclementic2c  thanks

jcoux:
@sebastienbeau I work on this today.

jcoux:
@sebastienbeau I have done a PR with your branch : 
https://github.com/akretion/account-financial-reporting/pull/15
You can review it and merge it if that's good for you.

pedrobaeza:
Cherry-picked commit from @jcoux

pedrobaeza (approved):
I confirm this solves the issue.

sebastienbeau:
Looking at the code it's seem to be ok for me.
@florian-dacosta can you test it with our customer data?

florian-dacosta:
I have tested the report on our customer data and it works fine.
(I compared the balance of each account between a report from version 7, a report from version 9 without the fix and a report from version 9 with the fix.)
With the fix, initial balances from v9 and v7 are coherent.
:+1:

PR #251, opened 2016-11-22
Add context to get language in accounts names according user's language

idealisconsultingbe:


feketemihai:
@idealisconsultingbe Do you still need this one for v8?

PR #254, opened 2016-12-04
[REF] .travis.yml: Enable build for new runbot travis2docker structure

moylop260:


pedrobaeza:
Better to put new global environments matrix

sbidoul:
I think this is resolved.

PR #259, opened 2016-12-30
8.0 add XLS report to Aged Partner Balance

luc-demeyer:
This PR covers the following:
- adds xls export to the Aged Partner Balance report
- Fix so that reports are not broken on last and first day ofthe fiscal year
- a minor change to the trial balance xls report to facilitate customisation via inherit
- a couple of code style changes to align with the OCA guidelines

PR #260, opened 2016-12-31
fix Aged Partner Balance, _get_current_fiscalyear

luc-demeyer:
This method has always been wrong (date_start & date_stop inverse logic) but on dec 31st, this error becomes visible via python dump.


luc-demeyer:
Closing since this fix is now included in PR https://github.com/OCA/account-financial-reporting/pull/259.

PR #267, opened 2017-01-30
[9.0][FIX] fix qweb financial reports takes company from user not from wizard

jjscarafia:
Step to reproduce:
1. User working on a parent company (parent company don't have date ranges defined)
2. User request general ledger of a child company
3. Error with empty "fy_start_date"

This happens because when computing "self.fy_start_date" the user company is used instead of the company of the wizard.


pedrobaeza (approved):

gonzaloruzafa (approved):
+1

alexis-via (approved):

PR #269, opened 2017-02-09
[9.0] Issue on unaffected earning account balance with qweb reports

florian-dacosta:
I think there is still a bug with the balance computation for the special account 'unaffected earning'.

When I compute the balance for the fiscal first year, it is all good, but from second fiscal year and after, I have a problem with the inital/final balances.

I made a test to show the bug. It should fail with the present v9 branch, but it should pass with the fix of this PR.

@jcoux @fclementic2c Could you take a look please?

### DIFF
--- a/account_financial_report_qweb/report/general_ledger.py
+++ b/account_financial_report_qweb/report/general_ledger.py
@@ -1231,9 +1231,17 @@ WHERE id = %s
     ):
         """ Return subquery used to compute sum amounts on
         unaffected earnings accounts """
-        sub_subquery_sum_amounts = """
+        if not include_initial_balance:
+            sub_subquery_sum_amounts = """
+        SELECT
+            -SUM(ml.balance) AS balance
+            """
+        else:
+            sub_subquery_sum_amounts = """
         SELECT
             SUM(ml.balance) AS balance
+            """
+        sub_subquery_sum_amounts += """
         FROM
             account_account a
         INNER JOIN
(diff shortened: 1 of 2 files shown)


fclementic2c:
@florian-dacosta Sorry to come back so late on this but better late than never :)

I see that move lines on CY earning account do not have the correct sign but it seems that this fix does correct the issue...

To clarify : see the use case as I understood it:
https://docs.google.com/spreadsheets/d/1ePDjHAAs8MMNrL9gcSrTMg9ccY-OU9hSoNgcvPcEfaw/edit#gid=0

cc: @jcoux

florian-dacosta:
@fclementic2c Yes, better late than never, I am glad you are confirming the bug!

So, your use case is similar to the one of the test n°4 I have added in this PR.
I juste made a quick test on an empty v9 database, doing exactly what is described in the google spreadsheets you have joined, and, without this fix, I find the wrong value (-46 for the final balance of the unaffected earning account for 2018) And, simply by merging my PR and restarting the server, the balance is fixed, I find -26.

I am surprised when you say "(ERROR IS : today we have -46 here even with this fix : https://github.com/OCA/account-financial-reporting/pull/269/files)"

fclementic2c:
@florian-dacosta Ok we will double check (next week). Thanks for confirming this point.

jcoux:
@florian-dacosta Ok for me : tested locally : it's work ! :+1:

fclementic2c:
:+1:

PR #275, opened 2017-02-18
[10.0] mis_builder: version 3.0

sbidoul:
This PR started with the goal of adding the data source concept to serve as a basis for mis_buidler_budget. It has become the incubator for mis builder version 3.0 with many new features.

See [CHANGELOG](https://github.com/acsone/account-financial-reporting/blob/10.0-mis_builder-3-sbi/mis_builder/CHANGES.rst).

- [x]  KpiData abstract class, to prepare for manual entry of values for some kpis (eg budgeted values)
- [x] handle `accumulation_method` in `mis.kpi.data` query method so budgeting periods do not need to match the reporting periods exactly, and compute budget theoretical amounts pro-rata temporis
- [x] alternative model for move lines (any model that has date, debit, credit, account_id), configurable as a data source at the column level; so you can have one column showing actuals, and other columns showing the same kpi's on alternative data: think committed purchases, etc
- [x] a comparison data source (compare two other columns): similar to the (now deprecated) comparison columns mechanism, but more flexible in the sense you can place the comparison columns anywhere now
- [x] implement `sumcol` column source to have columns that are sum (+/-) of other columns
- [x] `mis_buidler_budget` implements the `mis_budget` column source
- [x] hook for analytic filtering in budgets
- [x] basic workflow for `mis.budget`
- [x] `name_search()` for `mis.report.kpi.expression`, so budget items can be imported
- [x] implement `_add_mis_columns_budget` in subkpi situations
- [x] domains and constraints on `sumcol` and `cmpcol` source information
- [x] in report column definition view, display data source before date period, because some data sources require no dates
- [x] budget drilldown
- [x] refactor `prepare_and_compute_period`, there must be a simpler variant for manual kpis such as budgets
- [x] style property to hide empty rows
- [ ] ~~**multi company support**~~ to be done in another PR (see issue #295)
- [ ] ~~**basic multi currency support**~~ to be done in another PR (see issue #350)
- [ ] ~~make `mis_budget(_items)` read-only in non-draft state~~ added to the roadmap
- [ ] ~~for completeness, handle negative search operators in `kpi.expression.name_search()`~~ added as TODO in the code, not very important
- [ ] ~~migration script for comparison_columns to the new comparison source mechanism?~~ added as a caveat in the changelog, this is very hard to automate in a reliable way and very easy to recreate the comparison columns manually
- [x] mis_builder_budget README
- [x] **more tests**
- [x] changelog


sbidoul:
@adrienpeiffer

JordiBForgeFlow:
When I create a MIS Report template where one KPI is has subkpi's and another does not have, like here:

image

It results in an error message when the MIS Report is printed:

`Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/addons/base/ir/ir_qweb/qweb.py", line 315, in _compiled_fn
    return compiled(self, append, values, options, log)
  File "<template>", line 1, in template_527_1403
  File "<template>", line 3, in body_call_content_1402
  File "<template>", line 4, in foreach_1401
  File "<template>", line 5, in body_call_content_1400
  File "/root/build/OCA/account-financial-reporting/mis_builder/models/mis_report_instance.py", line 567, in _compute_matrix
    self._add_column(aep, kpi_matrix, period, period.name, description)
  File "/root/build/OCA/account-financial-reporting/mis_builder_budget/models/mis_report_instance.py", line 61, in _add_column
    aep, kpi_matrix, period, label, description)
  File "/root/build/OCA/account-financial-reporting/mis_builder/models/mis_report_instance.py", line 542, in _add_column
    aep, kpi_matrix, period, label, description)
  File "/root/build/OCA/account-financial-reporting/mis_builder/models/mis_report_instance.py", line 506, in _add_column_actuals
    period._get_additional_query_filter)
  File "/root/build/OCA/account-financial-reporting/mis_builder/models/mis_report.py", line 1222, in declare_and_compute_period
    locals_dict, eval_expressions, eval_expressions_by_account)
  File "/root/build/OCA/account-financial-reporting/mis_builder/models/mis_report.py", line 1089, in _declare_and_compute_col
    raise UserError(_("Probably not your fault... but I'm "
UserError: (u"Probably not your fault... but I'm really curious to know how you managed to raise this error so I can handle one more corner case!", None)

Error to render compiling AST
UserError: (u"Probably not your fault... but I'm really curious to know how you managed to raise this error so I can handle one more corner case!", None)
Template: 527
Path: /templates/t/t/t/t/t[1]
Node: <t t-set="matrix" t-value="o._compute_matrix()"/>`

Perhaps this action does not make sense? In that case, shouldn't it be necessary to prevent users to creating lines with different sub-kpi's across different kpi lines?

sbidoul:
Eh eh, one more corner case.

Ok, what you are trying to do here does not make sense indeed. But it will work if you use `current_assets + ...` as an expression for `total_assets`: it's a useful shortcut to create sum lines in your reports without having to enter the same sum for each subkpi.

JordiBForgeFlow:
@sbidoul Indeed `current_assets + balp[101000]` works, but not `balp[101000]`... strange magic happening behind the scenes... :)

sbidoul:
@jbeficent behind the scene, `current_asset` is a tuple (a SimpleArray actually) with each item a subkpi value. So `current_asset + balp[101000]` is actually adding balp[101000] to each current_asset subkpi.

JordiBForgeFlow:
So perhaps you should convert everything in the expression always to a SimpleArray?

JordiBForgeFlow (changes requested):
We identified issues #283 #282 #281 #280 #279

JordiBForgeFlow:
@sbidoul Do you think that it will be feasible to backport to 9.0 once this one is merged? We can provide further assitance if needed!

sbidoul:
@jbeficent backporting to 9.0 should be trivial. But I need to improve test coverage first.

JordiBForgeFlow:
@sbidoul Do you need help to prepare tests?

sbidoul:
@jbeficent I restarted work on mis_builder 3.0 with the goal of having it merged before Odoo Experience (where we could work on additional documentations and possibly new features).

All the issues you reported should be resolved.

JordiBForgeFlow:
Thanks. We can help you with the review, with @lreficent and @mreficent

sbidoul:
Ok, except a test for mis_builder_budget which I will add very soon, I consider this one complete.

I'd like to merge this soon so we can do another PR for multi-company before the OCA sprint and Odoo Experience.

sbidoul:
@jbeficent, @lreficent, @mreficent

sbidoul:
All green!

Martronic-SA:
Yes all green for me too!

Martronic-SA (approved):
Fine for me

benwillig (approved):
Functionnal test ok!

sbidoul:
Moved to https://github.com/OCA/mis-builder/pull/1

legalsylvain:
Hi @sbidoul. I just saw that mis module moved from account-financial-reporting into mis-builder. (well, two monthes ago !)
This action broke some repositories like l10n-france that are depending on the wrong repository. see https://github.com/OCA/l10n-france/pull/121. 
Maybe other repositories will have the same problem.

In a more general way, I'm not sure it's good operation to change a module from a repository to another if the module is set to instalable to True. It could break some production system based on buildout. Its maybe better to wait a migration. (for exemple, when OCA created 11.0 serie).

regards.

sbidoul:
@legalsylvain Old history is still in the previous repo, so if you froze your version with buildout, there is no impact. The breakage is explicit and trivial to fix when you refresh your repos to the last version. 

The advantage of the move is that I can use git to maintain feature parity between 9 and 10, and hopefully 11.

legalsylvain:
Hi @sbidoul ,
I understand the advantages, I just say that this operation has broken some repository regarding Travis. For the time being  l10n-belgium is still failing.
Regards.

PR #274, opened 2017-02-18
10.0 account financial report qweb fix import

yelizariev:
Applies https://github.com/OCA/account-financial-reporting/pull/229 to 10.0

PR #242, opened 2016-10-14
OCA reports XLSX: Better override to define workbook creation options

jcoux:
Little modification to fix this issue:
https://github.com/OCA/account-financial-reporting/issues/241

Depends of PR:
https://github.com/OCA/reporting-engine/pull/77

This PR contains a revert of https://github.com/OCA/account-financial-reporting/pull/229 which is useless now.

PR #276, opened 2017-02-20
OCA reports XLSX: Better override to define workbook creation options

yelizariev:
applies #242 to 10.0

I have no idea does it work, but it shall fix import issue. If it breaks system someway, consider to merge #274 for a while to prevent import error for ones who has account-financial-reporting in addons path, but don't use ``account_financial_report_qweb`` module


sbidoul:
@yelizariev thanks!

Looks good, as soon as https://github.com/OCA/reporting-engine/pull/98 is merged.

pedrobaeza (approved):

PR #284, opened 2017-03-16
[FIX] account_tax_balance: tax from bank statements is not reported

StefanRijnhart:


StefanRijnhart:
When I assign an (inclusive) tax to a bank statement during reconciliation, the tax is not reported by account_tax_balance because all moves that include a line on a liquidity account are ignored by design.

It seems to me that this can be easily fixed by including all moves in the 'regular' computed fields except for 'refund' type moves.

@eLBati what do you think? Any reason that you excluded these move types in the first place?

astirpe:
For the Dutch Vat Statement reports (eg.: module `l10n_nl_tax_declaration_reporting`) we definitely need this fix. Thanks @StefanRijnhart !

In case the authors of this module do not agree on this fix, there is no problem: the module `l10n_nl_tax_declaration_reporting`, or whatever module inherits this one, could extend method `def get_target_type_list()` and fix it properly to retrieve the needed data.

For me is a :+1: to have this

jvOrsouw:
I'm trying out this module. I agree with removing the constraint on 'internal type = payable , receivable'. It gives an accurate presentation of the taxes posted in the system. 

👍

eLBati (approved):
I am OK with this.

I would like to know what @antespi thinks, as he introduced ```get_target_type_list``` at https://github.com/OCA/account-financial-reporting/commit/cfb0b1322f3dc8b258785261002e5f6b79179606

StefanRijnhart:
@elBati thanks for the clarification. It is a regression due to the redesign then.

pedrobaeza (changes requested):
OK, seems correct also for us, but please change version number.

StefanRijnhart:
@pedrobaeza done

pedrobaeza (approved):

pedrobaeza:
Merged and cherry-pick for v9.

pedrobaeza:
Although the PR doesn't have problems, when merging it test have failed: https://travis-ci.org/OCA/account-financial-reporting/jobs/212446472

StefanRijnhart:
@pedrobaeza So in 9.0, the tax line is still created automatically https://github.com/OCA/account-financial-reporting/pull/286

StefanRijnhart:
Indeed, the default behaviour of applying taxes was changed from 9.0 to 10.0: https://github.com/odoo/odoo/commit/89cbef8540ab0#diff-dd671a54296b170ea1393dca1a5f7798L995

PR #286, opened 2017-03-18
[FIX] tests (tax line is created automatically)

StefanRijnhart:


pedrobaeza (approved):
Thanks for the follow-up. I fast-track it.

Issue #292, opened 2017-03-24
[9.0][10.0] account_tax_balance performance


sbidoul:
I notice account_tax_balance is very slow even when there are few account moves.

A little bit of sql tracing show it repeats this pattern for each tax:
* select all move lines for that tax
* perform additional filtering on move lines

```sql
SELECT "account_move_line_id" FROM "account_move_line_account_tax_rel" WHERE "account_tax_id" IN (383);

SELECT "account_move_line".id FROM "account_move" as "account_move_line__move_id","account_move_line" 
WHERE ("account_move_line"."move_id"="account_move_line__move_id"."id") 
AND (((((("account_move_line"."date" <= '2017-12-31')  
AND  ("account_move_line"."date" >= '2017-01-01'))  
AND  ("account_move_line"."company_id" = 1))  AND  ("account_move_line__move_id"."state" in ('posted')))  
AND  ("account_move_line"."id" in (1,2,3,4,5,6,7,8,9,10,11,12,13,15,16,17,18,19,21,22,23,24,25,26,27,29,30,31,42,43,44,45,46,47,48,60,62,63,64,65,67,68,69,70,71,72,73,74,75,76,77,78,79,80,81,82,83,84,85,86,87,89,90,91,93,94,95,97,98,107,108,109,110,11
1,112,113,115,116,118,119,120,121,122,123,124,129,130,137,138,139,140,141,143,144,145,146,147,149,150,151,152,153,154,155,157,158,159,160,162,163,165,166,167,168,169,170,172,173,175,176,177,178,179,180,181,182,183,184,185,186,187,189,190,191,193,194,195,196,197,198,199,200,201,202,203,204,205,206,207,208,209,210,211,212,213,215,216,217,218,219,220,221,22
2,223,224,225,226,227,228,229,230,231,232,233,234,235)))  
AND  ("account_move_line__move_id"."move_type" in ('receivable_refund','payable_refund'))) 
ORDER BY "account_move_line"."date" DESC,"account_move_line"."id" DESC,
```

I guess it's due to the many2many relation between move lines and taxes which is not optimizable with auto-join.

sbidoul:
@eLBati do you see the same problem? Any idea on how to optimize this?

sbidoul:
And this query pattern is repeated for each tax (we have ~400 in Luxemburg...).

sbidoul:
[...]

PR #293, opened 2017-03-24
[9.0] account_tax_balance performance

sbidoul:
Let the database do some computations.

### DIFF
--- a/account_tax_balance/models/account_tax.py
+++ b/account_tax_balance/models/account_tax.py
@@ -3,8 +3,7 @@
 # © 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import models, fields, api
-from openerp.tools.safe_eval import safe_eval
+from openerp import models, fields, api, _
 
 
 class AccountTax(models.Model):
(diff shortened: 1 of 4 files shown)


sbidoul:
This is a slight improvement, but it does not solve the main cause described in #292 .

account_tax_balance/models/account_tax.py
```diff
@@ -152,16 +153,15 @@ def get_move_lines_domain(self, tax_or_base='tax', move_type=None):
...
     def get_lines_action(self, tax_or_base='tax', move_type=None):
-        move_lines = self.get_move_lines_domain(
+        domain = self.get_move_lines_domain(
             tax_or_base=tax_or_base, move_type=move_type)
-        move_line_ids = [l.id for l in move_lines]
         action = self.env.ref('account.action_account_moves_all_tree')
         vals = action.read()[0]
         vals['context'] = {}
-        vals['domain'] = [('id', 'in', move_line_ids)]
+        vals['domain'] = domain
```
pedrobaeza, line 178: Here it's better to precompute the domain and past again only IDs for avoiding a recomputing each time you filter on this list.

sbidoul: @pedrobaeza even when you potentially have millions of ids?

pedrobaeza: Uhm, not sure now that you put me that example... What is your experience with this?

sbidoul: I'm quite sure it's better to put the domain in the action, otherwise you move around millions of ids and you kill the browser and the database.

pedrobaeza: OK, go ahead then. Thanks for the info. Check Travis status, which is red.

sbidoul: But you need an efficient domain too, which is not the case as explained in #292

pedrobaeza: I'll check it next week to see if something comes to my mind

sbidoul: Thanks Pedro, have a nice week-end. Travis should be green now.

account_tax_balance/models/account_tax.py
```diff
@@ -3,66 +3,28 @@
...
+from openerp import models, fields, api, _
 
 
 class AccountTax(models.Model):
     _inherit = 'account.tax'
 
     balance = fields.Float(
-        string="Total Balance", compute="_compute_balance",
-        search='_search_balance')
+        string="Total Balance", compute="_compute_balance")
```
pedrobaeza: For better readability, instead of removing the comma and put the `)` in the same line, please just remove the `search=...` line and keep the other line. I'm converting also my style habits lately to this one to be more diff-friendly, although it takes more lines:
```
balance = fields.Float(
    string="Total Balance", compute="_compute_balance",
)
```

pedrobaeza (approved):
I have tried the PR on a real installation and get a slight improvement, so this is one is good to be merged.

Only a diff-friendly comment on my side.

About the effective domain, I haven't found any other suggestion, but I'm not having special problem with it on a couple of thousands records.

account_tax_balance/models/account_tax.py
sbidoul: style adapted

adrienpeiffer (approved):
:+1:

PR #297, opened 2017-04-15
[9.0] [account_tax_balance] add init hook to speed up install

JordiBForgeFlow:
The objective of this hook is to speed up the installation of the module on an existing Odoo instance.

Without this script, if a database has a few hundred thousand journal entries, which is not unlikely, the update will take at least a few hours.

### DIFF
--- a/account_tax_balance/hooks.py
+++ b/account_tax_balance/hooks.py
@@ -56,7 +56,8 @@ def store_field_move_type(cr):
         ON acc.id = aml.account_id
         where aml.move_id = am.id
         and acc.internal_type ='payable' 
-        and aml.balance < 0)
+        group by aml.move_id
+        having sum(aml.balance) < 0)
         and am.move_type is null;
         """
     )
@@ -70,8 +71,9 @@ def store_field_move_type(cr):
         INNER JOIN account_account acc
         ON acc.id = aml.account_id
         where aml.move_id = am.id
-        and acc.internal_type ='payable' 
-        and aml.balance >= 0)
+        and acc.internal_type ='payable'
+        group by aml.move_id
+        having sum(aml.balance) >= 0)
         and am.move_type is null;
         """
     )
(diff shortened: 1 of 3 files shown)


account_tax_balance/hooks.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+# © 2017 Eficent Business and IT Consulting Services, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+try:
+    from openerp.addons.mail_tracking.hooks import column_add_with_value
```
pedrobaeza, line 7: Why do you import this?

pedrobaeza:
Lastly, my opinion about hooks that pre-compute fields via SQL that are somehow slow to run with the regular compute methods is that they introduce a possible divergence in the way fields are calculated in one place and the other, or desynchronizations when the module evolves.

Installing a module can be an operation that you can perform in parallel with the system running in production executing another instance with --stop-after-init switch. There are only 2 possible drawbacks:

* A lock on certain tables, that freezes the installation on line `module <module_name>: creating or updating database tables`. I haven't traced exactly the problem on which table is the lock, but killing the installing instance and starting it again solves this.
* A "concurrent update error" while writing the values of the computed fields. Although it's not the usual, this can be trickier to solve, as the computation stops, but the SQL column is created, so if you launch again the installation, the computation is not performed again. You have to manually drop the column in PG, and restart installation.

Thus, I prefer one of these 2 ways instead of providing more and more code:

* Try to optimize the computation method itself. Maybe an SQL inside the compute method that handles the computation at multi-record level. When this method is called at the installation time, `self` variable contains the full recordset and the times will be the same as the hook. The only problem is that this kind of expressions are less readable and maintainable, but the code is the same for all.
* Don't do anything, and just change your deployment method. If you don't face any of the drawbacks, you will see that the consumed time is not so important when done in parallel.

sbidoul:
I agree with Pedro, these init hooks are duplicating the logic.

I think it's possible to make the compute method efficient with one SQL query to get the type and balance (where id in self.id). And then partitioning the resulting set of ids and doing one write per move type.

pedrobaeza:
@jbeficent are you going to work on this?

JordiBForgeFlow:
@pedrobaeza I am not continuing with this work.

pedrobaeza:
OK, closing then.

PR #298, opened 2017-04-18
[9.0][ADD] Account Journal Report

MiquelRForgeFlow:
Journal Report
==============

Adds the following report in PDF and XLSX format:

* Journal Ledger. It groups the moves by their entries.

cc: @pedrobaeza @gboros-rgbconsulting @jbeficent


account_journal_report/README.rst
```diff
@@ -0,0 +1,67 @@
...
+#. In *Application / Accounting & Finance*, select *Accountant* or *Adviser* options.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Invoicing* or *Accounting* menu
+#. Press 'Reporting > PDF Reports > Journal Ledger'
+
```
pedrobaeza: Include a line for indicating that pressing button "Export", you'll get the report on Excel.

account_journal_report/README.rst
```diff
@@ -0,0 +1,67 @@
...
+please check there if your issue has already been reported. If you spotted it
+first, help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jordi Esteve <jesteve@zikzakmedia.com>
```
pedrobaeza: Remove this one, as the code has nothing to do with this original contributor now.

account_journal_report/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009 Zikzakmedia S.L. (http://zikzakmedia.com)
```
pedrobaeza: This file doesn't need this overloaded header.

account_journal_report/__openerp__.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009 Zikzakmedia S.L. - Jordi Esteve
```
pedrobaeza: Remove this copyright.

account_journal_report/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+
+{
+    'name': "Journal Report",
+    'version': "9.0.1.0.0",
+    'depends': [
+        "account",
+        "report_xlsx",
+    ],
+    'license': "AGPL-3",
+    'author': "Zikzakmedia SL, "
```
pedrobaeza: Remove the first author as stated.

account_journal_report/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+              "J. Gutierrez, "
+              "Tecnativa, "
+              "RGB Consulting, "
+              "Eficent, "
+              "Odoo Community Association (OCA)",
+    'website': "http://github.com/OCA/account-financial-reporting",
+    'category': "Accounting",
+    'summary': "PDF Report, XLSX Report",
```
pedrobaeza: Indicate here that is the journal report.

account_journal_report/report/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009 Zikzakmedia S.L. (http://zikzakmedia.com)
```
pedrobaeza: This file doesn't need this overloaded header.

account_journal_report/report/journal_ledger_report.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009 Zikzakmedia S.L. - Jordi Esteve
```
pedrobaeza: Remove this first contributor as stated

account_journal_report/report/journal_ledger_report.py
```diff
@@ -0,0 +1,53 @@
...
+# Copyright 2013 Joaquin Gutierrez (http://www.gutierrezweb.es)
+# Copyright 2015 Tecnativa - Pedro M. Baeza
+# Copyright 2017 RGB Consulting
+# Copyright 2017 Eficent - Miquel Raich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from datetime import datetime
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT
```
pedrobaeza: Don't use this, but `fields.Date.from_string()` and `fields.Date.to_string()`

account_journal_report/report/journal_ledger_report.py
```diff
@@ -0,0 +1,53 @@
...
+
+        date_start = data.get('date_start')
+        date_end = data.get('date_end')
+        journal_ids = data.get('journal_ids', [])
+
+        move_ids = account_move_obj.search(
+            [('date', '<=', date_end),
+             ('date', '>=', date_start),
+             ('journal_id', 'in', journal_ids),
+             ('state', '<>', 'draft')],
```
pedrobaeza: s/<>/!= (the other operator is deprecated)

account_journal_report/tests/test_account_journal_report.py
```diff
@@ -0,0 +1,88 @@
...
+        Tests for Account Journal Report.
+    """
+    def setUp(self):
+        super(TestAccountJournalReport, self).setUp()
+
+        self.res_users_model = self.env['res.users']
+        self.company = self.env.ref('base.main_company')
+        self.g_account_user = self.env.ref('account.group_account_user')
+
+        self.user = self._create_user('user_1', [self.g_account_user],
```
pedrobaeza: For assuring the test, you must provide account data. Please check my tests for v8: https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_account_financial_report/tests/test_journal_entries_report.py, that adds all the needed records.

account_journal_report/tests/test_account_journal_report.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp.tests.common import TransactionCase
+
+
+class TestAccountJournalReport(TransactionCase):
```
pedrobaeza: Use SavepointCase instead, that is more efficient. You need to switch setUp method, but only that. You can see an example here: https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_account_financial_report/tests/test_journal_entries_report.py

account_journal_report/tests/test_account_journal_report.py
```diff
@@ -0,0 +1,88 @@
...
+                                      self.company).id
+
+        self.report_model = \
+            self.env['report.account_journal_report.journal_ledger']
+        self.wiz = self.env['account.journal.entries.report']
+        self.report_name = 'account_journal_report.journal_ledger'
+        self.report_xlsx_name = 'account_journal_report.journal_ledger_xlsx'
+        self.report_title = 'Journal Ledger'
+
+    def _create_user(self, login, groups, company):
```
pedrobaeza: You don't need to create a new user for this, as tests are run as admin, and I see no use of this user for testing any permission.

account_journal_report/tests/test_account_journal_report.py
```diff
@@ -0,0 +1,88 @@
...
+            'email': 'example@yourcompany.com',
+            'company_id': company.id,
+            'company_ids': [(4, company.id)],
+            'groups_id': [(6, 0, group_ids)]
+        })
+        return user
+
+    def test_account_journal_report(self):
+
+        wiz_id = self.wiz.create({})
```
pedrobaeza: Control in your wizard creation data for your test, including the journal and dates you want to print. Check https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_account_financial_report/tests/test_journal_entries_report.py for my test in v8.

pedrobaeza (changes requested):

pedrobaeza:
Please include also full Spanish translation

account_journal_report/wizard/wizard_print_journal_ledger_view.xml
```diff
@@ -0,0 +1,64 @@
...
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">account.journal.entries.report</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">form</field>
+        <field name="view_id" ref="account_journal_ledger_report_view"/>
+        <field name="target">new</field>
+    </record>
+
+    <!-- Add XLSX export button -->
+    <record id="account_journal_ledger_xlsx_view" model="ir.ui.view">
```
geluboros: The export button should be added in the main view, there is no need to creat an extension view for it.

account_journal_report/tests/test_account_journal_report.py
MiquelRForgeFlow: I think it's not necessary because the journal and dates are established by default.

pedrobaeza: Well, it's a safeguard, because you have to make your tests state-consistent, and if you depend on the default values of the wizard, maybe any change can make this fail (for example, a custom module that changes these defaults in a customer).

MiquelRForgeFlow:
@pedrobaeza About the translations, why should I put them? The Transifex solves this, right?

pedrobaeza:
No if you don't translate them on Transifex, and this makes at least one week difference. For initial translation upload, you should provide it with the module.

account_journal_report/tests/test_account_journal_report.py
MiquelRForgeFlow: Ok, the data is added.

account_journal_report/tests/test_account_journal_report.py
```diff
@@ -0,0 +1,117 @@
...
+
+class TestAccountJournalReport(SavepointCase):
+    """
+        Tests for Account Journal Report.
+    """
+    def setUp(self):
+        super(TestAccountJournalReport, self).setUp()
+
+        self.company = self.env.ref('base.main_company')
+        self.g_account_user = self.env.ref('account.group_account_user')
```
pedrobaeza: As a note, this is not needed, but it doesn't make the module fail.

pedrobaeza (changes requested):
There are texts that are not translated yet. Can you please do it?

MiquelRForgeFlow:
All done.

pedrobaeza (approved):

JordiBForgeFlow (approved):
:+1:

PR #303, opened 2017-04-25
[ADD] compute parent_store before running tests

hbrunn:
this one is to fix a problem with tests on runbot (not travis):

Runbot does only one run, with `-i $all_my_module --test-enable`. Then problem is that the tests for this module heavily relies on https://github.com/OCA/OCB/blob/8.0/addons/account/account.py#L278 which in turn relies on the parent order fields being filled, otherwise you'll end up with key errors in https://github.com/OCA/OCB/blob/8.0/addons/account/account.py#L364 (children need to come before their parents, ordering by parent_left guarantees that), as all `parent_left` fields will be empty and the order is undefined

### DIFF
--- a/account_financial_report_webkit/report/common_balance_reports.py
+++ b/account_financial_report_webkit/report/common_balance_reports.py
@@ -23,6 +23,7 @@
 from operator import add
 
 from .common_reports import CommonReportHeaderWebkit
+from openerp import tools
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
This one has been difficult to debug for sure!

hbrunn:
hours of my life I'll never get back indeed, thanks for the fast check!

sbidoul (approved):

PR #316, opened 2017-05-25
[FIX] 10.0 added "tax_exigible" condition in "domain" variable

labaggio:


eLBati (approved):
Thanks

tafaRU (approved):

pedrobaeza:
That check is something that is put on each account.move.line? Shouldn't this be put as a filter on the view instead?

eLBati:
@pedrobaeza field ```tax_exigible``` is used by ```account_tax_cash_basis``` to identify when a move.line (with tax) represent a 'exigible' entry or not.

The use case fixed by this PR:
 - Install ```account_tax_cash_basis```
 - Configure a sale tax with ```Use Cash Basis```
 - Make a customer invoice and pay it
 - Run the wizard of account_tax_balance

The tax move line is included twice and total is wrong (double)

adrienpeiffer (approved):

pedrobaeza (approved):
Thanks for the explanation

PR #326, opened 2017-06-19
[10.0] account_financial_report_qweb: all accounts allowed in Open Items report

alexis-via:
The open items report can also be interesting for accounts with are not "payable" nor "receivable". For example, some companies like to reconcile their VAT accounts and want to use the Open Items report to inspect the unreconciled move lines in their VAT accounts.

This is a similar change as the one I introduced in v8 via this PR https://github.com/OCA/account-financial-reporting/pull/228

### DIFF
--- a/account_financial_report_qweb/report/open_items.py
+++ b/account_financial_report_qweb/report/open_items.py
@@ -201,8 +201,8 @@ WITH
         query_inject_account += """
             WHERE
                 a.company_id = %s
-            AND a.internal_type IN ('payable', 'receivable')
-                    """
+            AND a.reconcile IS true
+            """
         if self.filter_account_ids:
             query_inject_account += """
             AND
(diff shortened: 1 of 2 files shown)


sbidoul:
@alexis-via should it not at least filter on accounts that allow reconciliation ?

alexis-via:
@sbidoul you are right ! I updated my PR.

alexis-via:
@sbidoul are you OK with this PR now ?

sbidoul:
Yes. I was just wondering if it could surprise users to see more accounts displayed by default. Perhaps there should be a radio button to select payable/receivable/other reconcilable? Not sure and this could be another PR.

sbidoul (approved):

adrienpeiffer (approved):

pedrobaeza:
Changing module version number through direct commit with `[skip ci]`

PR #327, opened 2017-06-19
[10.0][MIG] account_journal_report

MiquelRForgeFlow:
Standard migration.


pedrobaeza (approved):

CasVissers-360ERP:
Menu ID is not available in Odoo Enterprise "account.menu_finance_legal_statement", so installation will fail.

pedrobaeza:
@CasVissers why is that menu not available? Odoo Enterprise removes some menus?

CasVissers-360ERP:
@pedrobaeza 
Odoo Enterprise unlinks this menu item because of the reporting tools available in the Enterprise edition. Module that does this is account_extension in Enterprise.

pedrobaeza:
Well, that a very bad practise, but we can do nothing about that... The only hack I can think of is to redeclare the menu here with the same full XML-ID

sbidoul:
In 8.0 we had `account_journal_report_xls` which, despite its name, also did pdf. 

Functionally, it had 2 advantages over this one:
- it has a similar look as the other OCA reports (account_financial_report_webkit)
- it displays tax information

Would it make sens to have this module use the same styles as account_financial_report_qweb, and put it under the same menu entry?

MiquelRForgeFlow:
Yes, that would make sense. But sorry @sbidoul, I won't do that. But I am ready to delete this module if someone does what you propose.  🙂

Meanwhile, please, review and proceed to merge if the module is okay.

AaronHForgeFlow:
LGTM. Pending to the desicion to use this module or port account_journal_report_xls from v8 to v10

sbidoul:
Regarding the choice of module, my point is more to strive to some uniformity in OCA reports, than actually migrating `account_journal_report_xls`.

Do you think you could have the budget/time to just change the look to make it similar to a_f_r_qweb and put it in the same menu (possibly including it in a_f_r_qweb if it's easier to reuse code/css)? 

Adding the tax part could be done by the first which needs it later.

MiquelRForgeFlow:
Let's get review and approve the PR meanwhile. I don't dislike your suggestion @sbidoul, and if someday the stars and planets get aligned, then we may try to do what you propose. 😉

JordiBForgeFlow:
@sbidoul We will add the taxes to the report. We discussed with @mreficent the integration with existing OCA reports, and found the following issues:
* There are differences in the underlying base. OCA reports are based on the standard odoo reports, and this one does not.
* The suffix _qweb would not apply in this case.
* As for adding them under the same menu, we would need to add a dependency with OCA reports. It is a pity to add an extra dependency just for a menu.

sbidoul:
@jbeficent what do you mean with "OCA reports are based on the standard odoo reports, and this one does not."?

Given the current trend to move accounting features to enterprise, I bet the "PDF Reports" menu will disappear from Odoo Community sooner or later. So my thinking is that it's better to put all OCA line-oriented accounting reports in account_financial_report_qweb (which does excel too...). So it's easy to share the parent menu (OCA accounting reports) and CSS styles.

When we have the journal reports in a_f_r_qweb under the same menu, and mis builder for the balance reports, then we can probably hide the PDF reports menu as everything is covered by OCA.

pedrobaeza:
@sbidoul but that will make more difficult to migrate the full module. I prefer to have a common "base" module (with the menu and maybe some helper methods), and then develop each of the reports in a module, so that you can migrate one but not the other if you prefer.

pedrobaeza:
That's for v11. About 10.0, where the main module is already migrated, I would say that this module  should be kept independently, but it can declare a parent menu with the same full XML-ID as the OCA one, and hang in there. This will make the module not to be dependent on the other, and also "compatible". The only trade-off is that if you uninstall the module, the menu will remain, but as parent menus without children are not shown, the problem is very small.

sbidoul:
Splitting a_f_r_qweb is even better of course, perhaps that can be done when migrating to v11. There is not only the menu to share, but also the CSS, [AbstractReportXslx](https://github.com/OCA/account-financial-reporting/blob/10.0/account_financial_report_qweb/report/abstract_report_xlsx.py) and probably more.

JordiBForgeFlow:
@sbidoul Well, I'm sure that OCA Accounting Sprints will now become very popular :) We can include this in the roadmap for v11

sbidoul:
@jbeficent @pedrobaeza we have some budget to work on this, mostly to add the tax summary at the end of each journal.

We would like to try and align this report with the rest of a_f_r_qweb (ie using the same styles). 

Are you ok if we move this code to a_f_r_qweb in the process?

MiquelRForgeFlow:
@sbidoul Thanks, we would be very grateful if you can contribute on this. We will support you.

pedrobaeza:
OK, go ahead.

sbidoul:
As discussed, closing in favor of #366

PR #328, opened 2017-06-20
[8.0] add module mis_builder_budget

luc-demeyer:
Backport (plus simplification) of the mis_builder_budget work done in https://github.com/OCA/account-financial-reporting/pull/275.


sbidoul:
Luc could you preserve the commit history from the 10.0 branch?

The technique explained here https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#technical-method-to-migrate-a-module-from-90-to-100-branch also works for backports.

Then add your backport and simplification commits on top of that.

Thanks!

luc-demeyer:
@sbidoul 
Stephane,
no problem to add the commit history from the 10.0 branch but the instructions on how to do it for 9.0 to 10.0 are like Chinese to me. I am not a git expert and unable to adapt these instructions so that I get the 10.0 commit history into this 8.0 PR.
If you can give me LITERALLY the set of commands to execute, than I'll try to do it, otherwise I prefer to keep this PR as is (your 10 PR is referred hence there is a link to find the 10.0 commit history).

sbidoul:
This PR results in a module that is quite different from the 9.0/10.0 version, therefore I'm not comfortable with merging this because it will be hard to maintain and migrate. 

In https://github.com/OCA/mis-builder/issues/13 I describe what I think is the simplest procedure to backport newest mis_builder features to 8.0, and then make it easy to maintain it across many Odoo series.

luc-demeyer:
@sbidoul 
I agree that it's better to have a backport aligned to the 10.0 functionality.
But than it comes to resources to do this backport while most of us are currently working on 10.0 stuff and soon we'll start on the 11.0 version.
I propose to publish the 8.0 mis_builder_budget module outside the OCA branches so that the functionality is available for the odoo community and remove it from our 8.0 apps branch as soon as we have a proper OCA 8.0 backport.

sbidoul:
@luc-demeyer I understand the ressource issue, of course. if you want to publish it on apps, could you do it with another name than mis_builder_budget to avoid confusion, and perhaps add a note in it's description to explain it's relationship to mis_builder_budget? Thanks!

luc-demeyer:
I prefer to keep the name since I have it running currently on 10+ databases and it will be easier to upgrade those to the OCA backport as soon as we have found funding/resourcing for it. 
I'll add this into the roadmap section the future alignment to the OCA mis_builder_budget (and remove the 'maintained by OCA' section.

sbidoul:
I think that will create a log of confusion, and migration issues for these users.

luc-demeyer:
By keeping the name we just avoid migration issues.
I also don't want to maintain our own 'quick.' backport and hence we'll write a script in the migrations folder of the oca mis_builder_budget 8.0 backport to facilitate the migration. 
Having another module name makes it a lot more difficult.

PR #334, opened 2017-06-30
[10.0][mis_builder] Adds multi bar Chart display choice

Martronic-SA:
in Mis report instance I've added a field to chosse between standard table or bar chart display.
It's using default odoo libraries nv.

TODO : add line chart, pie chart, maybe add a display options to each KPI so it's possible to mix different types of charts together.

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -58,7 +58,7 @@ class KpiMatrixRow(object):
                 self.kpi.auto_expand_accounts_style_id])
         self.chart_display = None
         if self.kpi.report_id.display == 'multi' and \
-                    self.kpi.chart_display:
+                self.kpi.chart_display:
             self.chart_display = self.kpi.chart_display
 
     @property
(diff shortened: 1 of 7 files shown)


sbidoul:
Interesting. Can you paste a screenshot?

Martronic-SA:
here a sample (amounts on left are partially hidden for confidentiality)

Martronic-SA:
Now the line chart 👍

sbidoul:
Nice. Not sure these chart are generic enough, though. Perhaps this should go in an extension module.

Have you considered adding a column displaying [sparklines](https://en.wikipedia.org/wiki/Sparkline) with data of other selected columns?

Martronic-SA:
Added multichart support:
Sample below comparision Revenu/Expenses/EBITA 2016 and 2017 per month and cumulative on year

adrienpeiffer:
As MIS Builder module hase been moved in https://github.com/OCA/mis-builder, I'll close this one

PR #337, opened 2017-07-06
[9.0][mis_builder] add self, date_from and date_to on mis_builder localcontext

jjscarafia:
Add self, date_from and date_to varaibles on local context for more flexibility on kpis design.

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -809,6 +809,9 @@ class MisReport(models.Model):
             'max': _max,
             'len': len,
             'avg': _avg,
+            'time': time,
+            'datetime': datetime,
+            'dateutil': dateutil,
             'AccountingNone': AccountingNone,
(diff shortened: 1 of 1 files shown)


sbidoul:
Merged @ https://github.com/OCA/mis-builder/pull/1/commits/0918b6fad621e91769e52b97f4d0c9d60cf45d0a

In https://github.com/OCA/mis-builder/pull/1/commits/c26c0572a53d64eebcfac47937978f362cdee43f I also used date types instead of string for these variables as I feel it's more immediately useful. 

I did not merge the `self` thing however, because `self` as a variable name does not mean much in the expression evaluation context. I'm considering adding the `mis.report.instance.period` in the context but I'd like to think for more use cases before doing it. What was your use case for adding self?

cc/ @Martronic-SA I kind of remember you contributed a similar patch somewhere.

Thanks for your patience and contribution.

PR #338, opened 2017-07-12
removes  version="7.0"

Martronic-SA:
as it breaks save buttons in Odoo 9.0


sbidoul:
merged @ https://github.com/OCA/mis-builder/pull/1/commits/294b95fbbe3f7548a309b529ef55f742b8e117c5

sbidoul:
@Martronic-SA how do you want to be referenced in the [contributors file](https://github.com/acsone/mis-builder/blame/10.0/mis_builder/README.rst#L100-L113)?

Martronic-SA:
you can add me in the list of contributors like that:
Fabien Moret - Martronic SA
Thank you!

PR #339, opened 2017-07-26
[10.0][mis_builder] Allow empty date_field for queries and period behavior

Martronic-SA:
So that it's possible to search things independently from period.
For example if you want to get the number of employees.


adrienpeiffer:
As MIS Builder module hase been moved in https://github.com/OCA/mis-builder, I'll close this one

PR #341, opened 2017-08-18
[10.0][MIG] account_financial_report_horizontal

MiquelRForgeFlow:
Standard migration.

### DIFF
--- a/account_financial_report_horizontal/README.rst
+++ b/account_financial_report_horizontal/README.rst
@@ -11,8 +11,6 @@ Usage
 
 After the module is installed, the balance sheet and profit and loss reports will be in landscape mode with assets left and liabilities right.
 
-* https://www.odoo.com/forum/help-1
-
 Credits
 =======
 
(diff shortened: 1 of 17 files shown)


AaronHForgeFlow (commented):
Travis configuration is wrong:

image

MiquelRForgeFlow:
Ok, but that Travis update can be done in another PR.

MiquelRForgeFlow:
@aheficent Travis configuration is fixed now.

PR rebased.

hbrunn (approved):

AaronHForgeFlow (approved):
:+1: Functional + soft code review

hbrunn:
I just notice this just gives me duplicate column on runbot: 
[Financial report.pdf](https://github.com/OCA/account-financial-reporting/files/2103119/Financial.report.pdf)

is this a known problem or am I missing some configuration?

MiquelRForgeFlow:
Unknown problem it seems.

hbrunn:
thanks for the reply, this is caused by changes in the account types. As I don't think we should be changing them, I came up with #438 - another alternative would be to just add a selection left/right for reports, then the user can configure it on her own

PR #343, opened 2017-08-30
[IMP] account_tax_balance: add index

sbidoul:
This indexes improve the display time of
account_tax_balance by a factor of 10 on a database
with 1M move lines.

### DIFF
--- a/account_tax_balance/__manifest__.py
+++ b/account_tax_balance/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Tax Balance",
     "summary": "Compute tax balances based on date range",
-    "version": "10.0.1.1.0",
+    "version": "10.0.1.1.1",
     "category": "Accounting & Finance",
     "website": "https://www.agilebg.com/",
     "author": "Agile Business Group, Therp BV, Tecnativa, ACSONE SA/NV, "
--- a/account_tax_balance/models/__init__.py
+++ b/account_tax_balance/models/__init__.py
@@ -4,3 +4,4 @@
 
 from . import account_move
 from . import account_tax
+from . import account_move_line
--- a/account_tax_balance/models/account_move_line.py
+++ b/account_tax_balance/models/account_move_line.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = 'account.move.line'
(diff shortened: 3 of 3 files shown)


pedrobaeza:
I think this is now best done in post_init_hook, but you will need to call it also from a migration script.

Please change also version number.

sbidoul:
@pedrobaeza version bumped.

Why should I use post_init_hook or a migration script? The index creation is properly applied with a -u so it looks ok to me.

pedrobaeza:
It's for avoiding that this code is executed each time you run the server, which happens using `init`.

JordiBForgeFlow:
I agree with @pedrobaeza. We have plenty of examples in OCA of an init hook
that is called also by a migratiln script

sbidoul:
@pedrobaeza @jbeficent, performance, seriously? The test for the index existence takes something like 150 microseconds during server startup. And it's done like this everywhere in Odoo.

pedrobaeza:
Well, this is the kind of analysis you have made me in the other PR, but OK to let it as is.

Just remember that summing all these operations in `init` of each installed module can suppose an slower startup time. And although Odoo does it, it can be a mix of old inheritance because `post_init_hook` didn't exist previously and bad practices.

JordiBForgeFlow:
Oh, so indexes are created in the init. Was not aware of that.

pedrobaeza (approved):

sbidoul:
@pedrobaeza @jbeficent I'm curious now. I did a quick test and it seems to me `init()` runs with -u or -i but *not* during normal server startup. Its docstring says it runs after `_auto_init` which runs during -u/-i, right? Do I miss something?

sbidoul:
@Cedric-Pigeon @adrienpeiffer You probably need this.

pedrobaeza:
@sbidoul you are totally right! The method is only called on upgrades, not each time you restart server. I have been avoiding this since a long time for this bad assumption. Crystal clear now.

Cedric-Pigeon (approved):

sbidoul:
@pedrobaeza So, post_init_hook + migration is for very heavy operations that we don't want to run during each -u, correct?

pedrobaeza:
Well, the main difference is that `post_init_hook` is only run once: when installing module. The `init` method instead is called each time you make `-u` or `-i`, so it's acceptable to include these operations here. Of course, heavy operations can't be done each time you upgrade.

As side note, check the new module LasLabs and we have been cooking for autoupdating modules: https://github.com/OCA/server-tools/tree/10.0/module_auto_update

JordiBForgeFlow:
I guess that we can merge this one now.

adrienpeiffer (approved):

PR #354, opened 2017-09-30
[9.0][DEL] mis_builder moved to OCA/mis-builder

sbidoul:
https://github.com/OCA/mis-builder/tree/9.0


pedrobaeza:
Do you think this is a good idea? A lot of people that has this repository added (I'm not talking about pip installations) will find after the update that they have missed the module. I suggest to move this for v11.

sbidoul:
Uh, I asked the question on the contributors mailing list and everyone was positive. Now I've done all the moving work. Besides I want to try and maintain feature parity between 9.0, 10.0 and 11.0 so I need all branches in the same repo.

pedrobaeza:
OK, the removal was implicit in your request then. Go ahead.

PR #356, opened 2017-09-30
[7.0][DEL] mis_builder moved to OCA/mis-builder

sbidoul:
https://github.com/OCA/mis-builder/tree/7.0


sbidoul:
Travis failure is unrelated.

PR #360, opened 2017-10-17
[10.0][IMP] account_financial_report_qweb: Allow filter accounts in general ledger wizard with One2many widget

sergio-teruel:
Now you can select a user permission to allow filter account with many2many_tags widget or One2many field as it was in version 9.0

cc @Tecnativa

### DIFF
--- a/account_financial_report_qweb/README.rst
+++ b/account_financial_report_qweb/README.rst
@@ -14,9 +14,16 @@ Accounting / Reporting / OCA Reports.
 - Open Items
 - Aged Partner Balance
 
+Usage
+=====
+
+You can filter accounts by tags or with a One2many field in General Ledger and
+Trial balance wizards selecting a user permission named
+"Filter financial reports accounts in One2many field"
+
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/91/9.0
+   :target: https://runbot.odoo-community.org/runbot/91/10.0
 
 Bug Tracker
 ===========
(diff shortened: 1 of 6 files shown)


carlosdauden (approved):

account_financial_report_qweb/README.rst
```diff
@@ -14,9 +14,16 @@ Accounting / Reporting / OCA Reports.
 - Open Items
 - Aged Partner Balance
 
+Usage
+=====
+
+You can filter accounts by tags or with a One2many field in General Ledger and
```
JordiBForgeFlow, line 20: Then the tags should be a included in the search view, so that the user does not have to use the "Add custom filter" option.

account_financial_report_qweb/security/account_financial_report_qweb.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Sergio Teruel - Tecnativa <sergio.teruel@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+<record id="group_one2many_financial_report_account" model="res.groups">
+    <field name="name">Filter financial reports accounts in One2many field</field>
```
JordiBForgeFlow, line 7: Objections here:
1. Why do you need a security group for this? I can see that it would be useful if you decide to activate it in the 'Accounting / Settings'. I can understand that some companies might want to keep this option invisible. But having to set it for each accounting user manually is, to me, not straightforward. 
1. I think that this name is very technical.  I think that it is easier as "Allow filtering on OCA financial reports by accounts"

JordiBForgeFlow (changes requested):

pedrobaeza:
I prefer to directly migrate https://github.com/OCA/web/tree/8.0/web_widget_many2many_tags_multi_selection

account_financial_report_qweb/README.rst
sergio-teruel: @jbeficent sorry by the confusion, I only want change the field widget from many2many_tags to One2many_tags, but I did not know about *web_widget_many2many_tags_multi_selection* module, I think that the migration of this module is the better solution, so when it is migrated I will close the PR.
Thanks.

pedrobaeza:
Closing as this has been implemented with the module I told about.

PR #361, opened 2017-10-18
[11.0][MIG] account_tax_balance

pedrobaeza:
Standard migration

@Tecnativa


pedrobaeza:
cc @jbeficent @sbidoul

account_tax_balance/README.rst
```diff
@@ -0,0 +1,63 @@
...
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Lorenzo Battistini <lorenzo.battistini@agilebg.com>
+* Giovanni Capalbo <giovanni@therp.nl>
+* Tecnativa - Antonio Espinosa
+* Tecnativa - Pedro M. Baeza
```
rafaelbn, line 47: Standard format is "Lorenzo Battistini <lorenzo.battistini@agilebg.com>"

Isn't it?

account_tax_balance/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+# Copyright 2016 Tecnativa - Antonio Espinosa
+# Copyright 2016 ACSONE SA/NV - Stéphane Bidoul
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Tax Balance",
+    "summary": "Compute tax balances based on date range",
+    "version": "11.0.1.0.0",
+    "category": "Accounting & Finance",
+    "website": "https://www.agilebg.com/",
```
rafaelbn, line 12: Should we change as discussed in the list to the OCA webpage?

account_tax_balance/README.rst
pedrobaeza: I prefer not to put here email

account_tax_balance/__manifest__.py
pedrobaeza: Well, it can be, but it's not a priority

oihane (approved):
LGTM only checked code

astirpe (approved):
LGTM (code review)

pedrobaeza:
Merged as it's a dependency for a lot of modules of l10n-spain and there are enough approvals.

levkar:
Taxes whose amount_type is group will always have zero balances. Would it be more convenient to hide/filter them?

pedrobaeza:
Well, that's depend on how it's defined the subtaxes. I prefer to show both parent and children, but it can be a configuration.

PR #362, opened 2017-10-23
[8.0] [FIX] open invoices report: Include open invoices from previous years still to be paid

jbaudoux:
When a fiscal year is selected, start from the first special period instead of the first period of the fiscal year.  Otherwise outstanding invoices from previous years are not shown.

README states that it "shows open invoices from previous years"

When starting from the first period of the year, the report does not include outstanding invoices of previous years. The first special period is the first opening period in odoo having the initial partner balance.

### DIFF
--- a/account_financial_report_webkit/report/open_invoices.py
+++ b/account_financial_report_webkit/report/open_invoices.py
@@ -120,7 +120,7 @@ class PartnersOpenInvoicesWebkit(report_sxw.rml_parse,
         group_by_currency = self._get_form_param('group_by_currency', data)
 
         if main_filter == 'filter_no' and fiscalyear:
-            start_period = self.get_first_fiscalyear_period(fiscalyear)
+            start_period = self._get_first_special_period()
             stop_period = self.get_last_fiscalyear_period(fiscalyear)
 
         # Retrieving accounts


sbidoul:
This should normally work correctly and the opening entry / special period should not influence open invoice reports.

Maybe you used the "Unreconciled" Deferral Method? If that' the case I suggest you try to redo your opening entries with only the "Balance" Deferal Method on all BS accounts.

jbaudoux:
This is exactly because for partners, the "Balance" deferral method is used that you cannot look only at the moves of the fiscal year. All move lines from previous years bound to a partner have to be taken into account. Those could be still to be paid.

For instance, when I look at what where the open invoices in 2016, I also want to see the invoices before 2016 that were still to be paid in 2016. Those are not generating any move lines in 2016 and are not reported in the open invoices report.

jbaudoux:
ccing @fclementic2c @jgrandguillaume @luc-demeyer @guewen for review

luc-demeyer:
when it comes to having a view on open accounting entries at the end of a fiscal period or fiscal year in V8 than you should not use this report since the logic is based on dates and not on fiscal periods.
The account_open_payables_receivables_xls (cf. https://github.com/luc-demeyer/noviat-apps/tree/8.0) does the job.

In case all of your supplier/customer invoices have dates within their accounting period, than also the Open Invoices report is correct (at least in the cases that I checked in the past) and this without the fix that you suggest.
Have a look to the data you have posted in the very first special period in your system (hopefully only account payable/receivables balances not the details).

PR #363, opened 2017-10-25
[8.0] [FIX] Partner balance - Initial balance and first special period

jbaudoux:
The partner balance initial balance is wrongly computed when you have moves on periods before your first special period.
The first special period is the period where you have your initial balance of each partner. (i.e start point of computation for partner balance)
The initial balance was computed taking also into account all moves before the start date of your partner balance report.
This fix ensure you do not go beyond the first special period.

### DIFF
--- a/account_financial_report_webkit/README.rst
+++ b/account_financial_report_webkit/README.rst
@@ -165,6 +165,7 @@ Contributors
 * Nicolas Bessi
 * Guewen Baconnier
 * David Dufresne <david.dufresne@savoirfairelinux.com>
+* Jacques-Etienne Baudoux <je@bcim.be>
 
 Maintainer
 ----------
--- a/account_financial_report_webkit/report/common_partner_reports.py
+++ b/account_financial_report_webkit/report/common_partner_reports.py
@@ -85,22 +85,6 @@ class CommonPartnersReportHeaderWebkit(CommonReportHeaderWebkit):
                    key=lambda p: datetime.strptime(p.date_start,
                                                    DEFAULT_SERVER_DATE_FORMAT))
 
-    def _get_period_range_from_start_period(self, start_period,
-                                            include_opening=False,
-                                            fiscalyear=False,
-                                            stop_at_previous_opening=False):
-        """We retrieve all periods before start period"""
-        periods = super(CommonPartnersReportHeaderWebkit, self).\
-            _get_period_range_from_start_period(
-                start_period,
-                include_opening=include_opening,
-                fiscalyear=fiscalyear,
-                stop_at_previous_opening=stop_at_previous_opening)
-        first_special = self._get_first_special_period()
-        if first_special and first_special.id not in periods:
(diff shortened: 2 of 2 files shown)


jbaudoux:
For review. @fclementic2c @guewen @luc-demeyer @jgrandguillaume

jbaudoux:
Thanks to this fix, the partner balance total for each account correspond now to the balance sheet amount for those accounts.

sbidoul:
@jbaudoux IMO using the special period to initialize the balances is not a good practice.

To the best of my knowledge, all these reports work fine when you initialise the balances at the last day of the previous fiscal year, then do a normal opening (using the Balance method for all BS accounts). 

Actually I'm afraid this PR would break some use cases.

jbaudoux:
@sbidoul This is not the way the initial balance is computed for each partner. The balance deferral method only generates a total in your fiscal year opening period and not a total for each partner.
For computing the initial balance of a partner, the report considers all moves for that partner before your report start date. But this needs to be limited to you first special period and cannot go beyond.

sbidoul:
I'm not sure I understand. Do you assume your initial partner balance is in the first special period?

jbaudoux:
When you report for instance your partner balance for 2016, the report show for each client/supplier account, each partner:
- the initial balance
- the debit = total of all move lines of that partner in 2016
- the credit = total of all move lines of that partner in 2016

The start date is 2016-1-1 and end date 2016-12-31

The initial balance is computed by looking at all move lines for that partner before 2016 excluding opening entries but including your first opening entry (= first special period) where you put your initial partner balance when you started your general accounting in odoo.
This is what the call to
`self._get_period_range_from_start_period(start_period, fiscalyear=False, include_opening=False)`
was doing.
If your odoo accounting starts the 2014-1-1, your first special period is 2014-1-1, and this is where you have the starting initial balance of each partner.
Consider now you have periods before 2014-1-1 because you wanted to keep an history of all transactions before doing the accounting in odoo.
The initial partner balance must be computed from 2014-1-1 (first special period) until 2016-1-1 (report start date) and include the first special period and exclude the start date.
This is what the report already does in the code except that it goes beyond 2014-1-1. What I changed is to give a start point in the period search.

sbidoul:
> If you odoo accounting start the 2014-1-1, your first special period is 2014-1-1, and this is where you have the starting initial balance of each partner.

That's where the problem is IMO: initial partner balances must not be in the special period. They must be loaded in an adjustment move on 2013-12-31. And then you generate the first opening entry normally using the standard Odoo feature.

jbaudoux:
You cannot do that if the historical data has different deferral methods than in odoo. The first special period allows you to do it differently without adjustment move.

`'first_special_period_id': fields.many2one(
              'account.period', 'First Special Period',
              domain=[('special', '=', True)],
              help="When specified, the accounting reports "
                   "will consider this period as the start period "
                   "for the calculation of opening balances."`

PR #364, opened 2017-10-27
[10.0][ADD] Add taxes description in general ledger report

benwillig:
This PR only adds taxes description in the general ledger report.


benwillig:
cc @sbidoul

account_financial_report_qweb/report/general_ledger.py
```diff
@@ -890,6 +892,28 @@ def _inject_line_not_centralized_values(
...
+    CASE
+        WHEN
+            ml.tax_line_id is not null
+        THEN
+            at.name
+        WHEN
+            (SELECT count(*)
+            FROM account_move_line_account_tax_rel aml_at_rel
+            WHERE aml_at_rel.account_move_line_id = ml.id
+            LIMIT 1) > 0
```
sbidoul: I think this WHEN on select count is unnecessary because `array_to_string(array_agg(at.name), ', ')` will return an empty string when there is no record. So it should be more performant without.

account_financial_report_qweb/report/general_ledger.py
```diff
@@ -890,6 +892,28 @@ def _inject_line_not_centralized_values(
     m.name AS entry,
     j.code AS journal,
     a.code AS account,
+    CASE
+        WHEN
+            ml.tax_line_id is not null
+        THEN
+            at.name
```
sbidoul: use `at.description` as in most localisation it is shorter than `at.name`. Or `coalesce(at.description, at.name)` which would more or less match what is done on the invoice template.

account_financial_report_qweb/report/general_ledger.py
benwillig: Indeed, I updated my code.

sbidoul:
@benwillig could you post an anonymized screenshot to facilitate review?

sbidoul (approved):

adrienpeiffer (approved):
:+1: (review and test)

melroy89:
Next is 11.0 branch as well! Don't forget v11.

melroy89 (approved):

feketemihai:
@pedrobaeza @jbeficent Can you please merge this one

PR #365, opened 2017-10-30
[10.0][add][bank_statement_reconciliation_summary]

JordiBForgeFlow:
Bank Statement Reconciliation Summary
=====================================

The Bank Reconciliation Summary shows how your actual bank statement balance
and the balance of the bank account in Odoo match, after taking into account
any unreconciled items.

This report is useful if your bank account's statement balance and balance
in Odoo do not match, and you need to check for duplicate or manually created
transactions that might cause the discrepancy.

This report was created, inspired by Xero’s bank reconciliation summary,
https://help.xero.com/Report_BankRec, and from a basic explanation of the
bank statement reconciliation summary:
http://www.accountingcoach.com/bank-reconciliation/explanation.

Usage
=====

#. Go to *Accounting / Reports / OCA accounting reports / Bank Statement Reconciliation*
#. Enter the end date for your analysis, the bank journal and the actual bank
   balance at that date.

The report displays the following sections:

* Balance in Odoo
* Plus Outstanding Payments - payments in odoo not yet reconciled
* Less Outstanding Receipts - payments in odoo not yet reconciled
* Plus Un-Reconciled Statement Lines - statement lines not yet reconciled
* Statement Balance

If the theoretical Statement Balance does not match with Odoo, it will
display the Computed (theoretical) balance, the Actual Balance and
the Unencoded Statement Amount.

Screenshot:

image


AaronHForgeFlow:
It works fine. The headers are not displayed well probably because wkhtmltopdf version in travis. 
image

ovnicraft:
I start working in something related, go to work with this, possible review to merge ?

jackjack82:
Hi, few notes. I think outstanding payments and receipts shall not be compulsory, i.e. add a boolean like "add outstanding paymentd and receipts".
Moreover, it would be interesting to add account.moves with the same journal but with no link to bank statement lines. Such moves can be created manually or by other modules, by mistake you can even have double account moves for one single statement line. I can add this if you want.

ovnicraft:
@jackjack82 sounds good, this PR has no update from 2017 if you take on its better.

JordiBForgeFlow:
@jackjack82 I have added the feature you mentioned of "add account.moves with the same journal but with no link to bank statement lines", and removed the option to search for outstanding payments/receipts that are outside of the bank journal. It could be added by a separate module, if someone needs it. 

It think that this now makes much more sense.

@ovnicraft Can you also review if it now looks good?

PR #366, opened 2017-11-07
[10.0] afr qweb journal report

sbidoul:
This is the WIP PR (by @benwillig) for the journal report, including the tax summary.

This should merge the best features of 8.0 [account_journal_report_xls](https://github.com/OCA/account-financial-reporting/tree/8.0/account_journal_report_xls) and 9.0 [account_journal_report](https://github.com/OCA/account-financial-reporting/tree/9.0/account_journal_report) as discussed in #327.

image

image

The Excel export remains to be implemented (later this week). Early review is welcome though.

### DIFF
--- a/account_financial_report_qweb/menuitems.xml
+++ b/account_financial_report_qweb/menuitems.xml
@@ -16,6 +16,14 @@
         />
 
     <menuitem
+        id="journal_report_wizard_menu"
+        name="Journal Ledger"
+        parent="menu_oca_reports"
+        action="journal_report_wizard_act_window"
+        sequence="15"
+        />
+
+    <menuitem
         parent="menu_oca_reports"
         action="action_trial_balance_wizard"
         id="menu_trial_balance_wizard"
--- a/account_financial_report_qweb/report/journal_report.py
+++ b/account_financial_report_qweb/report/journal_report.py
@@ -1,0 +1,764 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+
+DIGITS = (16, 2)
+
+
+class ReportJournalQweb(models.TransientModel):
+
+    _name = 'report_journal_qweb'
+
+    date_from = fields.Date(
+        required=True
+    )
+    date_to = fields.Date(
+        required=True
+    )
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        required=True,
+        ondelete='cascade'
+    )
+    move_target = fields.Selection(
+        selection='_get_move_targets',
+        default='all',
+        required=True,
+    )
+    sort_option = fields.Selection(
+        selection='_get_sort_options',
+        default='move_name',
+        required=True,
+    )
+    group_option = fields.Selection(
+        selection='_get_group_options',
+        default='journal',
+        required=True,
+    )
+    journal_ids = fields.Many2many(
+        comodel_name='account.journal',
+        required=True,
+    )
+    report_journal_ids = fields.One2many(
+        comodel_name='report_journal_qweb_journal',
+        inverse_name='report_id',
+    )
+    report_move_ids = fields.One2many(
+        comodel_name='report_journal_qweb_move',
+        inverse_name='report_id',
+    )
+    report_move_line_ids = fields.One2many(
+        comodel_name='report_journal_qweb_move_line',
+        inverse_name='report_id',
+    )
+    report_journal_tax_line_ids = fields.One2many(
+        comodel_name='report_journal_qweb_journal_tax_line',
+        inverse_name='report_id',
+    )
+    report_tax_line_ids = fields.One2many(
+        comodel_name='report_journal_qweb_report_tax_line',
+        inverse_name='report_id',
+    )
+    with_currency = fields.Boolean()
+    with_account_name = fields.Boolean()
+
+    @api.model
+    def _get_move_targets(self):
+        return self.env['journal.report.wizard']._get_move_targets()
+
+    @api.model
+    def _get_sort_options(self):
+        return self.env['journal.report.wizard']._get_sort_options()
+
+    @api.model
+    def _get_group_options(self):
+        return self.env['journal.report.wizard']._get_group_options()
+
+    @api.multi
+    def compute_data_for_report(self):
+        self.ensure_one()
+        self._inject_journal_values()
+        self._inject_move_values()
+        self._inject_move_line_values()
+        self._inject_journal_tax_values()
+        self._update_journal_report_total_values()
+
+        if self.group_option == 'none':
+            self._inject_report_tax_values()
+
+    @api.multi
+    def refresh(self):
+        self.ensure_one()
+        self.report_journal_ids.unlink()
+        self.compute_data_for_report()
+
+    @api.multi
+    def _inject_journal_values(self):
+        self.ensure_one()
+        sql = """
+            INSERT INTO report_journal_qweb_journal (
+                create_uid,
+                create_date,
+                report_id,
+                journal_id,
+                name,
+                code,
+                company_id,
+                currency_id
+            )
+            SELECT
+                %s as create_uid,
+                NOW() as create_date,
+                %s as report_id,
+                aj.id as journal_id,
+                aj.name as name,
+                aj.code as code,
+                aj.company_id as company_id,
+                COALESCE(aj.currency_id, company.currency_id) as currency_id
+            FROM
+                account_journal aj
+            LEFT JOIN
+                res_company company on (company.id = aj.company_id)
+            WHERE
+                aj.id in %s
+            AND
+                aj.company_id = %s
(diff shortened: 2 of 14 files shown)


account_financial_report_qweb/report/journal_report.py
```diff
@@ -0,0 +1,599 @@
...
+                aj.id in %s
+            AND
+                aj.company_id = %s
+            ORDER BY
+                aj.name
+        """
+        params = (
+            self.env.uid,
+            self.id,
+            tuple(self.journal_ids.ids),
```
adrienpeiffer, line 134: Currenlty, journal_ids isn't required. So, this line will raise an exception.

adrienpeiffer (changes requested):

account_financial_report_qweb/wizard/journal_report_wizard.py
```diff
@@ -0,0 +1,79 @@
...
+class JournalReportWizard(models.TransientModel):
+
+    _name = 'journal.report.wizard'
+
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        default=lambda self: self.env.user.company_id,
+        string='Company'
+    )
+    date_range_id = fields.Many2one(
```
adrienpeiffer: Can you add a filter on the company ?

adrienpeiffer: ondelete='cascade'

adrienpeiffer (changes requested):

account_financial_report_qweb/report/journal_report.py
benwillig: I forgot to add the constrains. Thanks

account_financial_report_qweb/wizard/journal_report_wizard.py
benwillig: If you meant to add a filter in the excel report, I just did it.

adrienpeiffer (approved):
:+1: Thanks !

sbidoul:
@mreficent would you like to test if this one meets your requirements?

JordiBForgeFlow:
I miss the part that indicates the amount in currency and currency.
IMHO the report should also provide the currency of the company if, none specified in the journal.

So, Customer Invoices **(eur)** - 01/01/2017 to 12/31/2017 - All Moves

Now I created a journal in EUR for a company that uses USD.

Then I create an invoice in EUR. I expect the journal report to show me the amounts both in EUR and USD.

Journal entry:
image

Journal report:

image

account_financial_report_qweb/menuitems.xml
```diff
@@ -15,6 +15,14 @@
         sequence="10"
         />
 
+    <menuitem
+        id="journal_report_wizard_menu"
+        name="Journal"
```
MiquelRForgeFlow: -> Journal Ledger

account_financial_report_qweb/wizard/journal_report_wizard.xml
```diff
@@ -0,0 +1,43 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <group>
+                    <field name="company_id" groups="base.group_multi_company"/>
+                    <field name="date_range_id"/>
+                    <field name="date_from"/>
+                    <field name="date_to"/>
+                    <field name="move_target" widget="radio" options="{'horizontal': true}"/>
+                    <field name="journal_ids" widget="many2many_tags"/>
+                </group>
```
MiquelRForgeFlow: I would divide this group in two groups, just to make smaller the `date_from` and `date_to` fields.

Also, I miss the Landscape mode option and the Entries Sorted By option.

account_financial_report_qweb/wizard/journal_report_wizard.xml
```diff
@@ -0,0 +1,43 @@
...
+                <footer>
+                    <button name="export_as_pdf"
+                            string="Print" class="btn-primary"
+                            type="object"/>
+                    <button name="export_as_xlsx"
+                            string="Export" class="btn-primary"
+                            type="object"/>
+                    <button string="Cancel"
+                                    class="btn-default"
+                                    special="cancel"/>
```
MiquelRForgeFlow: could you align `class` and `special` under `string`?

account_financial_report_qweb/wizard/journal_report_wizard.py
```diff
@@ -0,0 +1,85 @@
...
+               "('company_id', '=', False),"
+               "('company_id', '=', company_id)]",
+    )
+    date_from = fields.Date(
+        string="Start date",
+        required=True
+    )
+    date_to = fields.Date(
+        string="End date",
+        required=True
```
MiquelRForgeFlow, line 32: I miss a default here or in the date range

account_financial_report_qweb/wizard/journal_report_wizard.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    journal_ids = fields.Many2many(
+        comodel_name='account.journal',
+        string="Journals",
+        domain="[('company_id', '=', company_id)]",
+        required=True,
```
MiquelRForgeFlow, line 38: I don't know but maybe can be comfortable if all the journals are selected by default

JordiBForgeFlow: No.... Most of the time you will use the report for just one journal. Unclicking all others would be a mess

JordiBForgeFlow: No.... Most of the time you will use the report for just one journal. Unclicking all others would be a mess

account_financial_report_qweb/wizard/journal_report_wizard.py
JordiBForgeFlow: No. Better use date range. And no defauñt needed. Date range is simple enough

account_financial_report_qweb/wizard/journal_report_wizard.xml
benwillig: I updated the view and I added the sort option. About the landscape mode, we use the same paper format for all the reports included in ```account_financial_report_qweb``` so I don't think we should add this option.

benwillig:
@jbeficent I added the currency amount. To get the new columns, you just have to check the new option ```With currency``` in the wizard.

mreficent:
runbot is not working  :confused:

JordiBForgeFlow (changes requested):
The current approach will split by journal. In Spain we require that all journal entries must appear in sequence, regardless of what journal they belong to. 

This feature did not exist in 9.0 (https://github.com/OCA/account-financial-reporting/tree/9.0/account_journal_report) and represents a regression for us.

I propose that we let the user choose if he wants the report to show the results grouped by journal or not.

melroy89:
@jbeficent The same remark is applicable for The Netherlands, since these are European laws. thx Ps don't forgot to sync to branch 11.0.

benwillig:
@jbeficent I added an option to not group entries by journals. Your review is welcome.

pedrobaeza:
How is the state of this?

benwillig:
This PR is just waiting for reviews now.

pedrobaeza:
OK, thanks. I'll check then.

JordiBForgeFlow:
checking now

JordiBForgeFlow (approved):

pedrobaeza:
After reviewing it, I'm missing the name of the account. I know that sometimes it can be a bit large, but most of my customers need for knowing which account you are talking about. Example: 572001 is the bank A, and 572002 is account B, but you don't remember which one is. Seeing only the code you can't decide. 

On contrary, I see very large the width of the "Partner" column, not needing that size. Can this be done auto-adjustable not defining classes that limit the widths?

benwillig:
I added an option to display the account names or not. I tried to not set the columns width but then the other columns sizes also changes randomly.

JordiBForgeFlow (approved):

pedrobaeza:
@benwillig which columns? Maybe you can fix some of them that are "critical", but let free the rest.

MiquelRForgeFlow:
I think options like `with_account_name` should be applied also in the xlsx report for consistency 🤔

benwillig:
@mreficent Yes I forgot about it ^^. I updated.

@pedrobaeza About the columns width, I think it's better to have fixed sizes instead of letting the data decide. I think the report should looks the same in all cases

pedrobaeza:
@benwillig OK, let's fix the width, but please give more room for account name and decrease label and tax width for longer account names (it's usual in Spain for example to have some long account names).

pedrobaeza:
@benwillig can you please make the last change and we'll merge this?

benwillig:
@pedrobaeza I updated the column sizes.

MiquelRForgeFlow (approved):

pedrobaeza:
May I squash your commits on merge?

pedrobaeza (approved):

benwillig:
@pedrobaeza Done. Ready to merge now.

pedrobaeza:
Thanks!

PR #367, opened 2017-11-23
[11.0][mig] account_financial_report

JordiBForgeFlow:
Account Financial Reports
====================

This module adds a set of financial reports. They are accessible under
Accounting / Reporting / OCA Reports.

- General ledger
- Trial Balance
- Open Items
- Aged Partner Balance


jcdrubay:
Should we take the opportunity to:
 - Rename to account_report_xlsx (because those reports are not qweb reports)
 - Split into several modules, because it is really tough to do a code review of this PR :)
   - account_report_xlsx_base
   - account_report_xlsx_gl
   - account_report_xlsx_bs
   - ...

Apart from the benefit of the code review, it would be easier to delegate and less scary to get engaged into this work of porting from one version to the next.

JordiBForgeFlow:
All travis errors are not valid. See https://github.com/OCA/maintainer-quality-tools/issues/514

@jcdrubay The reports are both qweb AND xlsx. I have renamed to account_financial_report. I would prefer not to split the reports, for now (I spent an awful lot of time already :S)

fclementic2c:
LGTM. 
- I could run all reports, 
- initial balances ok

Only 1 issue, I have created a supplier invoice in EUR (on the OCA runbot linked to this PR), I do not see the countervalue in EUR on the open items report nor the GL.
I think on the GL we only display currency if the account+journal has a secondary currency... which is not ideal I think.
...but this is issue is also there on Odoo v10

JordiBForgeFlow:
@fclementic2c I have now fixed the issue with the currency, so that in case that you post an invoice in EUR, for a company that has USD as main currency, you will see the amout in EUR in the column amount currency.

I will also propose an equivalent fix for v10.

fclementic2c:
@jbeficent Thank you ! You're the man ! :)

JordiBForgeFlow:
@fclementic2c I missed the open items reports. It is now fixed as well.

pedrobaeza:
I have rebased the PR for checking about one thing

yvaucher:
Just for nitpicking:
account_financial_report/i18n/account_financial_report_qweb.pot was not renamed

HviorForgeFlow:
Thanks, is it fine this way?

hbrunn:
can we count the positive comments as positive reviews and merge?

PR #368, opened 2017-11-29
[10.0][account_financial_report_qweb] fix multicurrency support

JordiBForgeFlow:
This PR fixes the support for multicurrency.

Prior to this PR, the secondary currency would only appear in the general ledger and open items reports when the GL account had a secondary currency.

But this does not make sense because you often post transaction using another currency, to an account that is held in the company currency, and in those cases you still want to evaluate the general ledger and open items reports in terms of also the currency of the transaction, in addition to the company currency.

cc @fclementic2c


fclementic2c:
LGTM 
+1

MiquelRForgeFlow (approved):
LGTM :+1:

PR #369, opened 2017-12-06
[8.0][FIX] open invoices - last_rec_date field value cannot be before the period start date

jbaudoux:
The last_rec_date is used in the open invoices report.

If you allow to have account move line dates outside the related period boundaries, then you need to take care of the period dates when computing the last_rec_date.
Consider a purchase where you get an invoice dated on 2016 and you pay it in 2016 but you encode the invoice only in 2017. Looking at the move lines, the payment is declared in 2016 (debit) and the purchase is declared in 2017 (credit). However, looking only at the move lines dates, they both are dated in 2016. So it is important to not consider the 2 lines as reconciled in 2016 and keep the payment as open on 2016.

Note also this previous merged fix https://github.com/OCA/account-financial-reporting/pull/213
On both cases, you need to force the re-computation of the field (see migration script)

### DIFF
--- a/account_financial_report_webkit/migrations/8.0.1.2.0/post-migration.py
+++ b/account_financial_report_webkit/migrations/8.0.1.2.0/post-migration.py
@@ -17,13 +17,15 @@ def migrate(cr, version):
         """
         UPDATE account_move_line SET last_rec_date = rec_data.aml_date
         FROM (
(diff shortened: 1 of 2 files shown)


sbidoul (approved):

feketemihai:
@jbaudoux I would take max between actual line date and last_line from reconciled..

PR #372, opened 2017-12-22
[11.0][account_financial_report] remove dependency with account_fiscal_year

JordiBForgeFlow:
The dependency is not required for the reports to run correctly.

cc @sbidoul @pedrobaeza

PR #373, opened 2017-12-22
[IMP] account_financial_report_qweb: Remove unneeded dependency

pedrobaeza:
Module `account_fiscal_year` allows companies with fiscal years with duration different from 1 year to make financial reporting, but the method used is overwrting a standard method `compute_fiscalyear_dates` on company, so it's transparent for this module, and you don't need to have it installed for most of the companies, that have regular fiscal years.

This is the same as #372, but for 10.0

@Tecnativa

### DIFF
--- a/account_financial_report_qweb/__manifest__.py
+++ b/account_financial_report_qweb/__manifest__.py
@@ -16,7 +16,6 @@
     'depends': [
         'account',
         'date_range',
-        'account_fiscal_year',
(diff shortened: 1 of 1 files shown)


sbidoul (approved):

JordiBForgeFlow (approved):

pedrobaeza:
Thanks!

PR #379, opened 2018-02-01
[8.0][ADD]account_debtor_card

Kiplangatdan:


lfreeke:
:+1: (Test on runbot)

feketemihai:
@Kiplangatdan I don't think we should mix reports like this, one is the sale order, and other thing is the invoice, plus that you are printing the invoice without headers, will not look the same as teh invocie printed from invoice itself, but like i said, don't think this a good behaviour, what happens if the 2 report don't match in qty, pricing...from my point this should not be merged

PR #382, opened 2018-02-14
[10.0][IMP] add salesperson to account_financial_report_qweb

robyf70:


pedrobaeza:
You are mixing a sales concept with financial one, so this is a nonsense IMO. 👎

robyf70:
@pedrobaeza I might agree with you, but how can I give at an agent/salesman a list of due payments?

pedrobaeza:
You have the module account_credit_control for that, with the possibility of assigning a responsible for each due.

robyf70:
ahh! will give a look! is there also a similar report?

robyf70:
@pedrobaeza Not sure if I've understood the module you told me, but looks like I cannot give an agent a report of all due payments. The account_credit_control seems doing other automatized checks or I'm missing something?

pedrobaeza:
You can automate the responsible to be the "agent" and split the information.

robyf70:
... which is fine and useful for that purpose. But still, We don't have a report where we can filter all due payments by agent/salesperson.

pedrobaeza:
Then make it a separate report, but this can't be integrated in main one.

robyf70:
Ok! will do that

PR #387, opened 2018-02-22
[IMP][10.0] customer_outstanding_statement

HviorForgeFlow:
Minor changes proposal

### DIFF
--- a/customer_outstanding_statement/__manifest__.py
+++ b/customer_outstanding_statement/__manifest__.py
@@ -12,7 +12,8 @@
     'website': 'https://github.com/OCA/account-financial-reporting',
     'license': 'AGPL-3',
     'depends': [
-        'account',
+        'account_invoicing',
+        'web',
     ],
     'data': [
         'views/statement.xml',
--- a/customer_outstanding_statement/report/customer_outstanding_statement.py
+++ b/customer_outstanding_statement/report/customer_outstanding_statement.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Eficent Business and IT Consulting Services S.L.
-#   (http://www.eficent.com)
+# Copyright 2018 Fork Sand, Inc.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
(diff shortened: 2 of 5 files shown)


customer_outstanding_statement/report/customer_outstanding_statement.py
```diff
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Eficent Business and IT Consulting Services S.L.
```
MiquelRForgeFlow, line 2: Please, don't delete copyright history if not strictly necessary.

customer_outstanding_statement/__manifest__.py
```diff
@@ -12,7 +12,8 @@
     'website': 'https://github.com/OCA/account-financial-reporting',
     'license': 'AGPL-3',
     'depends': [
-        'account',
+        'account_invoicing',
+        'web',
```
MiquelRForgeFlow, line 16: Don't put web. web module is installed with base.

MiquelRForgeFlow (changes requested):
I think this PR is not for v10 but for v11.
Thus, it's a migration and then it should be done as all migrations:
 - [11.0][MIG] ...
 - Maintaining commit history. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0.

customer_outstanding_statement/report/customer_outstanding_statement.py
JordiBForgeFlow: remove # -*- coding: utf-8 -*- everywhere

customer_outstanding_statement/__manifest__.py
```diff
@@ -12,7 +12,8 @@
     'website': 'https://github.com/OCA/account-financial-reporting',
     'license': 'AGPL-3',
     'depends': [
-        'account',
+        'account_invoicing',
```
MiquelRForgeFlow, line 15: Also don't change account module to account_invoicing if strictly necessary.

This module is modifying things from account, not from account_invoicing.

MiquelRForgeFlow (changes requested):

PR #389, opened 2018-02-23
[11.0][MIG] customer activity statement

HviorForgeFlow:
No major changes
@mreficent @jbeficent @feketemihai


feketemihai:
@hveficent @mreficient i made on runbot 2 invoices for jackson group, print the activity at 26.02, you will see..

feketemihai:
Sorry, wrong pr.. I will make again on outstanding, but even on this one is not ok..

customer_activity_statement/views/statement.xml
```diff
@@ -0,0 +1,206 @@
...
+                <div class="row">
+                    <div class="col-xs-5 col-xs-offset-7">
+                        <span t-field="o.name"/><br/>
+                        <span t-raw="o.contact_address.replace('\n\n', '\n').replace('\n', '&lt;br&gt;')"/>
+                        <span t-field="o.vat"/>
+                    </div>
+                    <h4 style="padding-left:20em">
+                        Activity Statement
+                    </h4>
+                    <p>
```
feketemihai: @hveficient the same here, today is taken, not date_end of wizard.

customer_activity_statement/wizard/customer_activity_statement_wizard.py
```diff
@@ -0,0 +1,56 @@
...
+    _description = 'Customer Activity Statement Wizard'
+
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        default=lambda self: self.env.user.company_id,
+        string='Company'
+    )
+
+    date_start = fields.Date(required=True,
+                             default=fields.Date.to_string(
```
feketemihai, line 22: @hveficent by default, when you print a statement, it should be from 1st of january, begining of year.

HviorForgeFlow: @feketemihai we appreciate your feedback. We must reconsider the logic of this module. Thanks. Feel free to collaborate.

feketemihai: @mreficent since the report are close to each other, isn't it easier to have a option in wizard for outstanding, if set only take outstanding, otherwise all activity.

MiquelRForgeFlow: We will consider it, thanks.

feketemihai: @hveficient @mreficent i can do it on monday to consolidate report, plus issue of reconciled.

MiquelRForgeFlow: Thanks, but after some consideration, we prefer to do ourselves.

MiquelRForgeFlow: @feketemihai 
- <code>date_start</code> is updated now as you proposed.
- A fusion of the two reports will be considered in the future after the two PR get merged, just to don't mess all and to maintain all commit history.

MiquelRForgeFlow (approved):

customer_activity_statement/report/customer_activity_statement.py
```diff
@@ -0,0 +1,388 @@
...
+            END as credit,
+            CASE WHEN l.date_maturity is null
+                THEN l.date
+                ELSE l.date_maturity
+            END as date_maturity
+            FROM account_move_line l
+            JOIN account_account_type at ON (at.id = l.user_type_id)
+            JOIN account_move m ON (l.move_id = m.id)
+            WHERE l.partner_id IN (%s) AND at.type = 'receivable'
+                                AND '%s' < l.date AND l.date <= '%s'
```
feketemihai: AND '%s' < l.date i think here it should be AND '%s' <= l.date, to take move lines from date start as well.

customer_activity_statement/report/customer_activity_statement.py
```diff
@@ -0,0 +1,388 @@
...
+            END as debit,
+            CASE WHEN l.currency_id is not null AND l.amount_currency < 0.0
+                THEN sum(l.amount_currency * (-1))
+                ELSE sum(l.credit)
+            END as credit
+            FROM account_move_line l
+            JOIN account_account_type at ON (at.id = l.user_type_id)
+            JOIN account_move m ON (l.move_id = m.id)
+            WHERE l.partner_id IN (%s) AND at.type = 'receivable'
+                                AND l.date <= '%s' AND not l.blocked
```
feketemihai: AND l.date <= '%s' and here it should be AND l.date < '%s', to take initital before start date.

feketemihai (changes requested):
The rest is ok, tested on runbot, so only two small isues..

customer_activity_statement/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from . import report
+from . import wizard
+from . import tests
```
feketemihai: One more thing, no need to import tests...

feketemihai (approved):

feketemihai:
@mreficent @hveficent  Thanks for the work and fast reaction to comments.

JordiBForgeFlow (approved):

JordiBForgeFlow:
Thank you very much for your in-depth review @feketemihai!

PR #390, opened 2018-02-27
[10.0] General ledger, foreign currency in initial and final balance

SimoRubi:
With this modifications, the General ledger shows foreign currency also for initial and final balance, only if all the lines summarized have foreign currency.
Please check, thanks

### DIFF
--- a/account_financial_report_qweb/__manifest__.py
+++ b/account_financial_report_qweb/__manifest__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '10.0.1.3.0',
+    'version': '10.0.1.4.0',
     'category': 'Reporting',
     'summary': 'OCA Financial Reports',
     'author': 'Camptocamp SA,'
--- a/account_financial_report_qweb/i18n/it.po
+++ b/account_financial_report_qweb/i18n/it.po
@@ -670,6 +670,12 @@ msgid "Final balance"
 msgstr "Bilancio finale"
 
 #. module: account_financial_report_qweb
+#: model:ir.model.fields,field_description:account_financial_report_qweb.field_report_general_ledger_qweb_account_final_balance_currency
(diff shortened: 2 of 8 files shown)


hurrinico (approved):
👍 imho

JordiBForgeFlow:
@SimoRubi Should not it be better to express the balance in currency when the account has a secondary currency? As odoo already forces all the moves to be using this secondary currency. See https://github.com/odoo/odoo/blob/11.0/addons/account/models/account.py#L58

SimoRubi:
Thanks @hurrinico for the review.

@jbeficent thanks I didn't think of that, it would sure be more efficient from the performance point of view; but the following use case wouldn't be covered: 
- account with no currency,
- moves in only one currency 

This might happen if you filter for a specific partner that has a currency for instance.

JordiBForgeFlow:
@SimoRubi I have proposed this PR https://github.com/SimoRubi/account-financial-reporting/pull/1

In the general ledger the totals in foreign currency are only computed if the account has a secondary currency. Otherwise the move lines will still show the amount expressed in foreign currency, but it is assumed that the account could contain various moves in different currencies, and will not calculate the total in the foreign currency.

Adds the foreign currency initial and ending balance to the trial balance.

JordiBForgeFlow:
@feketemihai I plan to cherry pick this PR (considering https://github.com/SimoRubi/account-financial-reporting/pull/1) to 11.0 once it is merged in 10.0

feketemihai:
@jbeficent Ok, i will review when i have time.

SimoRubi:
@jbeficent you (and others) managed to convince me that your way is better.
I've added the diff of your PR https://github.com/SimoRubi/account-financial-reporting/pull/1, I didn't merge it in my repo because it had strange commit history (in first commit you modified some code that was then deleted in second commit), hope it's ok for you now.

JordiBForgeFlow (approved):

feketemihai (approved):
Not tested.

feketemihai:
@jbeficent To be merged...

JordiBForgeFlow:
Is the XLS also covered? @hveficent can you migrate this feature to v11?

SimoRubi:
No @jbeficent I'm working on it and on open items report in https://github.com/OCA/account-financial-reporting/pull/407

PR #393, opened 2018-03-03
[11.0] [IMP] Update account_financial_report

feketemihai:
Update links in report, add account group file, update trial balance with hierarchy.


JordiBForgeFlow:
Nice... Thanks! Will review asap!

Cc @rgarnau @etobella

feketemihai:
@jbeficent Still have to adjust tests.

account_financial_report/report/templates/trial_balance.xml
```diff
@@ -27,40 +27,123 @@
...
+                            </t>
+                            <t t-if="line.account_group_id and len(line.code) == 3">
+                                <t t-set="style" t-value="'font-size: 9px; font-weight: bold; color: blue;'"/>
+                                <t t-set="padding" t-value="8"/>
+                            </t>
+                            <t t-if="line.account_group_id and len(line.code) == 4">
+                                <t t-set="style" t-value="'font-size: 9px; font-weight: bold; color: blue;'"/>
+                                <t t-set="padding" t-value="12"/>
+                            </t>
+                            <t t-if="line.account_group_id and len(line.code) == 5">
```
JordiBForgeFlow: I suggest that instead of hardcoding we set a account.financial.report.style model that allows to define the style to apply to each level, similar to how it is done in MIS Builder. The style would define the font, background color, padding, etc..

feketemihai: @jbeficent I agree with the style, but probably will be easier to have a dedicated module for styling on reporting-engine, so it can be inheritable...

feketemihai: @jbeficent @OCA/mis-builder-maintainers @OCA/reporting-engine-maintainers I can try to separate the style from mis_builder module into a separate module in reporting-engine and in mis_builder and  account_financial_report to use the style from newly created module. What do you think?

feketemihai: @jbeficent I suggest to leave it like this, until we will have a decision.

rgarnau:
Tested on runbot :+1:

rgarnau (approved):
Tested on runbot :+1:

feketemihai:
I will add tests for hierarchy_on on monday...

JordiBForgeFlow:
It does not work wit l10n_es. I think that you are doing some tricks with code_prefix... Are they really needed?

feketemihai:
@jbeficent When you choose the hierarchy on compute or on childs? Compute should take account that start with code_prefix, childs takes direct accounts linked with the account.group

feketemihai:
@jbeficent I think the problem is from l10n_es, the group should act as the view accounts, code_prefix is the code from view account, and name is the name of the account, in l10n_es it's put the code_prefix as name...

feketemihai:
Please see l10n_ro account group https://github.com/odoo/odoo/pull/22105/files#diff-0dc5f7b8a9c55073ee43f00597c76514 versus spain https://github.com/odoo/odoo/blob/11.0/addons/l10n_es/data/account_group.xml
The idea is if i take the name of the group, you will only have 103..

feketemihai:
@jbeficent Updated code to take name if code_prefix is not set, for compute and in report...

JordiBForgeFlow:
Thanks! We will send a PR to odoo to update the account groups in l10n_es. Cc @pedrobaeza

feketemihai:
@jbeficent can you check also the vat report?

feketemihai:
@jbeficent Until then, i updated to take value from name if code_prefix is empty, but as a design it will be much better with code_prefix and name different.

feketemihai:
@jbeficent @rgarnau @pedrobaeza @Yajo  Could you help me with little a thing, i modified the widget to have link to mulri record( e.g. account_move_line) that is launched from the amounts, after i changed the widget of the amount field to monetary, it seems that only one string is linked with the action (depending on country format, could be the currency or the amount - second i think in the json that is contructed) and for the first one i have the error:
"<function Home.web_client at 0x7fae14f9fe18>, /web: Function declared as capable of handling request of type 'http' but called with a request of type 'json'"
How i can include all the string from monetary field to be linkable with the action?

pedrobaeza:
I think it's a problem of the widget itself and you are not going to being able to fix it without fixing it upstream.

feketemihai:
@jholze I need you designing ideas to improve the styling of the reports

jholze:
@feketemihai i send already fo Ahmed but we are still very stressed because of a migration. But yes, would be nice to improve the layout for it.

feketemihai:
@jbeficent @rgarnau @mreficent @etobella @pedrobaeza @jholze I think i finish with this, waiting for reviews, style can be improved later.

JordiBForgeFlow (approved):
Great improvements!

JordiBForgeFlow:
Can you squash your commits?

feketemihai:
Yep, waiting for comments in case need to change something..

elicoidal (approved):
LGTM: impressive work
Functional tests OK

feketemihai:
@elicoidal It was, i hope i will see a lot of PR on Odoo with Localization update.

jholze:
👍

PR #106, opened 2015-08-05
[FIX] [7.0] Add reference field to account_financial_report_webkit_xls

JordiBForgeFlow:
This fix adds the field 'Reference'  associated to the account move line to 'account_financial_report_webkit_xls', so that the resulting layout matches with  'account_financial_report_webkit'.

PR #399, opened 2018-03-21
[ADD] Reference col to General Ledger

sebastiken:
Add reference column like un **8.0 branch**.


pedrobaeza:
Can you please give more background about the problem? This is a very old version, so it's not likely to be reviewed by others I'm afraid.

sebastiken:
I haven´t seen that #106 is the same **PR**. Maybe, it could be good to merge #106 and close this one without merge it.

feketemihai:
@pedrobaeza can you close this one as it's duplicate of #106.

PR #401, opened 2018-03-23
[11.0][IMP] account_financial_report: adapt the wizards to multicompany

MiquelRForgeFlow:
This PR adapts the `account_financial_report `'s wizards to multicompany.

In multicompany, the company of the wizard is visible and can be changed. Thus, if in the wizard the company is changed, all the other data in the wizard should be adapted accordingly.

### DIFF
--- a/account_financial_report/wizard/aged_partner_balance_wizard.py
+++ b/account_financial_report/wizard/aged_partner_balance_wizard.py
@@ -18,6 +18,7 @@ class AgedPartnerBalance(models.TransientModel):
     company_id = fields.Many2one(
         comodel_name='res.company',
         default=lambda self: self.env.user.company_id,
+        required=True,
         string='Company'
     )
     date_at = fields.Date(required=True,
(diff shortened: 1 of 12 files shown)


account_financial_report/wizard/aged_partner_balance_wizard.py
```diff
@@ -18,6 +18,7 @@ class AgedPartnerBalance(models.TransientModel):
     company_id = fields.Many2one(
         comodel_name='res.company',
         default=lambda self: self.env.user.company_id,
+        required=True,
```
elicoidal, line 21: What if you want to launch the report in multi-company (company=false)?

elicoidal (approved):
LGTM (with one question on an edge case)

account_financial_report/wizard/aged_partner_balance_wizard.py
MiquelRForgeFlow: In multi-company (company=false in the report), you will access only to data without company (thus, zero accounts because accounts always has a company).

pedrobaeza: But that's because it's no considered in the SQL. I'm with @elicoidal that we should allow that.

MiquelRForgeFlow: And what behavior would you like to have it if company=false in the report?

feketemihai: @elicoidal @pedrobaeza @mreficent The reports contains already a lot of data, just think of a general ledger printed on 3 companies, better to use them as company dependant, and from one single company at once...At least for my company, we have more than 200000 account move lines in 1 year...and we have 2 companies...

pedrobaeza: But then there's no possibility of obtaining consolidated data. Any way, the problem with this module is the SQLs. It should be done with ORM, and let record rules to act... Performance is not a problem if you use proper `read_group` clauses.

feketemihai: @pedrobaeza Agree, but let's target this one for next Odoo version, since it will take some work to do, and not the PR's target, @elicoidal Can you make a issue about this, multicompany aware, and probably on the wizard a many2many_tags to select which company do you want to print...

pedrobaeza: Yeah, but as everything needs to be redone to be adapted to multi-company, I think the best is to change also this. It's something that worths to invest in. Current code is hell to be debugged / modified.

MiquelRForgeFlow: I like the idea of a many2many_tags for the company

elicoidal: Side note that I will add in the issue, the report must be multi-currency and properly aggregate the currencies

PR #403, opened 2018-03-26
[11.0][IMP] account_financial_report - foreign_currency & journal ledger report

HviorForgeFlow:
Adds:
* General Ledger and Open Items display foreign currencies in reports.
* Trial Balance also display foreign currencies
* Adds journal ledger report (migrated from v10)


feketemihai:
@hveficent On v11, i changed currency_name to many2one, 
https://github.com/OCA/account-financial-reporting/blob/11.0/account_financial_report/report/general_ledger.py#L187
and updated the same xlsx report to print currency_id.name, and to have the link in the html report to it.

account_financial_report/report/general_ledger.py
```diff
@@ -91,9 +91,12 @@ class GeneralLedgerReportAccount(models.TransientModel):
     initial_debit = fields.Float(digits=(16, 2))
     initial_credit = fields.Float(digits=(16, 2))
     initial_balance = fields.Float(digits=(16, 2))
+    currency_name = fields.Char()
```
feketemihai: @hveficent Can you please check my comment, to be consistent we should have here a many2one for currency...and for all modifications, not a lot to be changed...Thanks.

account_financial_report/report/general_ledger.py
```diff
@@ -130,9 +133,12 @@ class GeneralLedgerReportPartner(models.TransientModel):
...
```
feketemihai: The same, many2one...

account_financial_report/report/general_ledger.py
```diff
@@ -285,7 +291,9 @@ def _get_account_sub_subquery_sum_amounts(
                 a.id AS account_id,
                 SUM(ml.debit) AS debit,
                 SUM(ml.credit) AS credit,
-                SUM(ml.balance) AS balance
+                SUM(ml.balance) AS balance,
+                c.name AS currency_name,
```
feketemihai: c.id as currency_id

feketemihai (changes requested):

feketemihai:
@hveficent I added just some of the remarks, can you please update the currency as many2one..

account_financial_report/report/templates/general_ledger.xml
```diff
@@ -378,6 +381,10 @@
                         </span>
                     </div>
                     <t t-if="line.currency_id">
+                        <!--## currency_name-->
```
feketemihai: @hveficent Why add this one, isn't it easier to have the monetary widget with currency_id,

account_financial_report/report/templates/general_ledger.xml
```diff
@@ -391,8 +398,17 @@
                         </div>
                     </t>
                     <t t-if="not line.currency_id">
-                        <!--## amount_currency-->
-                        <div class="act_as_cell"/>
+                        <div class="act_as_cell amount" style="width: 2.27%;"/>
+                        <t t-if="account_or_partner_object.currency_id">
+                            <!--## amount_currency-->
+                            <div class="act_as_cell amount" style="width: 5.67%;">
```
feketemihai: monetary widget

account_financial_report/report/templates/open_items.xml
```diff
@@ -179,6 +181,10 @@
                         <span t-field="line.amount_residual" t-options="{'widget': 'monetary', 'display_currency': res_company.currency_id}"/>
                     </div>
                     <t t-if="line.currency_id">
+                        <!--## currency_name-->
+                        <div class="act_as_cell amount">
```
feketemihai: No need for the name, the monetary will add currency symbol

feketemihai (changes requested):

account_financial_report/report/templates/open_items.xml
HviorForgeFlow: I'd rather add foreign currency abbreviation because an scenario could have 2 kind of dollar US ones and Canadian. With this column you can differ them.

feketemihai:
@hveficent i think you also have to modify xlsx reports.

HviorForgeFlow:
@feketemihai In which way? It seems to be shown as well as preview report.

feketemihai:
@hveficent i made 2 moves for agrolait on runbot with different currency, can you print now the general ledger, please see the initial balance, final balance is not printed with foreign currency, neither in html, pdf, excel, it should be printed, no, that's you added the fields?

HviorForgeFlow:
@feketemihai That feature is not performed yet. It is not trivial to find right currency out of move line, has_second_currency field only relate true or false if a there is some currency_id different from null. I could do a workaround with the assumption that every account only can have a second currency at all, not like this case:
image

feketemihai:
@hveficent Ok then, by default we have that constrain https://github.com/odoo/odoo/blob/11.0/addons/account/models/account_move.py#L484, but in many localization these are superscripted, sice you will have to create a lot of accounts...so fine by me.

feketemihai (approved):

JordiBForgeFlow:
@feketemihai by design the general ledger only prints the initial and ending balance in foreign currency when the account has a secondary currency.

feketemihai:
@jbeficent Yes, this it should only be in one currency, not like in the printscreen of @hveficent , same account has moves posted in 2 currencies..

HviorForgeFlow:
@feketemihai @jbeficent what about this way?
I've added a tick to enable foreign currency, if there is no currency_id related to that account balances won't be displayed

JordiBForgeFlow:
@feketemihai @hveficent Sorry but I am getting more and more confused.....

The rules are simple and we agreed when the PR was made in V10 with the people involved.

General Ledger --> Regardless if the account has secondary currency or not, display currency and amount_currency of all move lines. BUT display only starting and ending balance if the account has a secondary currency. 

For example if I have an account 'Accounts Payable' WITHOUT secondary currency I still want to display the purchases that I made in a foreign currency, showing the amount in the foreign currency, but 'Accounts Payable' can potentially display purchases that I made in EUR, USD, CHF,.... I will never expect to have a starting and ending balance in foreign currency!

If I have a bank account that I hold in swiss francs (CHF) as secondary currency, by setting the secondary currency on the account Odoo will require that all move lines associated to this account include the currency CHF and an amount in that currency. When I print the general ledger for this account I EXPECT to view the starting end ending balances for that account also in CHF!

For trial balance we should only display the balance in foreign currency for accounts that have a specific secondary currency, for the same reasons.

Is that more clear now?

HviorForgeFlow:
All changes I've made aim these points at general ledger report.

feketemihai:
@hveficent Will you do it on trial balance as well in this PR, or a separate one?

HviorForgeFlow:
@feketemihai there it is.

JordiBForgeFlow:
I see a discrepancy between the general ledger as display in View/Qweb vs. XLSX.

For example, Accounts Payable is set with secondary currency EUR. In View/Qweb it totalizes for this currency:
image

In XLSX it does not:
image

HviorForgeFlow:
@jbeficent could it be labeled as needs review? Thanks

JordiBForgeFlow:
The decimal accuracy in the Excel format is not correct. It is correct in the View/PDF outputs.
For example, I am testing with cry ETH where I indicated an accuracy of 10 decimal places.

See in the view/pdf:
image


See in XLSX:
image

account_financial_report/report/general_ledger.py
```diff
@@ -540,9 +559,15 @@ def _get_partner_sub_subquery_sum_amounts(
...
-                SUM(ml.balance) AS balance
+                SUM(ml.balance) AS balance,
+                c.id as currency_id,
+                SUM(ml.amount_currency) AS balance_currency
             FROM
                 accounts_partners ap
+            INNER JOIN account_account ac
+                ON ac.id = ap.account_id
+            LEFT JOIN
+                res_currency c ON ap.account_id = c.id
```
SimoRubi: I think this is not correct: you are matching account_id with currency_id..?

account_financial_report/report/general_ledger_xlsx.py
```diff
@@ -125,6 +138,18 @@ def _generate_report_content(self, workbook, report):
             # 2 lines break
             self.row_pos += 2
 
+    def write_initial_balance(self, my_object, type_object):
```
SimoRubi: It could be nice to perform this check on the object itself, instead of passing the string parameter.
You can try something like `my_object._model == 'res.partner'`

SimoRubi (changes requested):
Nice work on the number formats!

HviorForgeFlow:
@jbeficent @feketemihai @SimoRubi Could it have a final review?

SimoRubi (approved):
Code review, tested on runbot, looks good to me

JordiBForgeFlow (approved):

PR #408, opened 2018-04-04
[10.0][account_financial_report_qweb] fixes the reporting of unaffected earnings account

JordiBForgeFlow:
Prior to this PR the unaffected earnings account was incorrectly being reported in the general ledger and trial balance.

This PR simplifies and fixes the logic to report on the unaffected earnings account.

### DIFF
--- a/account_financial_report_qweb/report/general_ledger.py
+++ b/account_financial_report_qweb/report/general_ledger.py
@@ -255,12 +255,6 @@ class GeneralLedgerReportCompute(models.TransientModel):
             if self.centralize:
                 self._inject_line_centralized_values()
 
-        # Complete unaffected earnings account
-        if (not self.filter_account_ids or
-                self.unaffected_earnings_account.id in
-                self.filter_account_ids.ids):
-            self._complete_unaffected_earnings_account_values()
-
         if with_line_details:
             # Compute display flag
             self._compute_has_second_currency()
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
cc @pedrobaeza

pedrobaeza:
Tests for unaffected earnings are failing due to the changed data.

account_financial_report_qweb/report/general_ledger.py
```diff
@@ -255,11 +255,11 @@ def compute_data_for_report(
             if self.centralize:
                 self._inject_line_centralized_values()
 
-        # Complete unaffected earnings account
-        if (not self.filter_account_ids or
-                self.unaffected_earnings_account.id in
-                self.filter_account_ids.ids):
-            self._complete_unaffected_earnings_account_values()
+        # # Complete unaffected earnings account
```
pedrobaeza: Don't comment code, but remove it

pedrobaeza (changes requested):

pedrobaeza:
@jbeficent I have tried the change, and on the general ledger, it seems very weird to present the debit and credit part, as they are going to increase always to amounts that are very uncomfortable to be seen. Check the screenshot:

seleccion_008

I think only balance should be presented there. I'm also not sure about the numbers, but let me check this deeper.

pedrobaeza (approved):

rafaelbn (approved):

PR #412, opened 2018-04-17
[FIX] account_financial_report_qweb: unaffected earnings account

pedrobaeza:
Backport of #408 

@Tecnativa


pedrobaeza:
It seems to be not needed.

PR #415, opened 2018-04-27
[8.0][FIX]pre init hook causing error

AaronHForgeFlow:
It's not a common case. But if there's any error when installing (e.g I'm installed other module depending on this one) I got that the column created in the script remains there so I'm not able to install the module.


pedrobaeza:
Travis is red, and better to make the reverse `if` and return for avoiding a huge diff.

MiquelRForgeFlow (approved):

PR #420, opened 2018-05-24
wip - [7.0][IMP] account_financial_report_webkit general_ledger reduce sql …

tfossoulw:
…query nbr

This redure drastically the number of sql query to the database so optimise the report generation.

### DIFF
--- a/account_financial_report_webkit/report/common_reports.py
+++ b/account_financial_report_webkit/report/common_reports.py
@@ -439,7 +439,7 @@ class CommonReportHeaderWebkit(common_report_header):
     ################################################
     # Account move retrieval helper                #
     ################################################
-    def _get_move_ids_from_periods(self, account_id, period_start, period_stop,
+    def _get_move_ids_from_periods(self, account_ids, period_start, period_stop,
                                    target_move):
         move_line_obj = self.pool.get('account.move.line')
         period_obj = self.pool.get('account.period')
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is changing some methods signature, it's a very old version and Travis is red, so I don't think this is going to be merged easily, as for sure no reviewers will arise. I think you should better focus on migrate your customers to newer versions.

Any way, I let it opened to see a bit.

@hbrunn what do you think about this?

hbrunn:
I don't think we should demotivate people working on old versions. But as you I'd be very cautious to merge this, because we don't have proper tests for 7.0. And I myself won't review it because I'm done with 7.0 indeed.

pedrobaeza:
It's not to demotivate, but being realistic. This is changing API signatures, and other modules depending on this one (OCA or not), might break after this change. The risk can be taken over a version well maintained (v10, v11, even v8...), but not v7

tfossoulw:
I could also keep the method signature intact but adapt the method to allow 1 or more "account_id".
Then other modules depending on this one shouldn't be impacted by this commit. ?

pedrobaeza:
There is the same problem, as if you pass several accounts, other modules won't process this multi-record properly.

tfossoulw:
No, you allow to pass either one or several account. So other modules could continue to work with it.

agb80:
As v8 and v7 versions for impacted module are pretty similar another approach could be create the PR for v8 that have many more test and when accepted for v8 make a backport for v7.

My 2cents

PR #428, opened 2018-06-21
[11.0][customer_activity_statement] extend to display receivables and payables

JordiBForgeFlow:
It is straightforward to provide the option to display also the payables of a partner. A pitty that the report was not named partner_activity_statement....

### DIFF
--- a/customer_activity_statement/README.rst
+++ b/customer_activity_statement/README.rst
@@ -3,17 +3,17 @@
    :alt: License: AGPL-3
 
 =================================
-Print Customer Activity Statement
+Print Partner Activity Statement
 =================================
 
-The activity statement provides details of all activity on the customer receivables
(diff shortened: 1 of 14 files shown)


MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: Tested on local and working as expected

pedrobaeza:
If you change the scope, shouldn't the module should be partner_activity_statement?

JordiBForgeFlow:
@pedrobaeza but perhaps it's better to rename the module in version 12?

etobella (approved):
Function test ok on runbot :+1:

pedrobaeza:
Please take note on the README then

MiquelRForgeFlow:
@pedrobaeza  done

PR #359, opened 2017-10-16
[10.0][FIX] account_financial_report_qweb: date condition in open items report.

adrienpeiffer:
This commit includes move lines on the report date by default. These lines shouldn't be considered in the futur.

PR #429, opened 2018-06-22
[FIX] account_financial_report_qweb: date condition in open items rep…

lfreeke:
…ort.

This commit includes move lines on the report date by default. These lines shouldn't be considered in the futur.

based on https://github.com/OCA/account-financial-reporting/pull/359


MiquelRForgeFlow:
Already fixed in https://github.com/OCA/account-financial-reporting/pull/439. This PR can be closed.

PR #433, opened 2018-06-28
[ADD] add fix on account-financial-reporting for trial balance report…

PieterPaulussen:
Group the balance calculation by account instead of calculating the balance for each account separately. This should prove to be quicker.

Also kept backwards compatibility when a single account entry is requested.

Credits to @StefanRijnhart for the technical suggestions.


StefanRijnhart (commented):
Thanks! The module for 9.0 and up is refactored completely and includes a similar grouping but this is a simple and backwards compatible way to enjoy a 6 fold or more performance improvement in the 8.0 series.

Issue #430, opened 2018-06-25
account_financial_report_qweb: CSS conflicts with backend


yajo:
STR:

1. Install odoo 10.
1. Install `account_financial_report_qweb`
1. Go edit some view in configuration > Technical > user interface > views.

Problem: Fonts should be monospace, but are Helvetica:
![captura de pantalla de 2018-06-25 12-43-39](https://user-images.githubusercontent.com/973709/41848633-f7071564-7875-11e8-8416-b1208c9389b4.png)

I'd say this addon is injecting some report-only layout fixes in backend client.

@Tecnativa

pedrobaeza:
@jbeficent was it you who changes this?

pedrobaeza:
@Yajo are you able to fix it?

yajo:
I guess so, but the problem is that I don't really know what that addon does, and I saw it includes report things (expected) and backend widgets (not expected). I'm not sure if moving the CSS to reports only would break the widgets...

I guess that somebody that actively uses the addon could be better off fixing it, but yes, I could try anyways.

pedrobaeza:
@jbeficent please help here at least to answer @Yajo's question.

HviorForgeFlow:
On statics/css/src/*.css there is a definition on top about font usage, in
that case Helvetica or Arial, removig that css definition would be
sufficient to display fonts as default/¿monospace?

yajo:
I think the bigger problem is including report css files in backend assets, isn't it?

HviorForgeFlow:
Yes, I noticed in v11 is happening too.

HviorForgeFlow:
Solved with #437 waiting FF in #436 
Issue should be closed after that

pedrobaeza:
Already solved

PR #435, opened 2018-07-04
[10.0][FIX] account_financial_report_qweb: css injection assets backend

HviorForgeFlow:
#430

### DIFF
--- a/account_financial_report_qweb/view/report_template.xml
+++ b/account_financial_report_qweb/view/report_template.xml
@@ -4,7 +4,6 @@
     <template id="account_financial_report_assets_backend"
               name="account_financial_report assets" inherit_id="web.assets_backend">
         <xpath expr="." position="inside">
-            <link href="/account_financial_report_qweb/static/src/css/report.css" rel="stylesheet"/>
             <script type="text/javascript"
                     src="/account_financial_report_qweb/static/src/js/account_financial_report_qweb_backend.js"/>
             <script type="text/javascript"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
No more is needed?

Please change module version number

HviorForgeFlow:
I don't think so, it shouldn't show css conflicts anymore.

account_financial_report_qweb/view/report_template.xml
```diff
@@ -4,7 +4,6 @@
     <template id="account_financial_report_assets_backend"
               name="account_financial_report assets" inherit_id="web.assets_backend">
         <xpath expr="." position="inside">
-            <link href="/account_financial_report_qweb/static/src/css/report.css" rel="stylesheet"/>
```
yajo, line 7: AFAICS ths file is not being used anymore, so you could remove it, right?

Or, if it is still needed for reports, then move it to some `assets_report` view.

yajo (changes requested):

account_financial_report_qweb/view/report_template.xml
HviorForgeFlow: It is used here: https://github.com/OCA/account-financial-reporting/blob/25e705cf833ac68bc3630d69ca2934f1e08de987/account_financial_report_qweb/report/templates/layouts.xml#L4-L8

yajo (approved):

pedrobaeza (approved):

pedrobaeza:
Thanks for this. Please forward-port it to v11.

pedrobaeza:
I think this has broken the view in screen option. Can it be?

PR #436, opened 2018-07-04
[11.0] [FIX] account_financial_report: css injection assets backend

HviorForgeFlow:
FF


pedrobaeza:
I'm afraid this is breaking the live view:
seleccion_036

HviorForgeFlow:
FF already solved in 10.0 #437

pedrobaeza (approved):
Fast-tracking this

PR #437, opened 2018-07-05
[10.0] [FIX] account_financial_report_qweb: css injection assets backend

HviorForgeFlow:
fixing a mess #435


alexis-via (approved):
I confirm it fixes the layout bug when using the "View" button for Balance.

alexis-via:
Mmm, I manage to see it working with this PR once... but now, I can't make it work again at other Odoo installations that I updated with this new code even when using private navigation session with my web browser... so not sure that this PR fully fixes the layout bug when using the "View" button for a Balance report.

HviorForgeFlow:
It doesn't I forgot a commit, damn.
Thanks for reviewing,

PR #438, opened 2018-07-05
[ADD] [account_financial_report_horizontal] special treatment for profit&loss reports

hbrunn:
as the account types for profit and loss are of type `other` in v10, profit and loss will be duplicated (see
[Financial.report.pdf](https://github.com/OCA/account-financial-reporting/files/2165479/Financial.report.pdf)), so we need special treatment for this reports.

### DIFF
--- a/account_financial_report_horizontal/models/account_financial_report.py
+++ b/account_financial_report_horizontal/models/account_financial_report.py
@@ -41,6 +41,14 @@ class AccountFinancialReport(models.Model):
             for report in reports:
                 if not report.parent_id:
                     result += report
+                # special treatment for profit and loss if they have the
+                # standard configuration type='other'
+                elif side == 'right' and report == self.env.ref(
+                        'account.account_financial_report_income0'
+                ) or side == 'left' and report == self.env.ref(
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

jado95 (changes requested):
@hbrunn please note that you have to reverse right and left. On right account.account_financial_report_expense0 and on left account.account_financial_report_income0.

hbrunn:
note that this is a negation. On runbot, we get [Financial report.pdf](https://github.com/OCA/account-financial-reporting/files/2235500/Financial.report.pdf) which is what we want for the Netherlands at least.

sbidoul:
Wondering how hard it would be to add some layout options to MIS builder.
We already have a landscape pdf option, we could add a 2 columns options.
Plus a page break / column break before in kpi styles.

Would that be sufficient to replace this module?

hbrunn:
I think this might just work, but then I'd still have this module provide the appropriate configuration for the MIS builder ;-)

StefanRijnhart (approved):
Thanks!

PR #441, opened 2018-07-17
[9.0][IMP] Improve transient models cleaning

p-tombez:
When there are a lot of ```account.move.line``` (several millions) and print any of the Qweb reports, that will generate also a lot of transient objects.
As these objects are created with an "insert" query, the cleaning normally triggered by the count of the records in transient tables is not done, so only the cleaning based on the age of the records is processed (by default, records older than 1 hours are deleted), but the cron task is only ran one time per day.
With all these things in mind, it's pretty easy to see that transient data could get out of control, leading to errors like : 
```
Call of self.pool.get('ir.autovacuum').power_on(cr, uid, *()) failed in Job 38
 File "/opt/odoo/src/openerp/addons/base/ir/ir_cron.py", line 128, in _callback
 getattr(model, method_name)(cr, uid, *args)
 File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
 return old_api(self, *args, **kwargs)
 File "/opt/odoo/src/addons/mail/models/ir_autovacuum.py", line 11, in power_on
 return super(AutoVacuum, self).power_on(cr, uid, *args, **kwargs)
 File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
 return old_api(self, *args, **kwargs)
 File "/opt/odoo/src/openerp/addons/base/ir/ir_autovacuum.py", line 31, in power_on
 self._gc_transient_models(cr, uid, *args, **kwargs)
 File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
 return old_api(self, *args, **kwargs)
 File "/opt/odoo/src/openerp/addons/base/ir/ir_autovacuum.py", line 18, in _gc_transient_models
 model._transient_vacuum(cr, uid, force=True)
 File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
 return old_api(self, *args, **kwargs)
 File "/opt/odoo/src/openerp/models.py", line 5214, in _transient_vacuum
 self._transient_clean_rows_older_than(cr, self._transient_max_hours * 60 * 60)
 File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
 return old_api(self, *args, **kwargs)
 File "/opt/odoo/src/openerp/models.py", line 5176, in _transient_clean_rows_older_than
 cr.execute(query, ("%s seconds" % seconds,))
 File "/opt/odoo/src/openerp/sql_db.py", line 154, in wrapper
 return f(self, *args, **kwargs)
 File "/opt/odoo/src/openerp/sql_db.py", line 233, in execute
 res = self._obj.execute(query, params)
DatabaseError: out of memory for query result
```


account_financial_report_qweb/report/abstract_report.py
```diff
@@ -0,0 +1,22 @@
...
+class AbstractReport(models.AbstractModel):
+    _name = 'report_qweb_abstract'
+
+    def _transient_clean_rows_older_than(self, cr, seconds):
+        assert self._transient, \
+            "Model %s is not transient, it cannot be vacuumed!" % self._name
+        # Never delete rows used in last 5 minutes
+        seconds = max(seconds, 300)
+        query = """
+DELETE FROM """ + self._table + """
```
guewen, line 17: Did you tested how much time it takes for deleting 80 millions of rows? (versus TRUNCATE CASCADE)

guewen: Not as much as I expected actually:

```
delete from report_general_ledger_qweb_move_line;
DELETE 86000001
Time: 74372.450 ms (01:14.372)
```
Admittedly my query does not include the where clause but at worse it adds a few seconds.
A couple of minutes for the vacuum is probably acceptable.

guewen (approved):

Tonow-c2c (approved):
LGTM

MiquelRForgeFlow:
I think this should include migration scripts

StefanRijnhart:
@mreficent what do you think needs to be handled by a migration script? Looks to me like this just overrides a bit of code from models.TransientModel.

StefanRijnhart (approved):

MiquelRForgeFlow:
Oh, right!

MiquelRForgeFlow (approved):

account_financial_report_qweb/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '9.0.1.0.2',
+    'version': '9.0.1.0.3',
```
MiquelRForgeFlow: now is 9.0.1.0.3 if you rebase. So it should be 9.0.1.0.4 or 9.0.1.1.0

StefanRijnhart: I updated the version number in the Github UI

MiquelRForgeFlow:
This fix should be forward-ported to the other versions, right?

StefanRijnhart:
I was already on it! Here is a port to 10.0: https://github.com/OCA/account-financial-reporting/pull/455

Issue #443, opened 2018-07-18
[10] customer_outstanding_statement - Error on SQL   


MarcoCalcagni:
Hi, i found an error on calculation of debit value. 
If I have an invoice of 1000 and on this invoice I have 2 payment of 200 balance 600 the report will show import of document 2000 (because on the account_move_line  there are more account_partial_reconcile record ) 
to fix I think to change 
`CASE WHEN (l.currency_id is not null AND l.amount_currency > 0.0)
                THEN sum(l.amount_currency)
                ELSE sum(l.debit)
            END as debit,` 
in 
`CASE WHEN (l.currency_id is not null AND l.amount_currency > 0.0)
                THEN avg(l.amount_currency)
                ELSE avg(l.debit)
            END as debit,` 

Marco 
version 10 - module customer_outstanding_statement  - repo account-financial-reporting 
file customer_outstanding_statement.py

gdgellatly:
@As400it This can be closed

MarcoCalcagni:
yes

PR #446, opened 2018-07-19
[FIX] wrong sql calculation on credit and debit value

MarcoCalcagni:
https://github.com/OCA/account-financial-reporting/issues/443


JordiBForgeFlow (approved):

JordiBForgeFlow:
This is good. I'm fast tracking, as this is a clear bug. Thanks for reporting!

PR #448, opened 2018-07-26
[10.0] [FIX] readd_css_link_to_report_template.xml

HviorForgeFlow:
Due to #437


pedrobaeza:
But restoring this, we go back to the source problem: ACE editor gets the incorrect font.

HviorForgeFlow:
In commit https://github.com/OCA/account-financial-reporting/pull/437/commits/c049016827053d8beb0034a50d9bc9c9d0dfb302 I moved font-family to a more specific class, I think this has fix fix the problem, but I forget readding link to resporte_template which is responsible of render all reports.

pedrobaeza (approved):
Ah, OK. Fast tracking this then for fixing it.

PR #454, opened 2018-08-20
[10.0] [IMP/ADD] account_financial_report_qweb: Improved the report functionality.

Chandresh-SerpentCS:
### General Ledger report without group by partner*

- You have the selection to print the general ledger report now without group by the partner.

- Added one boolean field in the report wizard. If it is checked then report will be print as per its default behavior. **If not checked then it will be print as per the new functionality. (Removed separate table of the partner for the same account from the report.)** 

*New feature available only for the PDF & View.


JayVora-SerpentCS:
@Chandresh-SerpentCS has signed already. I sent email long back to OCACLA

PR #456, opened 2018-08-21
[11.0][IMP] Improve transient models cleaning

StefanRijnhart:
When there are a lot of account.move.line (several millions) and print any of
the Qweb reports, that will generate also a lot of transient objects.
As these objects are created with an "insert" query, the cleaning normally
triggered by the count of the records in transient tables is not done, so only
the cleaning based on the age of the records is processed (by default, records
older than 1 hours are deleted), but the cron task is only ran one time per
day. For large setups this can lead to memory errors at that point. This change
prevents the memory error by executing the transient record cleanup for the
report models in this module in SQL.

Backport of https://github.com/OCA/account-financial-reporting/pull/441

### DIFF
--- a/account_financial_report/__manifest__.py
+++ b/account_financial_report/__manifest__.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Account Financial Reports',
-    'version': '11.0.2.2.0',
+    'version': '11.0.2.2.1',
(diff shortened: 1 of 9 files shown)


MiquelRForgeFlow (approved):

pedrobaeza (approved):

pedrobaeza:
For v12 we should definitively remove this ugly SQL/INSERT querys and use instead SQL views / `read_group` calls

PR #457, opened 2018-08-22
Bug fix

gdgellatly:
@mreficent @pedrobaeza @StefanRijnhart Your urgent attention please.

I somehow didn't test the conversion of the options dict from JSON to Python.  I was using --dev=reload etc but it seems it didn't take so there is an error as currency is an object not a string in the widget.


gdgellatly:
grr hold on.
OK now, had an extra file in there.

pedrobaeza:
OK, fast-tracking this.

StefanRijnhart:
Thanks!

PR #453, opened 2018-08-17
11.0 Activity Statement Report cleanup

gdgellatly:
The existing report was full of manually applied css that didn't play nicely with the Odoo layout options and was very repetitive.

* This PR pares it right back and adds a less file for fixes to make it look consistent, yet usable with all report style options to provide a better default experience.  
* Additionally the filename is changed from `Statement Action to PDF` to just `Activity Statement`.
* Duplicated template stuff encapsulated in if clauses simplified.
* Tidied up aging box.
* Added missing tbody tags - sorry it makes for a large diff
* Replace t-esc-options JSON with t-options Python

Note I just realised I did the demo files using a custom module.  The only difference is the custom module ages by months. No layout difference at all.  This can be seen at https://github.com/odoonz/odoonz-addons/tree/11.0/customer_activity_statement_monthly.

For review comparison files are included below.

[Statement Action to PDFnew-clean.pdf](https://github.com/OCA/account-financial-reporting/files/2295562/Statement.Action.to.PDFnew-clean.pdf)
[Statement Action to PDFnew-boxed.pdf](https://github.com/OCA/account-financial-reporting/files/2295563/Statement.Action.to.PDFnew-boxed.pdf)
[Statement Action to PDFnew-bg.pdf](https://github.com/OCA/account-financial-reporting/files/2295564/Statement.Action.to.PDFnew-bg.pdf)
[Statement Action to PDFnew-std.pdf](https://github.com/OCA/account-financial-reporting/files/2295565/Statement.Action.to.PDFnew-std.pdf)
[Statement Action to PDFexisting-clean.pdf](https://github.com/OCA/account-financial-reporting/files/2295566/Statement.Action.to.PDFexisting-clean.pdf)
[Statement Action to PDFexisting-boxed.pdf](https://github.com/OCA/account-financial-reporting/files/2295567/Statement.Action.to.PDFexisting-boxed.pdf)
[Statement Action to PDFexisting-bg.pdf](https://github.com/OCA/account-financial-reporting/files/2295568/Statement.Action.to.PDFexisting-bg.pdf)
[Statement Action to PDFexisting-std.pdf](https://github.com/OCA/account-financial-reporting/files/2295569/Statement.Action.to.PDFexisting-std.pdf)

PR #458, opened 2018-08-22
Format Improvements for customer outstanding statement

gdgellatly:
This is essentially the same as #453 except during testing I noticed:

* This report was based off report_overdue which hasn't had much new api love, so converted the manual address formatting to use the address tag same as invoice report.
* In testing #453 I was using a custom aging module, turns out the standard aging headers are too long, so removed Due from the `days` columns to stop wrapping.

Because this report has 2 extra columns in the top part, it will still wrap "Original Amount" on some formats - there is just no space unless we drop the word amount.

### DIFF
--- a/customer_outstanding_statement/__manifest__.py
+++ b/customer_outstanding_statement/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Partner Outstanding Statement',
-    'version': '11.0.2.0.0',
+    'version': '11.0.2.1.0',
     'category': 'Accounting & Finance',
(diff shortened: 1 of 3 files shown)


gdgellatly:
https://github.com/odoo/enterprise/pull/2635/files#diff-0 Note this bug - todo

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
cc: @StefanRijnhart @pedrobaeza

JordiBForgeFlow (approved):

PR #445, opened 2018-07-19
10.0 imp account financial report qweb

mpanarin:
Allows filtering by account journal in general ledger and trial balance

PR #460, opened 2018-09-05
[10.0] report_qweb: usability improvements

alexis-via:
Fix view layout issue introduced recently by PR https://github.com/OCA/account-financial-reporting/pull/445

Improve workaround for bug https://github.com/odoo/odoo/issues/14761 (bug with M2M + onchange in wizards): it replaces the workaround widget="many2many_tags" on field name="account_ids" which prevented from selecting several accounts at the same time (quite useful when you want to select an interval of accounts for example). We now use the regular M2M widget and inherit create().

### DIFF
--- a/account_financial_report_qweb/wizard/aged_partner_balance_wizard.py
+++ b/account_financial_report_qweb/wizard/aged_partner_balance_wizard.py
@@ -54,6 +54,27 @@ class AgedPartnerBalance(models.TransientModel):
         else:
             self.account_ids = None
 
+    @api.model
+    def create(self, vals):
+        """
+        This is a workaround for bug https://github.com/odoo/odoo/issues/14761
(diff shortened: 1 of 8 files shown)


alexis-via:
Something very strange about the M2M with regular widget: the widget for "account_ids" is readonly if you don't have create rights on account.account. And, by default, account.group_account_user only has read rights on account.account (only account.group_account_manager has create rights).
If someone has an idea about this problem... please help !

florian-dacosta (approved):
I use a similar fix in production in version 9 for a customer for 1 year and all is good so far

JordiBForgeFlow (approved):
:+1: Technical and functional review.

During testing I found that the filter by journal is not really filtering.... but not related to this PR. So I will analyse further.

PR #468, opened 2018-10-01
[12.0][MIG] account_tax_balance

astirpe:
Depends on:

- [x] date_range: https://github.com/OCA/server-ux/pull/25


astirpe:
Note: I added line https://github.com/OCA/account-financial-reporting/pull/468/commits/f0d532fc2358eb764a59d0235c44c67f22317346#diff-667f500e2abebc5b20fd927192aad5b5R150 in the tests. The reason is that, without that line, the computation of `_compute_balance()` is not automatically triggered. Not sure whether this is a possible regression present in Odoo 12.0.

account_tax_balance/models/account_move.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = 'account.move'
+
+    move_type = fields.Selection(
+        string="Move type", selection=[
```
mpanarin: please consider moving the selection to the separate method for better extensibility

account_tax_balance/models/account_tax.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @api.multi
+    def _compute_has_moves(self):
+        ids_with_moves = set(self._account_tax_ids_with_moves())
+        for tax in self:
+            tax.has_moves = tax.id in ids_with_moves
+
+    @api.model
+    def _search_has_moves(self, operator, value):
+        if operator != '=' or not value:
```
mpanarin: please consider moving the `'='` to a separate method or constant for ease of extensibility

mpanarin (approved):

.travis.yml
```diff
@@ -17,13 +16,9 @@ python:
 
 env:
   global:
-  - VERSION="12.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
-  - TRANSIFEX_USER='transbot@odoo-community.org'
-  - secure: MZ5W99aYtT5mntw9Aur7bZ0FFXHZYc3v2oWg9dEODeKcLWUTmQbPC/T/pcjex1MBQTIHy0GZ+M/kj7qQQilbtbMYnjt7VBcY+fY1cZskEohc7mK/c8sUNHeUYj7Ec7n7VriY0XjQVH7Ov3FkiDoUm/4um0VyhH8LpKWw0vMhqFY=
-  
+  - VERSION="12.0" TESTS="0" LINT_CHECK="0"
```
tafaRU, line 19: @astirpe, I think it should be better that you move the changes related to `TRANSIFEX` onto a separate PR.
While that one related to ` postgresql` is ok as it's specific to the current module.

Thank you!

astirpe: @tafaru I believe that the postgres 9.2 does not fit with Odoo 12.0, it seems that the minimum required version of postgres is 9.5. So IMO we could create one separate PR for both postgres version and transifex removal.

I'm at the moment busy at the Odoo Experience. If you want and can, you could open the PR and I will approve it quickly. Otherwise I will do asap.

Thank you

pedrobaeza:
I have rebased the branch, squashed old translation commits (now with Weblate we preserve the authorship, so no squashing), and remove temporary dependency branch now that module has been merged.

I don't see problem to have that .travis.yml thing for making this work. When migration to 13.0, that commit won't be preserved as the fetch is by folder. Actually, it has been my bad, adding the line for PG without replacing old ones.

.travis.yml
tafaRU: @astirpe, I resolve this conversation as https://github.com/OCA/account-financial-reporting/pull/468#issuecomment-426932450
Enjoy your Odoo Experience!

tafaRU:
ok @pedrobaeza, thanks for your clarification.

> adding the line for PG without replacing old ones.

Will these lines be removed automatically by the OCA-git-bot?

pedrobaeza:
> Will these lines be removed automatically by the OCA-git-bot?

Is it something very general? If so, I can develop an script, but most of the repositories doesn't have this line, relying on default BD version.

tafaRU:
> Is it something very general?

I noticed this in:

* https://github.com/OCA/l10n-italy/blob/12.0/.travis.yml
* https://github.com/OCA/account-financial-reporting/blob/12.0/.travis.yml

bud I didn't investigate further.

pedrobaeza:
OK, for now we change it manually and if it's pathological, I develop something.

pedrobaeza:
Merging this one as everything is correct

PR #469, opened 2018-10-01
[10.0] account_financial_report_qweb: LLN Code sprint 2018

alexis-via:
Trial Balance:
- Add column 'Period balance' (= debit - credit)
- Change option 'Hide account ending balance at 0' to 'Hide accounts at 0'


account_financial_report_qweb/report/trial_balance.py
```diff
@@ -247,9 +251,16 @@ def _inject_account_values(self, account_ids):
 WHERE
     acc.id in %s
         """
-        if self.hide_account_balance_at_0:
-            query_inject_account += """ AND
-    final_balance IS NOT NULL AND final_balance != 0"""
+        if self.hide_account_at_0:
+            query_inject_account += """ AND (
+    (initial_balance IS NOT NULL AND initial_balance != 0) OR
```
pedrobaeza, line 256: I wouldn't test about initial balance and final_balance, as you can have an initial balance but no activity on the period (so you will also have final balance), but it's not too much relevant for the trial balance. Don't you think?

rafaelbn:
@alexis-via is this PR needed after https://github.com/OCA/account-financial-reporting/pull/485 ?

pedrobaeza:
No, it's not. Closing.

alexis-via:
Yes. But there are still a number of fixes/improvements that are present here but not in #485... we need to grab them here and push in 485.

Issue #419, opened 2018-05-19
[11.0] Feedback on hierarchy for account_financial_report


pedrobaeza:
HI, @feketemihai @jbeficent, I would like to share with you several things after testing `account_financial_reporting` to see possible fixes/improvements:

* If you mark "Hide account ending balance at 0", indeed the accounts are hidden, but not their hierarchy parents, getting an ugly report:

![seleccion_010](https://user-images.githubusercontent.com/7165771/40270859-e7fc4254-5b94-11e8-9d01-08b88e0f54b3.png)

* The blue color and font gets a bit weird for seeing on PDF IMO.

* The option "Hierarchy On" Computed Accounts / Child Accounts gets without meaning, at least in our chart template. A help tooltip would help for knowing what is this.

* But the option I imagine for the hierarchy is the level of hierarchy to group (0 - no group, so the same as previously when no hierarchy exists, 1, 2, 3...) to specify the detail level.

* Another option is to simply not showing account, if you want a summary at one of the levels indicated before.

What do you think about this?

pedrobaeza:
@feketemihai I haven't received any reply from you, and haven't seen you around for a while. Are you OK?

lauracvilla-zz:
There is an option to get Hierarchy or Account Parents Information con Odoo 10 with account_financial_report? I like this module but I don't know how show parents account information on this module or how I need to create this accounts? Thanks for your help.

pedrobaeza:
@lauracvilla no. Please don't hijack threads.

PR #470, opened 2018-10-03
[11.0]remove lines with balance 0 on trial balance.

jesusVMayor:
When hide_account_balance_at_0 is checked remove the lines of groups with final_balance = 0.

I decided to make the delete instead of not create the records because the insert just takes the data from account_group table.
Another possibility is maybe make the insert and the _update_account_group_child_values/_update_account_group_computed_values at the same query and don't create the records. But i don't know if make the insert and after the update has some purpose.
cc: @pedrobaeza 
https://github.com/OCA/account-financial-reporting/issues/419

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -232,6 +232,11 @@ class TrialBalanceReportCompute(models.TransientModel):
         else:
             for line in self.account_ids:
                 line.write({'level': 0})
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@jbeficent can you please review this as you are more familiar with the structure?

jesusVMayor:
Improvement added by @jbeficent on https://github.com/OCA/account-financial-reporting/pull/478

PR #471, opened 2018-10-07
[12.0][FIX] account_tax_balance: at_install/post_install test decorators

astirpe:
Following odoo/odoo#27471, here is the patch for the `at_install`/`post_install` test decorators, the V12.0 way.


pedrobaeza (approved):
Fast-tracking as tests are not being executed.

Issue #462, opened 2018-09-18
11.0:account_financial_report -  res = {**res, **foreign_currency}  SyntaxError: invalid syntax


miquelalzanillas:
This error related to foreign_currency avoid to initialize the addon.

2018-09-18 07:12:58,470 27465 DEBUG ? odoo.http: Loading account_financial_report
2018-09-18 07:12:58,486 27465 WARNING ? py.warnings: /opt/odoo/odoo11-venv/lib/python3.4/site-packages/werkzeug/filesystem.py:63: BrokenFilesystemWarning: Detected a misconfigured UNIX filesystem: Will use UTF-8 as filesystem encoding instead of 'ascii'
  BrokenFilesystemWarning)

2018-09-18 07:12:58,615 27465 INFO ? werkzeug: 127.0.0.1 - - [18/Sep/2018 07:12:58] "GET /web?debug= HTTP/1.1" 500 -
2018-09-18 07:12:58,626 27465 ERROR ? werkzeug: Error on request:
Traceback (most recent call last):
  File "/opt/odoo/odoo11-venv/lib/python3.4/site-packages/werkzeug/serving.py", line 205, in run_wsgi
    execute(self.server.app)
  File "/opt/odoo/odoo11-venv/lib/python3.4/site-packages/werkzeug/serving.py", line 193, in execute
    application_iter = app(environ, start_response)
  File "/opt/odoo/odoo11/odoo/service/server.py", line 250, in app
    return self.app(e, s)
  File "/opt/odoo/odoo11/odoo/service/wsgi_server.py", line 164, in application
    return werkzeug.contrib.fixers.ProxyFix(application_unproxied)(environ, start_response)
  File "/opt/odoo/odoo11-venv/lib/python3.4/site-packages/werkzeug/contrib/fixers.py", line 152, in __call__
    return self.app(environ, start_response)
  File "/opt/odoo/odoo11/odoo/service/wsgi_server.py", line 154, in application_unproxied
    result = handler(environ, start_response)
  File "/opt/odoo/odoo11/odoo/http.py", line 1317, in __call__
    self.load_addons()
  File "/opt/odoo/odoo11/odoo/http.py", line 1349, in load_addons
    m = __import__('odoo.addons.' + module)
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
    
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
    
[...]

PR #474, opened 2018-10-19
account_financial_report: Switch to (Python 3.4 & below)-compatible syntax

MaxyMoos:
This commit makes the module compatible with Python 3.4 & below, fixing #462.
Even if Odoo 11 is supposed to be run using Python 3.5+, compatibility with previous Python 3 versions is good to have.

```dict3 = {**dict1, **dict2}```
is syntactic sugar available in Python 3.5+ to merge dictionaries (see [PEP 448](https://www.python.org/dev/peps/pep-0448/)) and functionally equivalent to:
```
dict3 = {}
dict3.update(dict1)
dict3.update(dict2)
```

### DIFF
--- a/account_financial_report/report/general_ledger_xlsx.py
+++ b/account_financial_report/report/general_ledger_xlsx.py
@@ -63,7 +63,7 @@ class GeneralLedgerXslx(models.AbstractModel):
                      'type': 'amount_currency',
                      'width': 14},
             }
-            res = {**res, **foreign_currency}
+            res.update(foreign_currency)
         return res
 
     def _get_report_filters(self, report):
--- a/account_financial_report/report/open_items_xlsx.py
+++ b/account_financial_report/report/open_items_xlsx.py
@@ -49,7 +49,7 @@ class OpenItemsXslx(models.AbstractModel):
                      'type': 'amount_currency',
                      'width': 14},
             }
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Odoo 11 simply don't work on Python 3.4, so it's to waste time trying to put code compatible with it. Why do you want this change?

MaxyMoos:
@pedrobaeza I've been running Odoo 11 with Python 3.4.2 for quite some time now, no problem so far.

I've proposed this change because it's a very simple one, that i've seen another user than me affected by it, and because Odoo has done the same thing [walking back this syntax](https://github.com/odoo/odoo/pull/27514) in their codebase.

If that's not in line with the OCA philosophy and we'd like to stick with 3.5+ syntax, fine by me, ultimately I'll have to switch on to Python 3.5+ anyway.

pedrobaeza:
Well, it's a question of being inline with Odoo. Although they don't try to break compatibility on purpose, the fact is that official Odoo requirement is Python 3.5, so you should stick to that, and as said, wasting time fixing all possible idiomatics that are only Python 3.5 in all OCA modules is worst than switching your deployment to 3.5.

MaxyMoos:
@pedrobaeza Fair enough. I'll close this PR, #462 can be closed too.

PR #475, opened 2018-10-23
[10.0] Fix import report_xlsx_helper without try-except

Tonow-c2c:
same things as [https://github.com/OCA/account-financial-tools/pull/716](https://github.com/OCA/account-financial-tools/pull/716)


jcoux (approved):
Code review only.

PR #476, opened 2018-10-26
[8.0] [FIX] account_financial_report_webkit: overlapping headers

Jerther:
When spanning multiple pages, those reports split lines and cause overlapping with the header on the next page.

**Before**
image

**After**
image


rafaelbn (approved):

agb80:
Tested and worked pretty well

PR #477, opened 2018-11-02
12.0 mig customer activity statement

gdgellatly:
Functionally this module is basically good to go, I just need to do a lint, remove some comments, change some variable names / stop their reuse and general review. Plus complete README.  But thats for next week.

This module refactors the existing customer_statement_* modules and introduces significant changes.

What hasn't changed:
- the sql queries (well except they are now mostly properly escaped and parameterized but the content is identical)
- function names mostly - although a couple got a new parameter or changed return type.

Changes:
- Technically the entire data structures have been made more sane (to me).  Rather than multiple dicts by section/partner/currency_browse_record, there is now a single dict of data of partner/currency_id/section, and a master dict of all currencies sent to the report for monetary display.
- Shared parent abstractmodels for both the wizard and report removing duplication across models.
- Settings screen to enable either or both wizards and to set defaults.
- Performance optimization by caching date and currency calls.
- Performance optimization, no more multiple looping over partners.
- Additional option to age by months - dates default to prior month.
- Additional option to filter by negative balances.
- Change default of days aging to 30 days prior on Activity Statement, start of year is not right.
- Conversion of strings to date types.
- Conversion of tests.
- Formatting of reports more in line with Invoices, removing final remnants of dedicated styling (a little, could do better, bit more needed here).
- Conversion of less to scss.
- Make wizard view more Odooish with multiple groups of 2.
- Defaults preparation at report entry to enable either email attachment or printing without wizard using defaults.

\\cc @jbeficent


JordiBForgeFlow:
@gdgellatly This is great! Can this be backported to 11.0 with relative ease? I guess that the basic difference is the module rename, right? I would very much prefer to make sure that all version share to the maximum extent the same code base, except for key differences that we cannot backport.

gdgellatly:
@jbeficent Thanks.

Well there is quite a bit of difference but nothing that can't be backported.  I merged 3 v11 modules then added a config screen and addressed roadmap items. It wasn't so much a rewrite, just a general cleanup and refactor but it was a lot of change - took a few days.

The first issue is new fields are added (and maybe removed / renamed) which is a breaking change, but I guess we have version numbering for that and its only wizards and reports so no data issues.

The only v12 specific features used were.
- scss (but that in itself is an identical port of the less file in v11 with a different grey used for litigation)
- native date types

To backport *this* module and run it alongside the existing two modules (or deprecate them) is not much work. An hour maybe once I've finished this up, to replace the scss file and make sure dates are strings again.

Given that we don't touch any existing model or store any data, just wizards and reports, I think it is reasonable to do this, as an uninstall of the old and install of the new doesn't risk data loss but it might leave old tables.  Then we can just update readme's/maturity/maybe logs effectively telling people to use the newer module.

To make it all play nicely and require no sysadmin effort for existing installs on an upgrade of existing module and to get all the improvements, I think would be quite a lot more work.  You effectively need to split this into 5 separate modules to make it all non breaking. And then you aren't really keeping the same code base.

partner_statement/static/description/index.html
```diff
@@ -0,0 +1,78 @@
+<section class="oe_container">
+    <div class="oe_row oe_spaced">
+        <div class="oe_span12">
+            <h2 class="oe_slogan">Partner Activity Statement</h2>
```
MiquelRForgeFlow: remove this `index.html` and the `Activity_Statement.png` files

levkar:
Great job! I have a few questions and some bug reports. 

Do we really need groups and configuration for enabling these statements? What is the use case for this? Wouldn't it be better if this group is simply added to billing group by default? Other options can still be on the configuration page. 

1) I am getting this error if i select "Show Aging Buckets" and "Age by days"

>   File "/partner_statement/report/report_statement_common.py", line 199, in _get_bucket_dates_days
>     'minus_30': date_end - timedelta(days=30),
> TypeError: unsupported operand type(s) for -: 'str' and 'datetime.timedelta'
> 

2) And this error if i select "Show Aging Buckets" and "Age by months"
 
>  File "/home/levent/code/levkar/v12/external-test/partner_statement/report/report_statement_common.py", line 216, in _get_bucket_dates_months
>     d = d.replace(day=1) - timedelta(days=1)
> TypeError: replace() takes no keyword arguments

3) Message for empty activity statement should be "The partner doesn't have any activity." instead of "The partner doesn't have due entries."

4) Start date is 1 month before today (which should be beginning of previous month according to the code). 

Thanks...

levkar:
5) Ending balance is incorrect (shows balance of the last invoice as far as i understand)

gdgellatly:
@levkar - thanks for the review

Yes the groups are necessary both for current and planned future behaviour, and is the well established method of hiding or showing content to users.  Otherwise if a company only wants Outstanding Statements, how do they hide the activity statement?

1 and 2. That is really weird, I tested it maybe 100 times. All I can think is now the ORM is passing strings instead of dates to reports in some very recent change. To safeguard, and aid v11 backwards compatibility I'll just guard for either type in the opposite way I do for datestrings.
3. I disagree - because even if a partner has no activity, if they have anything due this message won't show.  It is only if they have no due entries at all it will show, not if they have no activity.
4. Cannot reproduce, monthly is start of prior month, by days is 1 month ago as expected.
5. Cannot reproduce, ending balance is showing the accumulated balance as expected - which in general is the same as the accumulated balance for the previous line.

Thats even weirder, the tests still pass without calling the guard statements. There must be something in the web interface which means self._fields is different depending on interface

levkar:
@gdgellatly you're welcome.

I think installing is the intention for having it enabled by default. I wouldn't prefer to do another operation to get it working.  Why would one want to have outsanding but hide the activity statement? For the debt collectors?

3 -> ok
4 -> it happens when "age by days" is selected in config page. i think it is the intended way to use it. not a bug.
5 -> yes it is always the accumulated balance of previous line (is it?) 

Can't wait for the acceptance of this PR.
Thanks again.

gdgellatly:
Because typically companies send 1 type of statement only and will only
want one.  It is a periodic customer facing document. If you enable both by
default users can choose wrong one. And if a customer has received one type
for years, they get confused.

I originally wrote statement module in 2010 and send 20,000 or more a
month. There are 9 people responsible for this of varying competency. If I
give them choice, guarantee 1 gets it wrong. Same reason I put defaults in
config, and why I nearly put security groups on the wizard fields so only
finance manager could change.

It is entirely about usability and reducing complexity for users, not
admins.

But it is precisely zero extra work for the admin who wants both compared
to past. In past they needed 2 actions, install 2 modules. Now they need 2
actions, install 1 module, and configure 1 time.

But the other reasons which are for the future, and why I put the defaults
in there is for automated statement runs via email, individual emails, and
also for adding the reports to users who are not part of billing group,
I.e. let sales print an activity statement but only with the defaults.

levkar:
Thanks for the explanation and the patience.  I am enlightened and know why you did it that way. 

Activity statement is equivalent to partner ledger whereas outstanding statement is equivalent to followup letter. Sending them with e-mail would be the next step (as you have sorted out the defaults).

seancallahan88:
Been testing this on a demo database;

Found this error while trying to open the partner_statement.activity_statement with the report designer in odoo Studio:

> `Odoo Server Error
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/addons/web_studio/controllers/report.py", line 268, in _test_report
>     'studio': True,
>   File "/usr/lib/python3/dist-packages/odoo/addons/web_studio/models/ir_actions_report.py", line 15, in render_qweb_html
>     return super(IrActionsReport, self).render_qweb_html(docids, data)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_actions_report.py", line 710, in render_qweb_html
>     data = self._get_rendering_context(docids, data)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_actions_report.py", line 723, in _get_rendering_context
>     data.update(report_model._get_report_values(docids, data=data))
>   File "/var/lib/odoo/custom_addons/partner_statement/report/activity_statement.py", line 125, in _get_report_values
>     return super()._get_report_values(docids, data)
>   File "/var/lib/odoo/custom_addons/partner_statement/report/report_statement_common.py", line 320, in _get_report_values
>     company_id = data['company_id']
> KeyError: 'company_id'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 654, in _handle_exception
>     return super(JsonRequest, self)._handle_exception(exception)
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 312, in _handle_exception
>     raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
>   File "/usr/lib/python3/dist-packages/odoo/tools/pycompat.py", line 87, in reraise
>     raise value
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 696, in dispatch
>     result = self._call_function(**self.params)
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 344, in _call_function
>     return checked_call(self.db, *args, **kwargs)
>   File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 97, in wrapper
>     return f(dbname, *args, **kwargs)
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 337, in checked_call
>     result = self.endpoint(*a, **kw)
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 939, in __call__
>     return self.method(*args, **kw)
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 517, in response_wrap
>     response = f(*args, **kw)
>   File "/usr/lib/python3/dist-packages/odoo/addons/web_studio/controllers/report.py", line 157, in get_report_views
>     report_html = self._test_report(report_name, record_id)
>   File "/usr/lib/python3/dist-packages/odoo/addons/web_studio/controllers/report.py", line 275, in _test_report
>     "message": err.message,
> AttributeError: 'KeyError' object has no attribute 'message'`

HviorForgeFlow:
This is an error related with  EE code on web_studio module. Exceptions errors has no longer message attribute.

gdgellatly:
@seancallahan88 I doubt any wizard based reports with their own parsers will work with web studio because they are precisely designed to work as a pair. In order to make the report print standalone, we must call the wizard. When called, the wizard always returns a company_id.

Studio under the hood is a pretty basic tool. Only so much you can do. Even if the report did load, no chance of doing anything very useful. 

That said, maybe there is a case to make company_id required in the wizard, but it won't fix your problem.

MiquelRForgeFlow:
@gdgellatly Could you cherry-pick https://github.com/OCA/account-financial-reporting/pull/516 and https://github.com/OCA/account-financial-reporting/pull/517? :)

gdgellatly:
@mreficent Because of the version bumping requirement 'cherry-pick' has become a pseudonym for cut and paste. Besides, only 2 of the fixes in #517 are required.

Anyway it's done now.

MiquelRForgeFlow (approved):
LGTM

partner_statement/views/activity_statement.xml
```diff
@@ -0,0 +1,136 @@
...
+                    </div>
+                </div>
+
+                    <t t-if="d['currencies']">
+                        <br/>
+                        <t t-foreach="d['currencies'].items()" t-as="currency">
+                            <t t-set="display_currency" t-value="Currencies[currency[0]]" />
+                            <t t-set="currency" t-value="currency[1]" />
+                            <p>
+                                <span t-esc="'' if  account_type == 'receivable' else 'Supplier '"/>Statement between <span t-esc="d['start']" /> and <span t-esc="d['end']" /> in <span t-esc="display_currency.name"/>
```
levkar: Supplier is not translatable

gdgellatly: Oh well, I'm done with this PR. Supplier statements isn't really a thing anyway. I only put it in for migrations, as far as I'm concerned they shouldn't be in there. The issue is in 11.0 just worse, so once this is merged, if someone fixes it we can forward it.

gdgellatly:
NOTE: I found 2 errors in practice.

1: VAT was accessed incorrectly
2: Statement was addressed to partner address rather than partner invoice address.

Latest commit resolves.

seancallahan88:
Received this when trying to export a Outstanding statement.
Activity statement worked fine.

> Error:
> Odoo Server Error
> 
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 346, in _compiled_fn
>     return compiled(self, append, new, options, log)
>   File "<template>", line 1, in template_partner_statement_outstanding_statement_document_157
>   File "<template>", line 2, in body_call_content_156
>   File "<template>", line 4, in set_132
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb.py", line 405, in _get_widget
>     content = converter.value_to_html(value, field_options)
>   File "/usr/lib/python3/dist-packages/odoo/addons/snailmail/models/ir_qweb_fields.py", line 13, in value_to_html
>     return super(Contact, self).value_to_html(value, options)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb_fields.py", line 640, in value_to_html
>     if not value.exists():
> AttributeError: 'int' object has no attribute 'exists'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1675, in report_download
>     response = self.report_routes(reportname, converter=converter, **dict(data))
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 517, in response_wrap
>     response = f(*args, **kw)
>   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1612, in report_routes
>     pdf = report.with_context(context).render_qweb_pdf(docids, data=data)[0]
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_actions_report.py", line 682, in render_qweb_pdf
>     html = self.with_context(context).render_qweb_html(res_ids, data=data)[0]
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_actions_report.py", line 722, in render_qweb_html
>     return self.render_template(self.report_name, data), 'html'
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_actions_report.py", line 505, in render_template
>     return view_obj.render_template(template, values)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_ui_view.py", line 1284, in render_template
>     return self.browse(self.get_view_id(template)).render(values, engine)
>   File "/usr/lib/python3/dist-packages/odoo/addons/website/models/ir_ui_view.py", line 318, in render
>     return super(View, self).render(values, engine=engine, minimal_qcontext=minimal_qcontext)
>   File "/usr/lib/python3/dist-packages/odoo/addons/web_editor/models/ir_ui_view.py", line 27, in render
>     return super(IrUiView, self).render(values=values, engine=engine, minimal_qcontext=minimal_qcontext)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_ui_view.py", line 1293, in render
>     return self.env[engine].render(self.id, qcontext)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb.py", line 59, in render
>     result = super(IrQWeb, self).render(id_or_xml_id, values=values, **context)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 274, in render
>     self.compile(template, options)(self, body.append, values or {})
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 348, in _compiled_fn
>     raise e
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 346, in _compiled_fn
>     return compiled(self, append, new, options, log)
>   File "<template>", line 1, in template_4928_126
>   File "<template>", line 2, in body_call_content_125
>   File "<template>", line 3, in foreach_124
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 353, in _compiled_fn
>     raise QWebException("Error to render compiling AST", e, path, node and etree.tostring(node[0], encoding='unicode'), name)
> odoo.addons.base.models.qweb.QWebException: 'int' object has no attribute 'exists'
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/qweb.py", line 346, in _compiled_fn
>     return compiled(self, append, new, options, log)
>   File "<template>", line 1, in template_partner_statement_outstanding_statement_document_157
>   File "<template>", line 2, in body_call_content_156
>   File "<template>", line 4, in set_132
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb.py", line 405, in _get_widget
>     content = converter.value_to_html(value, field_options)
>   File "/usr/lib/python3/dist-packages/odoo/addons/snailmail/models/ir_qweb_fields.py", line 13, in value_to_html
>     return super(Contact, self).value_to_html(value, options)
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb_fields.py", line 640, in value_to_html
>     if not value.exists():
> AttributeError: 'int' object has no attribute 'exists'
> 
> Error to render compiling AST
> AttributeError: 'int' object has no attribute 'exists'
> Template: partner_statement.outstanding_statement_document
> Path: /templates/t/t/t[2]/address
> Node: <address t-esc="o.address_get(['invoice'])['invoice']" t-options="{&quot;widget&quot;: >&quot;contact&quot;, &quot;fields&quot;: [&quot;address&quot;, &quot;name&quot;], >&quot;no_marker&quot;: True}"/>

gdgellatly:
@mreficent I had to make some fairly major changes to fix activity statements. Payments were showing pages long, broken down by each reconciled invoice. I did it for minimal change but its pretty hacky.

gdgellatly:
See #546

MiquelRForgeFlow:
Merge https://github.com/odoonz/account-financial-reporting/pull/1 to solve the issue.

gdgellatly:
@mreficent I have deployed and tested. It still has the issue of a payment reconciling another payment after statement date, but I think ultimately that is an Odoo bug in how it does full reconciliations and the overall impact is minor. At least it balances now which I think is best we can do.

MiquelRForgeFlow (approved):
Yeah :+1:

HviorForgeFlow (approved):

partner_statement/README.rst
```diff
@@ -0,0 +1,132 @@
...
+#. Click *Save*
+
+Removing the wizard from menus follows the same process.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Invoicing > Sales > Master Data > Customers or Invoicing > Purchases > Master Data > Vendors and select one or more
```
MiquelRForgeFlow: Could you update this line? I think in v12 this is different...

gdgellatly: Well it is different in Enterprise as well. Just made it generic. You can run this from any list of partners.

MiquelRForgeFlow: haha good

gdgellatly:
@hveficent @mreficent note I just added another improvement/fix for this related to its use as an email attachment.

JordiBForgeFlow:
@gdgellatly when I print the activity or outstanding statement choosing Aging Method = "Age by Months" the report results in a blank page. When I choose "Age by Days" it works fine.

See screenshot:

image

JordiBForgeFlow (approved):
:+1: Tested functionally. Works as expected. Please disregard my last comment.

Thanks @gdgellatly for this great contribution!

PR #467, opened 2018-10-01
[11.0][account_financial_report] fix broken wizard views in afr. 

JordiBForgeFlow:
Forward port of PR #460 

cc @alexis-via

PR #478, opened 2018-11-07
[11.0] [account_financial_report] Improvements to trial balance

JordiBForgeFlow:
This PR includes:
* Fixes in https://github.com/OCA/account-financial-reporting/pull/469 made by @alexis-via 
* Fixes in https://github.com/OCA/account-financial-reporting/pull/470 made by @jesusVMayor 

- Adds 'Period balance' column
- Renames the option 'Hide accounts at 0'. Means no initial, no debit, credit
  or ending balance for the period.
- Fixes logic to remove lines with 0 activity for the period.
- Adds to 'Hierarchy On' the option 'No hierarchy'. Sometimes it is necessary to print a plain list of accounts.
- The line Undistributed Profits/Losses accounts is duplicated, and the second line shows only the application of P&L for the current fiscal year. By doing this we ensure that:
1)The total trial balance initial and final balances are 0, and total debits = total credits, which is something that an accountant wants to make sure happens.
2) Provides clarity on what portion of the Undistributed Profits/Losses belong to the application of current year's P&L.
image

If contributors agree we can merge in 11.0 and I will backport to 10.0.


JordiBForgeFlow:
@alexis-via @jesusVMayor @pedrobaeza Can we move ahead with this one?

JordiBForgeFlow:
I have now improved the Trial Balance with respect to the Undistributed Profits/Losses account:
* I am adding to the Undistributed Profits/Losses account the current period's debits and credits, corresponding to journals made in the period to both this account and also to the P&L.
* I am adding a footer note to explain why the trial balance is not totaling to zero in the ending balances. 

image

JordiBForgeFlow:
As discussed during OCA Days 2018 with @pedrobaeza and @alexis-via I have now added an extra line that represents the portion of P&L application to the Undistributed Profits/Losses accounts. By doing this we ensure that:
1) The total trial balance initial and final balances are 0, and total debits = total credits, which is something that an accountant wants to make sure happens.
2) Provides clarity on what portion of the Undistributed Profits/Losses belong to the application of current year's P&L.

image

JordiBForgeFlow:
I have added tests for the new features.

jgrandguillaume (approved):
Hi,

Functionnal test made, that's a great whole lot of feature. Thanks !

:+1: 

Joël

JordiBForgeFlow:
@jgrandguillaume thanks for the review!

fclementic2c:
Thanks a lot for all these improvements, it is really great.

I am just not happy with one of them: the extra line to make the trial balance totals at zero. It is a non-sense. 

Let me please clarify : A trial balance total is always zero because other software creates a opening entry at the beginning of the new fy.

The opening entry goal is, as you know it, to net to zero all P&L accounts and transfer the cumulated balances to the Current year earning account.

So, in fact the problem is that **we should not display at all the Current year earning account line** on this report since we already have all p&l accounts.

Maybe, it could make sense to make it visible in case you request a TB with balance sheet accounts only...but this is the only case I see.

What is important is that we show the previous fy P&L balances (which are usually netted by accountant after the yearly company board meeting deciding how profit is distributed).

Current year earning on the Profit and Loss report make sense but not on the TB (and same rational for the GL).

JordiBForgeFlow:
@fclementic2c The problem is that the "Current year earning account" is effectively a balance sheet account and DOES have some balance during the year. For example, imagine it's January 10th 2019, and you have not yet allocated the previous year earnings.  Previous year results as a profit of 10.000€.

When you run the Trial Balance you expect to see in the "Current year earning account" the balance from previous year, as -10.000€, because you expect that the system performs an automated rolling of P&L to current year earning.

If you hide the "Current year earning account" the Trial Balance for 2019 will not net to 0.

JordiBForgeFlow:
@fclementic2c in addition you can make manual entries to the current year earning account, for example to distribute profits to shareholders.

So an acccountant expects to see  activity in this account during the year.

JordiBForgeFlow:
@fclementic2c I have now changed the logic.

Essentially one has to think that the intention for the removal of manual fiscal year closing is for P&L entries made during the previous year to roll-over to the current year's 'Current Earnings Account' balance sheet account. 

The purpose of the Trial Balance and the General Ledger is to display the status and movements of a specific GL account during a period. When they represent the account 'Current Year Earnings' they should assume that prior year's P&L have already been allocated to this account, but not current year's because the year has not yet been finalized, and that's why you have entries in P&L.

On the other hand, the Balance Sheet is DESIGNED to work differently. The Balance Sheet works together with the P&L, and rolls over the P&L result into the 'Current Year Earnings' account, for the current year.

Even in Odoo 8 and before, where fiscal year closing entries existed, the balance sheet needed to reflect the current year earnings as the sum of current year P&L, dynamically, in order to be balanced. But the Trial Balance did not require to reflect the application of current year P&L, because that only happened at year end.

So, differeces in the treatment 'Current Year Earnings' account are EXPECTED to occur between Balance Sheet and Trial Balance. And the accountants will quickly understand that the difference is the current year P&L.

The current code reflects the statements described above. And I think that this is correct now.

rafaelbn (approved):
Thanks Jordi! Great Work! :+1:

jesusVMayor (approved):

JordiBForgeFlow:
@fclementic2c what do you think about this last approach?

rafaelbn:
@jbeficent what about https://github.com/OCA/account-financial-reporting/pull/467 is it included in this PR?

fclementic2c:
@jbeficent I agree with your explanations but I would like to test further. I should managed to do this on friday. Can you wait a bit for me please ? 
Just a question : you stick to the 2 CYE account lines or did you change anything ?

JordiBForgeFlow:
@fclementic2c just one line. And this current year earnings line shows:

As initial balance the cumulated balance of previous year, assuming automated allocation of prior years PL balances.
As initial debit, the initial balance, if it was debit side
As initial credit, the initial balance, if it is credit side
As debit, the sum of debits of journal items made explicitly to this account, for the period selected in the Trial Balance.
As credit, same as above
As period balance, debit - credit
As final balance, initial balance + period balance
As final debit, initial debit + period debit
As final credit, initial credit + period credit

Therefore the current year PL is not going to be allocated to the Unaffected Current Year Earnings in the Trial Balance.

fclementic2c:
You won @jbeficent , I tested it :)
Great job, it is all good now :)
:+1: 


[20181114 - Test new trial balance OCA.xlsx](https://github.com/OCA/account-financial-reporting/files/2579900/20181114.-.Test.new.trial.balance.OCA.xlsx)

fclementic2c (approved):

JordiBForgeFlow:
Thanks to all for your reviews!! :)

pedrobaeza:
@jbeficent this will be on v10 as well?

JordiBForgeFlow:
@pedrobaeza yes. Will work on that.

pedrobaeza:
Great, thanks for doing this nasty work. We need to convert this to something maintenable on v12.

PR #479, opened 2018-11-13
[11.0][MIG] account_financial_report_date_range

MiquelRForgeFlow:
Standard migration.


pedrobaeza:
Are you still using these ugly reports?

MiquelRForgeFlow:
ping: @lreficent

LoisRForgeFlow:
@pedrobaeza haha not really, but it is auto-install with account and date_range in v10 so odoo logs the warning of module missing in v11... But thinking it twice, you are right: You can close it and I will just uninstall after migration or We can just fast-forward it for completeness (as it is already migrated). Whatever you prefer :wink:

pedrobaeza:
Yeah, as Odoo has removed all reports in v12, it's not interesting to preserve this IMO.

MiquelRForgeFlow:
This is v11. I mean, I like completeness!

pedrobaeza:
It's not a question of completeness, but as said, that reports are not used, so you don't need that extra feature. If you have it in v10, it will probably not being used already, so you can uninstall there, but if so, just teach your users to not use it anymore and uninstall on v11.

MiquelRForgeFlow:
Who knows, maybe really exist somebody who use it. But I made a work and I don't like to see my work wasted. Efforts should be compensated.

LoisRForgeFlow (approved):
My 2 cents.

pedrobaeza (approved):
OK, let's use it...

PR #482, opened 2018-11-16
account_financial_report: set backgroud-color to table

leemannd:
If not defined, in odoo enterprise, the table in accounting financial reports will be the same as the base backgroud-color of odoo. Which makes it unreadable.

This is a forward-port from https://github.com/OCA/account-financial-reporting/pull/473 . It hasn't been tested, but has been asked by @jbeficent

### DIFF
--- a/account_financial_report/static/src/css/report.css
+++ b/account_financial_report/static/src/css/report.css
@@ -1,5 +1,6 @@
 .act_as_table {
     display: table !important;
+    background-color: white;
 }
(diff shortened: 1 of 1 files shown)


leemannd:
@fclementic2c forward-port of the fix done in V10

fclementic2c:
fast-tracking this beacuse without this the module is impossible to use on with Odoo enterprise.

PR #483, opened 2018-11-16
[10.0][FIX] account_financial_report_qweb: Widget date for t-raw date

Abranes:
Fix issues with date format on `t-raw` attributes.

Before:
image

After:
image

### DIFF
--- a/account_financial_report_qweb/report/templates/aged_partner_balance.xml
+++ b/account_financial_report_qweb/report/templates/aged_partner_balance.xml
@@ -219,7 +219,7 @@
                                t-att-data-res-model="'account.move.line'"
                                class="o_account_financial_reports_web_action"
                                style="color: black;">
-                            <t t-raw="line.date"/></a>
+                            <t t-raw="line.date" t-options="{'widget': 'date'}"/></a>
                         </span>
                     </div>
                     <!--## move-->
@@ -279,7 +279,7 @@
                                t-att-data-res-model="'account.move.line'"
                                class="o_account_financial_reports_web_action"
                                style="color: black;">
-                            <t t-raw="line.date_due"/></a>
+                            <t t-raw="line.date_due" t-options="{'widget': 'date'}"/></a>
                         </span>
                     </div>
                     <!--## amount_residual-->
--- a/account_financial_report_qweb/report/templates/aged_partner_balance.xml
+++ b/account_financial_report_qweb/report/templates/aged_partner_balance.xml
@@ -219,7 +219,8 @@
                                t-att-data-res-model="'account.move.line'"
                                class="o_account_financial_reports_web_action"
                                style="color: black;">
-                            <span t-field="line.date"/></a>
+                            <!--## We don't use t-field because it throws an error on click -->
+                            <t t-esc="line.date" t-options="{'widget': 'date'}"/></a>
                         </span>
                     </div>
                     <!--## move-->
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why `t-raw` and not `t-esc`?

Abranes:
I have tried to make the minimum possible changes.
Also, I have not seen any date in the template that is in a `t-esc`.

pedrobaeza:
Yeah, I know, but now that we are revisiting this I wonder why and if working with `t-esc`, it's better to avoid `t-raw` expressions.

Abranes:
I don't know the real purpose of `t-raw` here, but we can change it for `t-esc` and add the `t-options` attribute or change the `t` label for `span` label and `t-raw` for `t-field` and remove the `t-options` attribute.
What do you think could be better?

pedrobaeza:
`t-field` would remove the need of `t-options` I think.

account_financial_report_qweb/report/templates/aged_partner_balance.xml
```diff
@@ -219,7 +219,7 @@
                                t-att-data-res-model="'account.move.line'"
                                class="o_account_financial_reports_web_action"
                                style="color: black;">
-                            <t t-raw="line.date"/></a>
+                            <t t-raw="line.date" t-options="{'widget': 'date'}"/></a>
```
Abranes: So, let's do it this way, ok?
```suggestion
                            <span t-field="line.date"/></a>
```

pedrobaeza: Yes, please!

pedrobaeza:
Is it working correctly with this?

Abranes:
I was testing and with `t-field` and `span` display correctly but when you click over the `date` raise an error but with `t-esc` and `t-options` is working correctly.

The error:
`Traceback (most recent call last):`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 642, in _handle_exception`
`    return super(JsonRequest, self)._handle_exception(exception)`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 684, in dispatch`
`    result = self._call_function(**self.params)`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 334, in _call_function`
`    return checked_call(self.db, *args, **kwargs)`
`  File "/home/odoo/10.0/odoo/odoo/service/model.py", line 101, in wrapper`
`    return f(dbname, *args, **kwargs)`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 327, in checked_call`
`    result = self.endpoint(*a, **kw)`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 942, in __call__`
`    return self.method(*args, **kw)`
`  File "/home/odoo/10.0/odoo/odoo/http.py", line 507, in response_wrap`
`    response = f(*args, **kw)`
`TypeError: call_kw() takes at least 5 arguments (5 given)`

pedrobaeza:
OK, thanks for the test. Maybe that was the reason to put it. Then go back to `t-esc` and it's good to put a comment in the XML file line for avoiding to wonder this thing again next time someone views the code.

Abranes:
@pedrobaeza It's ok?

pedrobaeza:
Yes, but it should be reviewed by other contributor for merging this.

Abranes:
Someone tested?

JordiBForgeFlow (approved):

pedrobaeza:
Cherry-picked for v11 in a436052687c4bf8d57a59816e6056d731fda2578

PR #416, opened 2018-05-04
[ADD][10.0] account_financial_report_qweb: add analytic tags

jcoux:

PR #484, opened 2018-11-22
[11.0][account_financial_report][IMP] Refactoring

JordiBForgeFlow:
Changes included:
All reports:
- Rename field to hide accounts at 0 to 'hide_account_at_0'
Trial Balance:
- Add possibility to filter by hierarchy levels
- XLSX format will show the hierarchy levels in bold

General Ledger:
- Add the possibility to filter by analytic tags
- Fixes an error on the default date
Journal Ledger:
- The filter on Journals is now optional. If the user does not choose
a journal, by default it will display all journals.

Aged Partner Balance:
- Fixes an error on the default date

Hierarchy level filtering
------------------------------

With this change a user can filter what hierarchy level they want to report on.

For example, I want to show only up to level 1:
image

image

Now I want to show all levels:
image

image

cc @rafaelbn @pedrobaeza


pedrobaeza:
You are my idol!

feketemihai:
@jbeficent I tested on runbot and not working as expected, i think,I created 3 groups: Class 1 - > Class 10 and assigned 2 accounts to class 10, Class 2 and assigned 1 account to class 2.
Can you check if you limit to 1 or 2 levels, the result is not as expected, also we should show the accounts that under the hierarchy level, plus some layout issues,since the accounts are not added right after the hierarchy, but at the end. Also there is a layout issue in the wizard view, for Filter Accounts.

feketemihai:
@jbeficent @pedrobaeza Although , nice improvement and all others, sorry not having a lot of time to be more involved.

JordiBForgeFlow:
@feketemihai I display the trial balance UP TO the level that the user chooses.
So, for example, if the user chooses to display up to level 1, it will display account groups in level 0 and level 1, but not level 2 and beyond.

JordiBForgeFlow:
@feketemihai My mistake. You are right. Should only display accounts in the level you choose to show. Fixing now.

JordiBForgeFlow:
@feketemihai Fixed.

JordiBForgeFlow:
@jcoux can you please review this PR? It includes a forward-port to v11 of your PR https://github.com/OCA/account-financial-reporting/pull/416 to add the filters on analytic tags to the general ledger

feketemihai:
@jbeficent probably it was some misunderstanding in what i commented before, i attach how it should look like, probably much easier to check them
[Hierarchy level.pdf](https://github.com/OCA/account-financial-reporting/files/2620325/Hierarchy.level.pdf)

fclementic2c:
I am a bit surprise by this feature since you take the assumption that level are consistent through a chart of account but this is far from being the case.

Ex: A level 2 in the accounts 6... can be equivalent to a level 3 or 4 for the accounts starting by 7...  

In my view you are going a bit to far here (but this is just an opinion, I will not block the PR)

JordiBForgeFlow:
@fclementic2c In Spain (and perhaps in other countries too) the levels are specific. Level 3 has a certain meaning. But since this could not be the case in other countries, this would be an optional feature.

@feketemihai I understand that you want to show all the levels above, right? I was debating with this. But then, are we talking about a Trial Balance? A Trial Balance report is supposed to total to 0 the initial and final balances. With a hierarchy this is impossible.

feketemihai:
@jbeficent I think your wrong, if you take the hierarchy plus accounts on the same level, the total will always be zero, it's just you are not showing the next levels...that's it, it's too much to work on this, and yes, it's probably much better to only show the hierarchy

feketemihai (approved):

rafaelbn:
OMG! @jbeficent great!!!! :astonished:

rafaelbn:
Tested :+1: 

But I discovered a BUG in the module in both v10 an d v11. The journal ledger now is by journals and you cannot extract the journal ledger for all the journals :astonished: Always we could have a journal ledger for all the journals. I did it in v10 and now it is nor possible.

In Spain this is a legal requirement to present the company accounting...

journal ledger for all journals

rafaelbn (approved):
Tested functionally :+1:

account_financial_report/report/general_ledger.py
```diff
@@ -550,7 +615,7 @@ def _inject_account_values(self):
                 tuple(self.filter_cost_center_ids.ids),
             )
         query_inject_account_params += (
-            self.id,
+            self.id or 'NULL',
```
jcoux: Why do you add the `or 'NULL'`?
The report_id must not be null.
We need the `report_id` to distinct each different asking of reports.

jcoux (changes requested):

JordiBForgeFlow:
> Tested 
> 
> But I discovered a BUG in the module in both v10 an d v11. The journal ledger now is by journals and you cannot extract the journal ledger for all the journals  Always we could have a journal ledger for all the journals. I did it in v10 and now it is nor possible.
> 
> In Spain this is a legal requirement to present the company accounting...
> 
> journal ledger for all journals

This is now fixed.

account_financial_report/report/general_ledger.py
JordiBForgeFlow: Fixed here and in https://github.com/OCA/account-financial-reporting/pull/485/commits/f9d47e5b3274bcbdc03f044b345772e4eaea8c50

JordiBForgeFlow:
@feketemihai @fclementic2c I have fixed the hierarchy level filtering according to your requirements. They will also work for Spain.

Can you do a final review?

Please notice that for simplicity the first hierarchy level is 1 now.

PR #485, opened 2018-11-23
[10.0][account_financial_report_qweb] backport latest changes from 11.0

JordiBForgeFlow:
This PR backports from 11.0 several highly requested features:

- Introduce dependency with module account_group from OCA/account-financial-tools

All reports:
- Rename field to hide accounts at 0 to 'hide_account_at_0'
Trial Balance:
- Add hierarchy levels.
- Add possibility to filter by hierarchy levels
- XLSX format will show the hierarchy levels in bold

General Ledger:
- Add the possibility to filter by analytic tags
- Fixes an error on the default date
Journal Ledger:
- The filter on Journals is now optional. If the user does not choose
a journal, by default it will display all journals.

Aged Partner Balance:
- Fixes an error on the default date


JordiBForgeFlow:
@alexis-via @pedrobaeza @rafaelbn you might want to test thie one...

alexis-via:
Thanks for this great work.

I made a test on real-life data, using a script to generate account groups (https://github.com/akretion/odoo-usability/blob/10.0/account_usability/account.py#L256). Here is my feedback:
1) I think the option "Hide accounts at 0" should be enabled by default (it was the case in my PR)
2) When I generate a PDF or XLSX or Web trial balance using the default options, I do have the account groups, but all the amounts for the account groups are "0.00".
3) when the option "Hide accounts at 0" is enabled, I don't have the account groups in the balance any more !
4) Also, the account groups have a special bold fonts in PDF and Web, but not in XLSX, which make it hard to differentiate regular account lines and account group lines.

alexis-via:
My users just discovered a very stupid bug in the financial report, that we could easily fix in this PR: when launching the wizard "Open Items" or "Aged Partner Balance", the default value of the field "Date at" is NOT today's date, but the date when the Odoo server was started !!! It's surprising that nobody reported/fixed that kind of very stupid bug already ! I love those crazy bugs :)

alexis-via:
I prepared a PR for the date_at fix, but I can't select Eficent/account-financial-reporting as "base fork":... I don't know why.
Well, the fix is here: https://github.com/akretion/account-financial-reporting/commit/8dd415489e5922b7780fcb27541b970c937ea6ad

rafaelbn:
Amazing work @jbeficent thank you!

Thank you @alexis-via for testing. Do you think you could fix in this PR? Then Jordi only have to accept to changes. You maybe could use https://twitter.com/github/status/1062401381082968064

JordiBForgeFlow:
@alexis-via regarding your points:

1) It is now set as default
2) 
Please check that the account groups are configured correctly. When you set 'hierarchy_on' = 'computed' it is going to assume that the code of the account group represents the starting digit of the accounts. For example account group code "1" means that it should take into account all accounts starting with "1". If you are not using this rule, but have a correct parent-child relationship in account groups, please choose the 'hierarchy_on' = 'relation'. Perhaps the option should be set at company level, not to be set by the user.

3) Please show me on a runbot. I cannot reproduce.

JordiBForgeFlow:
@alexis-via Now the hierarchies in the XLSX format appear in bold. Applied this same change also in https://github.com/OCA/account-financial-reporting/pull/484

image

JordiBForgeFlow:
> hy.
> Well, the fix is here: [akretion@8dd4154](https://github.com/akretion/account-financial-reporting/commit/8dd415489e5922b7780fcb27541b970c937ea6ad)

This is now fixed @alexis-via

JordiBForgeFlow:
@alexis-via @rafaelbn All your comments have been attended. This is ready for a final review.

alexis-via:
Great ! I'll do more tests today.

account_financial_report_qweb/wizard/trial_balance_wizard.py
```diff
@@ -131,10 +156,10 @@ def onchange_show_partner_details(self):
         """Handle partners change."""
         if self.show_partner_details:
             self.receivable_accounts_only = self.payable_accounts_only = True
-            self.hide_account_balance_at_0 = True
+            self.hide_account_at_0 = True
```
alexis-via: Now that hide_account_at_0  is True by default, it should be completely removed from this onchange method onchange_show_partner_details() (2 lines of code to remove). Otherwise, the default value at True is immediately removed by this onchange !

alexis-via:
OK, subtotal for account groups works fine for when I select "Hierarchy on" = "Child Accounts".

I don't understand the goal of 'Hierarchy On' = 'Computed'. It is just for odoo DB with "broken" account group configuration i.e. account groups that were not properly configured with a hierarchy ? If yes, then I think users should just reconfigure their account groups to have a good hierarchy...

rafaelbn (approved):
Functionally tested :+1:

alexis-via:
In the introduction of this PR, there is a very important change that is not mentioned: the new behavior of the account with type = "Current Year Earnings". This account is now properly handled in the trial balance, which was not the case before !

alexis-via:
When I configure the trial balance with:
- HIerarchy on = Child Accounts
- Limit hierarchy level = true
- Hierarchy levels to display = 1
Then the report contains the account groups with level = 1 (totals seem OK), but the "regular" accounts are not displayed in the report! I thought I would get the usual report without the lines/totals for account groups with level > 1. For me, I think it's a bug...

alexis-via:
@jbeficent I ported to this branch useful improvements of PR #469 and I fixed the problem with onchange in trial balance for the option accounts at 0.
I still can't select eficent/account-financial-report to make a PR to this branch... there must be a problem with your fork, it's very strange! I can select plenty of other forks, but not yours !
So please grab my commit https://github.com/akretion/account-financial-reporting/commit/15c902a60c2ea7d417654915bdfe0d3b83e33b0b and put it here

JordiBForgeFlow:
Now merging, as all comments have been addressed and we have all the approvals.

pedrobaeza:
Are the latest changes of this PR also in v11?

JordiBForgeFlow:
@pedrobaeza Yes.

pedrobaeza:
OK :)

pedrobaeza:
I'm seeing that this PR includes a new dependency on web_widget_many2many_tags_multi_selection, which will break some installations if they don't have it installed, so we should make it optional, not mandatory.

JordiBForgeFlow:
@pedrobaeza Same as account_group. In fact adding 'web_widget_many2many_tags_multi_selection' makes both reports very close from a view definition perspective.

pedrobaeza:
OK, let's see if someone complains...

alexis-via:
They can't complain with so much benefits introduced by this new version of account_financial_report_qweb !!! :)

PR #489, opened 2018-11-29
11.0 imp account financial report multicompany 2

JordiBForgeFlow:
Supersedes https://github.com/OCA/account-financial-reporting/pull/401

### DIFF
--- a/account_financial_report/report/open_items_xlsx.py
+++ b/account_financial_report/report/open_items_xlsx.py
@@ -9,8 +9,10 @@ class OpenItemsXslx(models.AbstractModel):
     _name = 'report.a_f_r.report_open_items_xlsx'
     _inherit = 'report.account_financial_report.abstract_report_xlsx'
 
-    def _get_report_name(self):
-        return _('Open Items')
+    def _get_report_name(self, objects):
+        report = objects
+        return _('Open Items - %s - %s') % (
+            report.company_id.name, report.company_id.currency_id.name)
 
     def _get_report_columns(self, report):
         res = {
--- a/account_financial_report/README.rst
+++ b/account_financial_report/README.rst
@@ -97,6 +97,7 @@ Contributors
 * Akim Juillerat <akim.juillerat@camptocamp.com>
 * Alexis de Lattre <alexis@via.ecp.fr>
 * Mihai Fekete <feketemihai@gmail.com>
+* Miquel Raïch <miquel.raich@eficent.com>
 
 Much of the work in this module was done at a sprint in Sorrento, Italy in
 April 2016.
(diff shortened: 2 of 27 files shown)


JordiBForgeFlow:
@rafaelbn @alexis-via @pedrobaeza Can you please review this small PR and https://github.com/OCA/account-financial-reporting/pull/490? This will make my last contribution for the day in the AFR :)

JordiBForgeFlow:
@alexis-via happy new year!! 

This PR now adds the titles correctly in all the account financial reports.

Can you please review?
image

account_financial_report/report/open_items_xlsx.py
```diff
@@ -9,8 +9,10 @@ class OpenItemsXslx(models.AbstractModel):
     _name = 'report.a_f_r.report_open_items_xlsx'
     _inherit = 'report.account_financial_report.abstract_report_xlsx'
 
-    def _get_report_name(self):
-        return _('Open Items')
+    def _get_report_name(self, objects):
+        report = objects
+        return _('Trial Balance - %s - %s' % (
```
alexis-via: Wrong string : Trial Balance -> Open Items
Other problem: It seems to me that the "_" method should be closed before string replacement i.e. _('Open Items - %Ss- %s) % (report.company_id.name, ...). If you agree on this second point, it should be changed in the _get_report_name() methods of all the _xlsx.py files.

alexis-via: @jbeficent Do you agree with my 2 remarks ?

JordiBForgeFlow: @alexis-via I have applied the fixes according to your remarks. Once Travis is Green can you give your approval? Once this is merged I will backport to 10.0

alexis-via (approved):
Great work !

AaronHForgeFlow (approved):
Funtional test on runbot. Run reports for two different companies and data seem consistent.
:+1:

pedrobaeza:
The changes you have made forces to put a company, which virtually forbids to generate multi-company consolidated information. You must go back with this and let the possibility to issue them.

I don't understand also why do you need to modify the signature of the `_get_report_name` method for including a parameter that is not used at all (`objects`).

JordiBForgeFlow:
@pedrobaeza I am fixing the issue with multicompany. 

as for the `_get_report_name`, it is used. See https://github.com/OCA/account-financial-reporting/blob/11.0/account_financial_report/report/aged_partner_balance_xlsx.py#L14

pedrobaeza:
But why not using `self`?

JordiBForgeFlow:
@pedrobaeza because `self` is `report.a_f_r.report_aged_partner_balance_xlsx` whereas object is model `report_aged_partner_balance`, the model that has the data.

pedrobaeza:
OK, I see.

PR #493, opened 2018-12-22
Improved dependency - account_financial_report

anandkansagra:
account_financial_report
    After installation of `account_invoicing`, `account` will be installed automatically so no need to put in depends


pedrobaeza (approved):
This is not very critical, as it works as is, but it doesn't hurt either.

PR #497, opened 2019-01-08
[10.0][ADD] account_financial_report_web: add index on account_move_line to speed up report rendering

Cedric-Pigeon:
ON a big database, requesting the Trial balance with details for all partners nerver ends, same for Aged Balance.

By index a combined index on account_move_line to speed up joins, the results are much better.

On my database with more than 2.000.000 records on move lines the time fails from 9 minutes to few seconds....

### DIFF
--- a/account_financial_report_qweb/models/account_move_line.py
+++ b/account_financial_report_qweb/models/account_move_line.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).-
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = 'account.move.line'
+
+    @api.model_cr
+    def init(self):
+        """
+            The join between accounts_partners subquery and account_move_line
+            can be heavy to compute on big databases.
+            Join sample:
+                JOIN
+                    account_move_line ml
+                        ON ap.account_id = ml.account_id
+                        AND ml.date < '2018-12-30'
+                        AND ap.partner_id = ml.partner_id
+                        AND ap.include_initial_balance = TRUE
+            By adding the following index, performances are strongly increased.
+        :return:
+        """
+        self._cr.execute('SELECT indexname FROM pg_indexes WHERE indexname = '
+                         '%s',
+                         ('account_move_line_account_id_partner_id_index',))
+        if not self._cr.fetchone():
+            self._cr.execute("""
+            CREATE INDEX account_move_line_account_id_partner_id_index
+            ON account_move_line (account_id, partner_id)""")
--- a/account_financial_report_qweb/__manifest__.py
+++ b/account_financial_report_qweb/__manifest__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '10.0.2.0.0',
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Do you know what was the original intention of that code? Any possible side effect?

JordiBForgeFlow:
@Cedric-Pigeon You cannot remove this code, because it would lead to the report missing to display the journal entries associated to receivable or payable accounts that have been allocated to no partner.

For example, I created this journal entry:
image

That shows in the general ledger report as:

image

Cedric-Pigeon:
@jbeficent If you request a report per partner, is it a problem to ignore unallocated entries. My problem is that I have a big amount of them and the reports never ends. The queries generated runs for hours.

pedrobaeza:
Maybe put it as an option? Any way, why is this so slow?

Cedric-Pigeon:
@pedrobaeza That's what I am trying to understand but these queries are a nightmare to understand and debug...

pedrobaeza:
I know... I hate the way this was implemented.

Cedric-Pigeon:
@jbeficent @pedrobaeza I finally succeed to have a cleaner solution.

pedrobaeza (approved):
Does this applies to 11.0 as well?

account_financial_report_qweb/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).-
+from odoo import api, models
+
+
+class AccountAccount(models.Model):
+    _inherit = 'account.move.line'
+
+    @api.model_cr
+    def init(self):
```
pedrobaeza, line 11: I think you should indicate in a docstring why we are adding this for our future we to know what we are doing.

Cedric-Pigeon: Done

rousseldenis (approved):

Cedric-Pigeon:
@pedrobaeza I think this applies to 11.0 as well as the queries are still the same

JordiBForgeFlow (approved):

account_financial_report_qweb/models/account_move_line.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).-
+from odoo import api, models
+
+
+class AccountAccount(models.Model):
```
sbidoul: AccountMoveLine?

Cedric-Pigeon: Fixed !

JordiBForgeFlow:
@Cedric-Pigeon When we merge I will forward-port to 11.0

account_financial_report_qweb/models/account_move_line.py
```diff
@@ -0,0 +1,31 @@
...
+                        AND ap.include_initial_balance = TRUE
+            By adding the following index, performances are strongly increased.
+        :return:
+        """
+        self._cr.execute('SELECT indexname FROM pg_indexes WHERE indexname = '
+                         '%s',
+                         ('account_move_line_account_id_partner_id_index',))
+        if not self._cr.fetchone():
+            self._cr.execute("""
+            CREATE INDEX account_move_line_account_id_partner_id_index
```
pedrobaeza: Trailing whitespace

Cedric-Pigeon: Fixed

PR #501, opened 2019-01-10
[10.0][IMP] account_financial_report_qweb: color scheme

i-vyshnevska:
Change backgroud to blank on financial reports


JordiBForgeFlow (approved):

JordiBForgeFlow:
In Odoo CE, when I check without your changes:

image


And with your PR:

image

Seems better not to apply the default white color, right?

pedrobaeza:
:-1: I prefer previous one.

JordiBForgeFlow:
I agree. @i-vyshnevska any opinion?

fclementic2c:
@jbeficent @pedrobaeza 
Check it on the enterprise version too:
screenshot-beta-erp bsonetwork net-2019 01 10-13-31-32

Can't we have a transparent background so it fit both community & enterprise versions ?

pedrobaeza:
@fclementic2c it's not possible to check it in the enterprise version, and this doesn't target enterprise users any way. For that enterprise lovers, it's supposed that enterprise account reports are "better" than this...

fclementic2c:
@pedrobaeza you're too fast. see my updated comment.

...and no, OCA financial reports are far better than enterprise reports. Yesterday again I had the confirmation from my bigger customer at the moment in Paris. There is no way the enteprise report can be acceptable for a bigger company.

pedrobaeza:
I see no bigger improvement putting the initial blank. Any way, that kind of change should affect only enterprise, no community. And if not, switch also to OCA responsive. In v12, we have made a fine work and I consider now it's also better than enterprise one: https://github.com/OCA/web/pull/1066

fclementic2c:
@i-vyshnevska Can we do the background change only if we install the module on a enterprise version ?

i-vyshnevska:
@fclementic2c need to check

i-vyshnevska:
@jbeficent  it's better now
screenshot

fclementic2c (approved):

account_financial_report_qweb/static/src/css/report.css
```diff
@@ -96,11 +97,12 @@
...
 .o_account_financial_reports_page {
-    background-color: @odoo-view-background-color;
+    background-color: white;
     color: @odoo-main-text-color;
     padding-top: 10px;
     width: 90%;
     margin-right: auto;
     margin-left: auto;
     font-family: Helvetica, Arial;
+    padding: 40px
```
pedrobaeza: Why changing padding?

account_financial_report_qweb/static/src/css/report.css
```diff
@@ -96,11 +97,12 @@
 }
 
 .o_account_financial_reports_page {
-    background-color: @odoo-view-background-color;
+    background-color: white;
```
pedrobaeza: Isn't a variable for enterprise background? I still prefer to keep "transparent" background

account_financial_report_qweb/static/src/css/report.css
```diff
@@ -65,6 +65,7 @@
 .act_as_cell.amount {
     word-wrap:normal;
     text-align:right;
+    padding: 2px;
```
pedrobaeza: Why this extra padding? This makes bigger full table.

pedrobaeza (changes requested):

account_financial_report_qweb/static/src/css/report.css
i-vyshnevska: so this looks better https://github.com/OCA/account-financial-reporting/pull/501#issuecomment-453060324

pedrobaeza: Background color has nothing to do with padding.

account_financial_report_qweb/static/src/css/report.css
simahawk: this is broken anyway, is not LESS ;)

account_financial_report_qweb/static/src/css/report.css
simahawk: @i-vyshnevska pls post the difference w/ and w/out.
The issue is: it's really ugly w/out padding as you get the dark gray BG in touch w/ the borders and very next to the text.

account_financial_report_qweb/static/src/css/report.css
simahawk: we should move the whole file to LESS and then play w/ the variables

account_financial_report_qweb/static/src/css/report.css
pedrobaeza: Take into account that this also affects PDF report, where a padding would affect to have 200 pages, or 400...

i-vyshnevska: @pedrobaeza  this padding not depend of amount of account entries if you have a lot of pages it add padding only in the end 
image

account_financial_report_qweb/static/src/css/report.css
i-vyshnevska: w/ padding 
image

without 
image

maybe decrease it 1px would be optimal variant

account_financial_report_qweb/static/src/css/report.css
i-vyshnevska: @pedrobaeza  what do we agreed? keep transparent for community and white for enterprise?
to do that, I don't see other option, only a creating separate module just for changing background, in a way that we want

pedrobaeza: My question was:

> Isn't a variable for enterprise background?

and my purpose is to let it transparent, but in both versions.

account_financial_report_qweb/static/src/css/report.css
pedrobaeza: Well, why wasting that precious pixels in not so wide screens? I don't see any advantage.

i-vyshnevska: it probably not about wide, cos there a lot of free space and padding left, right have no influence on columns width,
 maybe only  on mobile devices, but for mobile there problems and without  padding 
and looks fine for screens from 1024px 
image

it effects on height of report  but easier to read, open some item

pedrobaeza: I prefer to remove all the paddings.

i-vyshnevska: done

account_financial_report_qweb/static/src/less/report.less
```diff
@@ -103,4 +103,5 @@
     margin-right: auto;
     margin-left: auto;
     font-family: Helvetica, Arial;
+    padding: 20px;
```
pedrobaeza, line 106: But this padding is still changed...

pedrobaeza:
And what has happened with background?

account_financial_report_qweb/static/src/less/report.less
i-vyshnevska: because w/ less file background compiled to white color, and this padding we need

pedrobaeza: @jbeficent what do you think about these changes? I still don't like the white background and the unneeded padding

simahawk (approved):

account_financial_report_qweb/static/src/css/report.css
fclementic2c: @pedrobaeza but transparent in enterprise version means grey and I show you above that this is not very much readable. Do I understand properly ?

i-vyshnevska: @fclementic2c  it fixed and @odoo-view-background-color will give a white color in report

fclementic2c:
@pedrobaeza And if we just set the background to white only on the line where the account name appear ? because in the end all I need is to be able to read the account (black on grey is not ideal)

pedrobaeza:
I let the final word to @jbeficent

leemannd (approved):
Working.
Tested only with enterprise license.

fclementic2c:
I close it

fclementic2c:
@simahawk says it is not usefull to close it so... I reopen it

i-vyshnevska:
@jbeficent @pedrobaeza passby this PR or close it 
pls consider that it has fixes as less variables file are not reachable

PR #502, opened 2019-01-10
[10.0][IMP] account_financial_report_qweb : filter for partner

i-vyshnevska:
On wizards (for all reports) where a partner selection is possible, we should not be able to see contacts -> we only work on legal entity


simahawk (approved):
@jbeficent can you pass by? :)

leemannd (approved):
Thanks

pedrobaeza:
Please forward-port it to v11

PR #507, opened 2019-01-15
[10.0] [IMP] initialize wizard by current year

i-vyshnevska:
- set start date to 01/01/2018 if today's date is 12 dec 2018 and that fiscal year end up on 31/12 as per accounting settings - see fiscalyear_last_month in the account_config_settings
- set end date to today


i-vyshnevska:
move to https://github.com/OCA/account-financial-reporting/pull/508

PR #516, opened 2019-02-05
[FIX] customer_activity_statement: reference table on the query

etobella:
It limits how the fields can be named on res.company (also, ref is a standard field of partner_id)

### DIFF
--- a/customer_activity_statement/__manifest__.py
+++ b/customer_activity_statement/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Partner Activity Statement',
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow (approved):
Small flake 8 issues outstanding.

MiquelRForgeFlow (approved):
please, squash

and please, could you apply same fix to `_display_lines_sql_q3` method of `customer_outstanding_statement` module? 😉

etobella:
@mreficent I have rebased, I will create another PR for the other module as soon as I can :wink:

PR #517, opened 2019-02-08
[11.0][customer_outstanding_statement] minor fixes

JordiBForgeFlow:
Applies similar fixes as in #516. 
Renames the report to 'Outstanding Statement'.

### DIFF
--- a/customer_outstanding_statement/report/customer_outstanding_statement.py
+++ b/customer_outstanding_statement/report/customer_outstanding_statement.py
@@ -94,9 +94,10 @@ class CustomerOutstandingStatement(models.AbstractModel):
 
     def _display_lines_sql_q2(self):
         return """
-            SELECT partner_id, currency_id, move_id, date, date_maturity,
-                            debit, credit, name, ref, blocked, company_id,
-            CASE WHEN currency_id is not null
+            SELECT Q1.partner_id, Q1.currency_id, Q1.move_id,
+            Q1.date, Q1.date_maturity, Q1.debit, Q1.credit,
+            Q1.name, Q1.ref, Q1.blocked, Q1.company_id,
+            CASE WHEN Q1.currency_id is not null
                     THEN open_amount_currency
                     ELSE open_amount
             END as open_amount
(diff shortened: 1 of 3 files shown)


customer_outstanding_statement/report/customer_outstanding_statement.py
```diff
@@ -94,8 +94,9 @@ def _display_lines_sql_q1(self, partners, date_end, account_type):
 
     def _display_lines_sql_q2(self):
         return """
-            SELECT partner_id, currency_id, move_id, date, date_maturity,
-                            debit, credit, name, ref, blocked, company_id,
+            SELECT Q1.partner_id, Q1.currency_id, Q1.move_id, 
+            Q1.date, Q1.date_maturity, Q1.debit, Q1.credit, 
+            Q1.name, Q1.ref, Q1.blocked, Q1.company_id,
```
MiquelRForgeFlow, line 99: I think this change is not needed, there is no joint, thus no problems. But if you want to proceed with this change, then you should also put the Q1 in the `currency_id` of the line below and fix the related flake8.

MiquelRForgeFlow (changes requested):

customer_outstanding_statement/report/customer_outstanding_statement.py
JordiBForgeFlow: Fixed.

JordiBForgeFlow: Fixed

MiquelRForgeFlow: still flake8 fails 🙁

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:  Functional tests ok

customer_outstanding_statement/report/customer_outstanding_statement.py
gdgellatly: why is it Q1 in a function called q2? am I just being stupid?

EDIT: Yes I am

gdgellatly: @mreficent IRT flake8, I have a more general problem with formatting SQL syntax to fit python style. Using triple quoted strings and laying it out nicely makes for easy reading of python code, but it generates truly horrible to read SQL logs. Honestly, I feel like we are better with flake8 overrides, but just a personal opinion.

But if you are going to lay it out for python, then you should indent continuation lines of major clauses (e.g. Q1.date and Q1.name should be indented) for readability.

MiquelRForgeFlow: Previous flake8 error was that there was blank spaces and the end of some lines

HviorForgeFlow (approved):

PR #527, opened 2019-03-26
[10.0][account_financial_report_qweb] Fix to Trial Balance with parameter 'Hide account at 0'

JordiBForgeFlow:
when computing the trial balance, do not pass the parameter 'hide_account_at_0' to the computation
of the general ledger.

cc @alexis-via


alexis-via (approved):
I verified that the fix works on 2 different DB with production data

pedrobaeza (approved):

Issue #546, opened 2019-05-17
[BUG] Activity/Partner Statements Aging Balances


gdgellatly:
This is a bit tricky to explain and I can't quite diagnose it as I didn't do the original SQL.

The aging takes into account partial reconciliations, initially from Q0, but here is where the action happens. I guess the issue could be tied to Q0, and need to check the date of the associated full_reconcile_id?

https://github.com/OCA/account-financial-reporting/blob/348fb0489d0ae2717e7213cfe8e22619891e2ade/customer_activity_statement/report/customer_activity_statement.py#L181-L184

To reproduce this do the following.
With (adjust dates for 30 day buckets)

- 1 payment of $1000 in Feb 2019
- 5 invoices in March 2019 of 10,000
- 1 invoice in April 2019 for 100,000.
- 1 payment May 2019 Have a payment of 49,000

In May reconcile 1,2,4 leaving the 100k outstanding, either through bank statement or otherwise.

Print a statement for 30/04/2019.
Expected Closing Balance at 30/04 is $149,000 - this is correct on statement
Expected Aging at 30/04 is 

Aging | 1 mth | 2 mth | 3 mth | Total
---------|---------|---------|---------|-------
Expected  | 100,000 | 50,000 | -1,000 | 149,000 |
Observed | 100,000 | 40,000 | 0 | 140,000 |

MiquelRForgeFlow:
No, the Aging buckets only show the outstanding. You should compare the Outstanding Statement with the Aging buckets, not the Activity Statement with the Aging buckets.

But I checked your issue and yes, there is an error in the outstanding and the aging buckets. The problem is in the Q0 subquery and also maybe the Q1 and Q2. I will try to fix.

Aging | 1 mth | 2 mth | 3 mth | Total
-- | -- | -- | -- | --
Expected | 100,000 | 49,000 | 0 | 149,000
Observed | 100,000 | 40,000 | 0 | 140,000

gdgellatly:
@mreficent no that is incorrect. The closing balance of an activity statement must equal the sum of aging. 

Also I don't know why you have changed the expected again here. That is also wrong. The 1 payment for $1000 is unreconciled as at 30/04/2019 in the example. It therefore must be -$1000 in the 3 mth column. 

What happens is when you reconcile a full payment, that also includes another credit balance, Odoo creates a partial reconciliation record as well as a full reconciliation record.

gdgellatly:
[...]

PR #547, opened 2019-05-20
[11.0][FIX] Wrong outstanding amounts for partial reconciliations

MiquelRForgeFlow:
Fixes https://github.com/OCA/account-financial-reporting/issues/546.


JordiBForgeFlow (approved):
:+1:  Tested. Works as expected.

gdgellatly (approved):
Tested corresponding patch for V12 against live data and confirm working. Code comparison here with that patch. No test.

PR #548, opened 2019-05-20
IMP tax_balance: breadcrumb name

bealdav:
Without this code it only display `Taxes Balances` in breadcrumb

tax_bal

cc @eLBati @sbidoul @antespi @astirpe @alexis-via

### DIFF
--- a/account_tax_balance/wizard/open_tax_balances.py
+++ b/account_tax_balance/wizard/open_tax_balances.py
@@ -1,7 +1,7 @@
 # © 2016 Lorenzo Battistini - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 
 class WizardOpenTaxBalances(models.TransientModel):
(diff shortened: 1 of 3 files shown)


eLBati (approved):
Thanks! Could you just increase version following https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#version-numbers ?

account_tax_balance/wizard/open_tax_balances.py
```diff
@@ -31,7 +31,23 @@ def onchange_date_range_id(self):
...
+            [('code', '=', self.env.user.lang or 'en_US')])
+        date_format = lang and lang.date_format or "%m/%d/%Y"
+        infos = {'name': vals['name'], 'target': _(self.target_move),
+                 'from': self.from_date.strftime(date_format),
+                 'to': self.to_date.strftime(date_format),
+                 'company': self.company_id.name}
+        # name of action which is displayed in breacrumb
+        vals["name"] = _(
+            "%(name)s: %(target)s from %(from)s to %(to)s"
+            " (%(company)s)") % infos
```
astirpe: IMO the company should be displayed only in case the user belongs to the "Multi Companies" group.

astirpe (approved):
:+1: 
Tested locally, since Runbot is not available.

bealdav:
Thanks for your review

account_tax_balance/wizard/open_tax_balances.py
```diff
@@ -31,7 +31,25 @@ def onchange_date_range_id(self):
     def open_taxes(self):
         self.ensure_one()
         action = self.env.ref('account_tax_balance.action_tax_balances_tree')
-        vals = action.read()[0]
+        act_vals = action.read()[0]
+        # override action name doesn't work in v12 or v10
+        # we need to build a dynamic action on main keys
+        vals = {x: act_vals[x] for x in act_vals
```
pedrobaeza, line 37: No need to make this filtering

bealdav: @pedrobaeza  If I don't do that it doesn't work, sorry ;-)

pedrobaeza: That's very weird, as previous code was working also with a raw read...

bealdav: In previous code there was no override of the action name.

In returned vals the override name is transmit to all args of the stack but the problem comes from _compute_display_name() which compute another value than overrided name

pedrobaeza: OK, let's go this way.

pedrobaeza (approved):

account_tax_balance/wizard/open_tax_balances.py
bealdav: Don't hesitate to try both code to see if I missed smth or not if any doubt 

Thanks Pedro

pedrobaeza: And thanks for the explanation :)

PR #549, opened 2019-05-20
[10.0][FIX] Wrong outstanding amounts for partial reconciliations

MiquelRForgeFlow:
Fixes https://github.com/OCA/account-financial-reporting/issues/546. Backport of https://github.com/OCA/account-financial-reporting/pull/547 and https://github.com/OCA/account-financial-reporting/pull/565.

### DIFF
--- a/customer_activity_statement/__manifest__.py
+++ b/customer_activity_statement/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Customer Activity Statement',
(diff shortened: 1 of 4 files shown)


gdgellatly (approved):
Tested corresponding patch for V12 against live data and confirm working. Code comparison here with that patch. No test.

dominiquekonsuntack:
in v10,
the PDF report also shows the invoices that are paid (balance = 0.00)

PR #554, opened 2019-05-28
[11.0][IMP] account_financial_report: show inactive partners in wizard

MiquelRForgeFlow:
As @pedrobaeza requested.

### DIFF
--- a/account_financial_report/wizard/aged_partner_balance_wizard.py
+++ b/account_financial_report/wizard/aged_partner_balance_wizard.py
@@ -53,7 +53,8 @@ class AgedPartnerBalanceWizard(models.TransientModel):
                 self.account_ids = self.account_ids.filtered(
                     lambda a: a.company_id == self.company_id)
         res = {'domain': {'account_ids': [],
-                          'partner_ids': []}}
+                          'partner_ids': ['|', ('active', '=', False),
+                                          ('active', '=', True)]}}
         if not self.company_id:
             return res
         else:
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

rafaelbn (approved):

alexis-via:
With the current code of the v10.0 branch, the inactive partners that have credit/debit/balance are shown in the report ; this is very important and it works well without this PR.

This PR is about the ability to select an inactive partner in the M2M field "Filter Partners" in the wizard. I think it's not a good idea ; if you want to explicitly filter on an inactive partner, you should just re-active it. For me, it makes more sense.

So I would prefer that we cancel this PR.

pedrobaeza:
Thanks for checking, @alexis-via.

This was detected in 11.0 by one of our customers, and they say they are filtering the account, but not the partner, but inactive partners are not shown with this combination. I thought this search was the one for populating partners, not for showing them in the partner selector. Are you sure this is properly working in both 10.0 and 11.0 without this PR?

pedrobaeza:
@jbeficent did you read what @alexis-via wrote? Maybe this is not the proper patch.

pedrobaeza:
I'm checking and indeed it's not. This only changes domain for `partner_ids` field. What we want is to include inactive partners in open items report. Steps to reproduce the problem:

1. Create a partner with customer=True.
2. Create a customer invoice for that partner and validate it.
3. Deactivate the partner.
4. Print open items report.
5. The report doesn't show the receivable of that partner.

PR #562, opened 2019-06-02
[FIX] Aging Buckets in partner_statement

gdgellatly:
In certain scenarios involving mutliple payments for multiple
invoices, where the reconciliation happens together and the debit
move is the smaller item the aging duplicates its value, making
the balance greater than it should be.

This fix works by magic by adding the partial reconciliation ids
to the group by clause of _show_buckets_sql_q1

### DIFF
--- a/partner_statement/report/report_statement_common.py
+++ b/partner_statement/report/report_statement_common.py
@@ -74,7 +74,7 @@ class ReportStatementCommon(models.AbstractModel):
                                 ) AND l.date <= %(date_end)s AND not l.blocked
             GROUP BY l.partner_id, l.currency_id, l.date, l.date_maturity,
                                 l.amount_currency, l.balance, l.move_id,
-                                l.company_id
+                                l.company_id, l.id
         """, locals()), "utf-8")
 
     def _show_buckets_sql_q2(self, date_end, minus_30, minus_60, minus_90,


gdgellatly:
@mreficent I can't explain this fix, I can't even reproduce the issue. I've never had this happen to me before. 

I started out manually running the statement SQL to try and reconcile a difference between aging buckets and closing balances. Then I added some more columns to get more information to try and investigate which entries caused the issue. In doing so added the partial_reconcile ids to the group by clause, figuring they were part of the problem. 

All of a sudden in my test SQL, on the partner in question I got 2 extra lines at 0.00 and 1 less line at 158.13 (which was the amount the statement was out by, of which I had 2 previously) and a correct balance (so 80 vs 79 lines). So I scaled the query back to this fix you see here.

I then tested it against another known problem account and it too was fixed. I still want to run some kind of assertion test against all balances. Do you have any idea why this might work? I also wonder if it isn't somehow related to those other fixes we had the other day on lines and opening balances.

gdgellatly:
OK I think I worked it out. Because the GROUP BY clause was grouping on move ids, it was possible that a move could have multiple individual lines with the same partial reconciliation that meant they were grouped. By adding the line id to the GROUP BY clause we prevent this issue.

Note I updated commit message, but left PR message as original.

EDIT: OK I definitely worked it out, and this is definitely a correct fix. The problem was that the payment in question paid 2 invoices of identical amounts. Both invoices credit counterparts therefore appear in the same move, and were incorrectly grouped.

EDIT EDIT: The second 158.13 was spurious, it was actually a third invoice for the same amount that should be outstanding.

MiquelRForgeFlow (approved):
LGTM :+1:  thanks!

gdgellatly:
@hveficent @jbeficent do you mind looking at this please? I've found another rather critical bug in balance calculations (not considering whether a move is posted or cancelled) and I'd rather this was finished up first before building on it.

HviorForgeFlow (approved):

PR #570, opened 2019-06-07
[10.0] account_financial_report_qweb: supplier ref in general ledger

p-tombez:
Display supplier reference in column label in general ledger report

### DIFF
--- a/account_financial_report_qweb/__manifest__.py
+++ b/account_financial_report_qweb/__manifest__.py
@@ -16,7 +16,8 @@
     "website": "https://odoo-community.org/",
     'depends': [
         'account',
-        'account_group',    # account-financial-tools
+        'account_group',                        # account-financial-tools
+        'account_invoice_supplier_ref_unique',  # account-invoicing
         'date_range',
         'report_xlsx',
         'report',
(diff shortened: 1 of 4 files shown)


account_financial_report_qweb/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'QWeb Financial Reports',
-    'version': '10.0.3.1.0',
+    'version': '10.0.3.0.1',
```
guewen: Hmm?

```suggestion
    'version': '10.0.3.2.0',
```

guewen (changes requested):
LGTM but there is a problem with the version.

asaunier (approved):
Try perhaps to rebase the branch in order to fix Travis errors.

asaunier (approved):

vrenaville (approved):

guewen (approved):

account_financial_report_qweb/__manifest__.py
```diff
@@ -16,7 +16,8 @@
     "website": "https://odoo-community.org/",
     'depends': [
         'account',
-        'account_group',    # account-financial-tools
+        'account_group',                        # account-financial-tools
+        'account_invoice_supplier_ref_unique',  # account-invoicing
```
pedrobaeza, line 20: You can't change dependencies in a stable version, and more adding this dependency that is not mandatory. You must do a bridge module if you want that data.

pedrobaeza (changes requested):

Issue #571, opened 2019-06-07
In account_financial_report, href to an unexisting css file


alexeirivera87:
In account_financial_report/report/templates/layouts, branch 11.0, we have this line
`<link href="/account_financial_report/static/src/css/report.css" rel="stylesheet"/>`

but that file report.css does not exists

Regards

pedrobaeza:
Fixed in 2fe67661b1433b810902727fa1db8ff8c3ea210a

i-vyshnevska:
@pedrobaeza reverting for v11 is wrong, as reported branch is wrong 
in v11 used **account_financial_report_qweb** and bug reported about **account_financial_report**

PR #572, opened 2019-06-08
[FIX] set apropriate link to style file

i-vyshnevska:
related to https://github.com/OCA/account-financial-reporting/issues/571

### DIFF
--- a/account_financial_report/report/templates/layouts.xml
+++ b/account_financial_report/report/templates/layouts.xml
@@ -3,7 +3,7 @@
 
     <template id="account_financial_report.internal_layout">
         <div class="article">
-            <link href="/account_financial_report/static/src/css/report.css" rel="stylesheet"/>
+            <link href="/account_financial_report/static/src/less/report.less" rel="stylesheet"/>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Sorry, but as stated on the commit, this is putting a non desired background and now the renaming is reverted, so there's no sense for this.

i-vyshnevska:
@pedrobaeza if you want to have specific color so set it in styles in https://github.com/OCA/account-financial-reporting/blob/2fe67661b1433b810902727fa1db8ff8c3ea210a/account_financial_report/static/src/css/report.css#L98 but this is totally wrong to keep broken files, it'll give you unpredictable behavior

pedrobaeza:
Offending variables removed in https://github.com/OCA/account-financial-reporting/commit/44c30b45517843be0c4a710469c5b6afae1036ab

i-vyshnevska:
for v 10 pls too

pedrobaeza:
Done in d52934409bf396de6b12fa2e819190598150181f

PR #573, opened 2019-06-12
[IMP] add index in journal ledger transient model

sbidoul:
I noticed an issue when garbage collecting this financial report transient model.
When deleting from report_journal_ledger_move, the cascading delete of 
reporting move lines is extremely slow.
This is apparently due to a missing foreign key index.


sbidoul:
Probably not a good idea to add such index, for performance reason. A better solution might be to cleanup the report move lines before the report moves and thus avoiding the slow cascading delete.

pedrobaeza:
Closing as definitive solution is in #582

PR #574, opened 2019-06-13
12.0 ensure onchanges applied to direct report calls

gdgellatly:
[FIX] Ensure onchanges applied when calling report directly  …
Prior to this commit create was called in the report in order to
get the required data. However date_end was not defaulted and relied
on an onchange to get the correct date.

In order to fix we use the new Odoo tests Form functionality to
emulate calling the wizard form just as in UI.

[FIX] Add missing default in res_config_settings and bump version


partner_statement/report/activity_statement.py
```diff
@@ -140,9 +141,9 @@ def _get_report_values(self, docids, data):
         if not data:
             data = {}
         if 'company_id' not in data:
-            wiz = self.env["activity.statement.wizard"].with_context(
+            wiz = Form(self.env["activity.statement.wizard"].with_context(
```
MiquelRForgeFlow, line 144: Don't! This `Form` server was created only for testing. That's why it is found in odoo.tests 😏

MiquelRForgeFlow (changes requested):

partner_statement/report/activity_statement.py
pedrobaeza: @mreficent it's on tests, but it's used in other parts in Odoo (see vendor bill import). I see this better than simply call all the onchanges you know (and leaving unknown onchanges), or to depend on `onchange_helper` module, that is buggy in some cases (when there's a context in fields).

gdgellatly: @mreficent In this particular case, it is the precise behaviour desired. As though the user opened the form, the defaults were populated, and triggered the onchanges. I figured if it is considered good enough for core for basically the same cases, then it must be OK.

gdgellatly:
Wow I must have messed this up sometime, guess we just forked and kept going seperately.

PR #582, opened 2019-06-21
[11.0][FIX] account_financial_report: Extremely auto vaccum query execute to delete report_journal_ledger table

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/account_financial_report/readme/CONTRIBUTORS.rst
+++ b/account_financial_report/readme/CONTRIBUTORS.rst
@@ -16,6 +16,10 @@
 * Mihai Fekete <feketemihai@gmail.com>
 * Miquel Raïch <miquel.raich@eficent.com>
 * Isaac Gallart <igallart@puntsistemes.com>
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * Pedro M. Baeza
+  * Sergio Teruel
 
 Much of the work in this module was done at a sprint in Sorrento, Italy in
 April 2016.
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -612,6 +612,19 @@ class ReportJournalLedger(models.TransientModel):
     def get_html(self, given_context=None):
         return self._get_html()
 
+    @api.model
+    def _transient_vacuum(self, force=False):
+        """Remove journal ledger subtables first for avoiding a costly
+        ondelete operation.
+        """
+        # Next 3 lines adapted from super method for mimicking behavior
+        cls = type(self)
+        if not force and (cls._transient_check_count < 21):
+            return True  # no vacuum cleaning this time
+        self.env.cr.execute("DELETE FROM report_journal_ledger_move_line")
+        self.env.cr.execute("DELETE FROM report_journal_ledger_move")
+        return super(ReportJournalLedger, self)._transient_vacuum(force=force)
+
 
 class ReportJournalLedgerJournal(models.TransientModel):
 


rafaelbn (approved):

pedrobaeza:
@sbidoul I have been investigating this, and indeed deleting rows directly avoid the problem. I have implemented the deletion on the more accurate method I have found before actual vacuum. If it's OK for you, let's merge this and I will cherry-pick it for 10.0 and 12.0 (I have kept `super` full form for avoiding problems in v10/P2).

pedrobaeza:
Another comment is that other tables like `report_general_ledger` don't suffer this problem because sub-tables already have index on the m2o field, but I wonder if that index was added as worse solution to the same problem...

sbidoul:
@pedrobaeza in this module there is already an override of `_transient_clean_rows_older_than`. Would it be simpler to override that one?

pedrobaeza:
There are 2 possible vacuum methods:

* By expiration time:

  https://github.com/odoo/odoo/blob/0458337bc362b4ecc020fd95a5896dc9d394322b/odoo/models.py#L4187

* By expiration count:

  https://github.com/odoo/odoo/blob/0458337bc362b4ecc020fd95a5896dc9d394322b/odoo/models.py#L4190

That's why I have preferred to overwrite this one for always be sure that we remove sub-tables lines before.

javiermatos:
> Another comment is that other tables like `report_general_ledger` don't suffer this problem because sub-tables already have index on the m2o field, but I wonder if that index was added as worse solution to the same problem...

My issue was with the `report_journal_ledger` table as explained in #607

pedrobaeza:
@javiermatos that table is exactly handled here. Please read carefully whole thread and the other linked closed PR

javiermatos:
> @javiermatos that table is exactly handled here. Please read carefully whole thread and the other linked closed PR

Yeah, sorry, my mistake. Thank you for the fix you created.

I have noticed the following in **report_journal_ledger_move** for v11.0:

```
TABLE "report_journal_ledger_move_line" CONSTRAINT "report_journal_ledger_move_line_report_move_id_fkey" FOREIGN KEY (report_move_id) REFERENCES report_journal_ledger_move(id) ON DELETE CASCADE
```

Which I think it means that by deleting records from **report_journal_ledger_move** you will also delete records from **report_journal_ledger_move_line**. Would it make sense to run those two queries in the opposite order to avoid that CASCADE? I think it should run even faster that way.

pedrobaeza:
@javiermatos Well spotted, thank you very much! Changed that order as well.

rafaelbn:
Do we agree here please? this is already tested in production with :+1: results

pedrobaeza:
@sbidoul do you agree with this?

pedrobaeza:
As no answer, merging as the solution has proved to be correct

/ocabot merge patch

pedrobaeza:
Cherry-pick for v12 in dceebedd5c6f55478f0b1cf8b2333cc17453f769

ljsalvatierra-factorlibre:
Hi @pedrobaeza, sorry for commenting on a closed issue, just out of curiosity, is there any reason to use `DELETE` instead of `TRUNCATE`, in theory should be quicker.

pedrobaeza:
If I don't remember bad, you can't use truncate without cascade, and that can lead to undesired deletions on related tables.

ljsalvatierra-factorlibre:
> If I don't remember bad, you can't use truncate without cascade, and that can lead to undesired deletions on related tables.

Thank you for your time, i will look into it.

claudiofrigerio:
Hi, 
i have issues also with this fix. I use V11 
I have 500k rows on report_journal_ledger_move and it's impossible to run _DELETE FROM public.report_journal_ledger_move;_
I tried to run this query for many hours.

i solved this issue on test environment with this query:
_TRUNCATE ONLY report_journal_ledger CASCADE_

I get this output:

_NOTICE:  truncate cascades to table "account_journal_report_journal_ledger_rel"
NOTICE:  truncate cascades to table "report_journal_ledger_journal"
NOTICE:  truncate cascades to table "report_journal_ledger_journal_tax_line"
NOTICE:  truncate cascades to table "report_journal_ledger_move"
NOTICE:  truncate cascades to table "report_journal_ledger_move_line"
NOTICE:  truncate cascades to table "report_journal_ledger_report_tax_line"
TRUNCATE TABLE_

Anyone know if there are any problems if all rows of the output tables are deleted? This tables seem transient models, but I would like to have confirmation. Thank you
I solved this issue on test e

ljsalvatierra-factorlibre:
> Hi, i have issues also with this fix. I use V11 I have 500k rows on report_journal_ledger_move and it's impossible to run _DELETE FROM public.report_journal_ledger_move;_ I tried to run this query for many hours.
> 
> i solved this issue on test environment with this query: _TRUNCATE ONLY report_journal_ledger CASCADE_
> 
> I get this output:
> 
> _NOTICE: truncate cascades to table "account_journal_report_journal_ledger_rel" NOTICE: truncate cascades to table "report_journal_ledger_journal" NOTICE: truncate cascades to table "report_journal_ledger_journal_tax_line" NOTICE: truncate cascades to table "report_journal_ledger_move" NOTICE: truncate cascades to table "report_journal_ledger_move_line" NOTICE: truncate cascades to table "report_journal_ledger_report_tax_line" TRUNCATE TABLE_
> 
> Anyone know if there are any problems if all rows of the output tables are deleted? This tables seem transient models, but I would like to have confirmation. Thank you I solved this issue on test e

Hi @odoogitlexter, you can also run it like this:

```sql
TRUNCATE ONLY report_journal_ledger_move, report_journal_ledger_move_line RESTART IDENTITY
```

claudiofrigerio:
thanks, it's works. Can we insert this fix to master branch?

PR #583, opened 2019-06-21
[10.0][FIX] account_financial_report: Extremely auto vaccum query execute to delete report_journal_ledger table

sergio-teruel:
Backport of #582

cc @Tecnativa

### DIFF
--- a/account_financial_report_qweb/report/journal_report.py
+++ b/account_financial_report_qweb/report/journal_report.py
@@ -715,6 +715,7 @@ class ReportJournalQwebMoveLine(models.TransientModel):
         comodel_name='report_journal_qweb_move',
         required=True,
         ondelete='cascade',
+        index=True,
     )
     move_line_id = fields.Many2one(
         comodel_name='account.move.line',


sbidoul:
I made the same in #573.

But I think if the only benefit is the autovacuum performance, then it's better to delete the report move lines in a a first step before deleting the moves, so we don't rely on the cascading delete.
This way we don't have the performance hit of maintaining an index that is otherwise useless.

pedrobaeza:
@sbidoul Isn't the query `DELETE FROM .. WHERE report_id = X` will be the same slow as DELETE CASCADE without the index?

sbidoul:
@pedrobaeza I'm not sure. The delete of the move needs to search the related move lines to cascade the delete, then the index is needed. If we delete the lines first and don't do a cascade from the moves, then there is no need to search move lines by move id and the index is not necessary. I've not tested.

rafaelbn:
Functionaly this works so :+1: 

If @sbidoul did the same in https://github.com/OCA/account-financial-reporting/pull/573 we can merge that one then

rafaelbn (approved):

pedrobaeza:
Closing as we will cherry-pick #582

PR #561, opened 2019-05-31
[11.0][account_financial_report] Change GL PDF report layout

Tonow-c2c:
To fix this issue [https://github.com/OCA/account-financial-reporting/issues/556](https://github.com/OCA/account-financial-reporting/issues/556)

And this #558

Issue #558, opened 2019-05-29
[11.0][account_financial_report] :warning: pdf and html report have same name


Tonow-c2c:
![image](https://user-images.githubusercontent.com/32262135/58571765-b6e9cb00-823a-11e9-9540-9bfe1684805b.png)

For report creation we pass by this method [_get_report_from_name](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/ir/ir_actions_report.py#L424-L425)

> She get the first record of ir.actions.report having the ``report_name`` as value for the field report_name.

So in some case (like change paperformat of one of them) we have a issue if we have same name for different reports  :warning: 

I propose to re-name report like : 
* `account_financial_report.report_...._qweb_pdf` for pdf report
* `account_financial_report.report_...._qweb_html` for html report

I will create PR soon

pedrobaeza:
OK, good for me.

Tonow-c2c:
@pedrobaeza should we merge this ?

pedrobaeza:
See my comments in the PR, and it's also conflicting right now in v11.

PR #585, opened 2019-06-27
Fix issues 558

Tonow-c2c:
To fix this issus https://github.com/OCA/account-financial-reporting/issues/558

@pedrobaeza when this PR will be merge I refactor https://github.com/OCA/account-financial-reporting/pull/561/commits
that will be easier like this


pedrobaeza:
@jbeficent what do you think about this?

@Tonow-c2c I'm personally seeing that this seems a bit more complicated that I initially thought. What are the advantages of splitting this? The paperformat you mention on the issue doesn't apply to HTML reports.

PR #592, opened 2019-07-12
[12.0][FIX] - remove many2many_tags on account_ids field in general ledger wizard view

sbejaoui:
many2many_tags don't support the selection of multiple records at the same time in "search more" wizard.
This make users sets accounts one by one. A huge waste of time

Before:

After:

### DIFF
--- a/account_financial_report/wizard/general_ledger_wizard_view.xml
+++ b/account_financial_report/wizard/general_ledger_wizard_view.xml
@@ -34,7 +34,6 @@
                             </group>
                             <field name="account_ids"
                                    nolabel="1"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This has been treated several times and even said in the README on some version: use `web_widget_many2many_tags_multi_selection` module for having full power on the selection, but don't remove this widget as it's very useful for saving screen space.

sbejaoui:
thanks @pedrobaeza

pedrobaeza:
If the mention is not done in v12 README, we can add it, as this is somehow a FAQ

PR #599, opened 2019-07-25
[12.0][ADD] partner_statement: migration scripts to v12

MiquelRForgeFlow:
Adds migration scripts.


gdgellatly:
@mreficent given it is a different module, and you could just drop the tables if they exist is it really necessary? Anyway I have no objections to the actual code if you think it is, so consider this a +1.

MiquelRForgeFlow:
We have migrated several data bases with this, it works fine.

gdgellatly (approved):

AdriaGForgeFlow (approved):
Code review :+1:

HviorForgeFlow (approved):

PR #601, opened 2019-07-31
[12.0][FIX] partner_statement: Covers the case where name and ref are strictly the same in account_move_line

cvinh:
Fixes the following use case : 
If the fields name and ref are the same in account_move_line, nothing appears in the description column of the reports

### DIFF
--- a/partner_statement/views/activity_statement.xml
+++ b/partner_statement/views/activity_statement.xml
@@ -76,6 +76,9 @@
                                                     <t t-if="line['ref'] not in line['name']">
                                                         <span t-esc="line['ref']"/>
                                                     </t>
+                                                    <t t-else="">
+                                                        <span t-esc="line['name']"/>
+                                                    </t>
                                                 </t>
                                             </t>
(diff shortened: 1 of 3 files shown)


gdgellatly:
This looks wrong to me. It appears to break the case where they aren't the same, however one is in the other, and you would have both appear.

partner_statement/views/activity_statement.xml
```diff
@@ -76,6 +76,9 @@
                                                     <t t-if="line['ref'] not in line['name']">
```
MiquelRForgeFlow:
```suggestion
                                                     <t t-if="line['ref'] not in line['name'] or line['ref'] == line['name']">
```
please, do this instead

MiquelRForgeFlow:
and rebase

MiquelRForgeFlow:
Superseded by https://github.com/OCA/account-financial-reporting/pull/652.

MiquelRForgeFlow (changes requested):
Should be closed.

PR #608, opened 2019-08-22
[11.0]account_financial_error: Problem with partner filter when partners are shared among companies

jesusVMayor:
The company_id domain has no sense when the partners are shared among the companies.
https://github.com/OCA/account-financial-reporting/pull/539

I don't know what i made in the other PR, but i broke it trying to resolve the conflicts.


pedrobaeza:
This is what translating company domain from record rules to "multi-company force" context implies... I don't like to need to do all this stuff. I vote for removing all the company stuff, and @etobella must handle this in custom.

PR #613, opened 2019-09-16
[12.0][IMP] account_tax_balance: multicompany fiscal unit

astirpe:
In some countries it is allowed to declare to the tax authorities one single tax report that contains data of multiple companies. For example, in The Netherlands a group of companies (one parent company and N children companies) can form a fiscal unit; in this case, the parent company declares one single tax statement for all companies of the fiscal unit (see https://github.com/OCA/l10n-netherlands/issues/237).

In the current `account_tax_balance` module, you can only see the data regarding move lines of the company you're logged in. This is fine when using the module itself but it could be a bit uncomfortable when extending the module to generate tax statements for multiple companies.

This PR make the extension of this module easier in case of generating reports for multiple companies.
Without this PR, it is still possible to do it, but that requires overriding the methods `get_move_line_partial_domain()` and `_account_tax_ids_with_moves()`.

Not sure if this patch should set the version of the module to `12.0.2.0.0`, since the type of one returned value of `get_context_values()` method is changed. In any case, I think it's a good idea to include this patch when migrating `account_tax_balance` module to V13.

### DIFF
--- a/account_tax_balance/models/account_tax.py
+++ b/account_tax_balance/models/account_tax.py
@@ -34,10 +34,11 @@ class AccountTax(models.Model):
 
     def get_context_values(self):
         context = self.env.context
+        actual_company_id = context.get("company_id", self.env.user.company_id.id)
         return (
             context.get('from_date', fields.Date.context_today(self)),
             context.get('to_date', fields.Date.context_today(self)),
-            context.get('company_id', self.env.user.company_id.id),
+            context.get('company_ids', [actual_company_id]),
             context.get('target_move', 'posted'),
         )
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is incorrect, as it's bypassing the multi-company way of working in version 12. You can change the expression for using child of operator and take parent companies + children, but not to show all the authorized companies at the same time.

astirpe:
@pedrobaeza the list of companies ids is passed by context from any localization (or custom) module extending this one. The list of companies depends on how the localization module is supposed to work.

See for example https://github.com/OCA/l10n-netherlands/issues/237 and its proposed solution https://github.com/OCA/l10n-netherlands/pull/238/files.
As you can notice, the list of companies to be reported is calculated and can be different from the classic parent-child. In our module it also depends on the country and the currency.

pedrobaeza (approved):
OK, seen

PR #615, opened 2019-09-18
account_financial_report: fix Trial Balance account.group ordering issue

### LINKED ISSUE #609: [BUG] account_financial_report: Trial Balance groups order
module: account_financial_report
version: 12.0 (11.0 and even 10.0 seems to also be effected)

Probably related issues #604 #463 #495 that people might find when having this problem.

**Current behavior**
In Trial Balance report, groups are ordered incorrectly. See image below. As far as I figured it out, this is caused by how the module uses `account.group` module's field `parent_path` to generate a sequence. The problem is that this field is of type Char that is made up of integer ids separated by slashes. Which means that in case where there is an account group **1** with id `1`, group **2** with id `2` and group **3** with id `10`, the ordering is seen like:
`"1/" < "10/" < "2/"`
And this is pseudo correct programming statement when working with char strings.
This is exactly what I managed to reproduce on community runbot, where the screenshot is taken from.

image

**Steps to reproduce**
- On a freshly created new database with demo data (so that new `account.group` records will get ids starting from `1`), create account groups in this order:
1. *name:* `1` *Code Prefix:* `1` *Parent:* ` `
2. *name:* `2` *Code Prefix:* `2` *Parent:* ` `
3. *name:* `21` *Code Prefix:* `21` *Parent:* `2 2`
4. *name:* `22` *Code Prefix:* `22` *Parent:* `2 2`
5. *name:* `23` *Code Prefix:* `23` *Parent:*`2 2`
6. *name:* `24` *Code Prefix:* `24` *Parent:* `2 2`
7. *name:* `25` *Code Prefix:* `25` *Parent:* `2 2`
8. *name:* `26` *Code Prefix:* `26` *Parent:* `2 2`
9. *name:* `27` *Code Prefix:* `27` *Parent:* `2 2`
10. *name:* `3` *Code Prefix:* `3` *Parent:* ` `
- Try generating the Trial Balance. The dates don't matter, the important part is to have **Hide accounts at 0** option unchecked, so that all groups get printer.

gaikaz:
Fixes #609 and probably related issues mentioned there.

This simply changes how `sequence` field for Trial Balance lines is generated.
Before this PR, `parent_path` is used, which is not always correct for this use, as it is a `Char` field made out of integer IDs. That makes this a valid ordering statement: `'11' < '2'`.
After this PR, `code_prefix` is used, which is also a `Char` field, but represents actual Account Group codes that should be sequential by idea.
Also, `code_prefix` is used as the default for `_order` field in `account.group` model:
https://github.com/odoo/odoo/blob/12.0/addons/account/models/account.py#L367

More details and a way to reproduce the problem can be found on the referenced issue thread: #609

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -75,8 +75,8 @@ class TrialBalanceReportAccount(models.TransientModel):
     )
     hide_line = fields.Boolean(compute='_compute_hide_line')
     # Data fields, used to keep link with real object.
-    # Sequence is a Char later built with 'parent_path' for groups
-    # and parent_path + account code for accounts
+    # Sequence is a Char later built with 'code_prefix' for groups
+    # and code_prefix + account code for accounts
     sequence = fields.Char(index=True, default='1')
     level = fields.Integer(index=True, default=1)
 
@@ -413,7 +413,7 @@ SELECT
     accgroup.parent_id,
     coalesce(accgroup.code_prefix, accgroup.name),
     accgroup.name,
-    accgroup.parent_path,
+    accgroup.code_prefix,
     accgroup.level
 FROM
     account_group accgroup"""


amcor (approved):
Tested with the spanish chart account

pedrobaeza (approved):

PR #617, opened 2019-09-25
[12.0][FIX] - Hide ending balance if general ledger is filtered by partners

sbejaoui:


account_financial_report/report/templates/general_ledger.xml
```diff
@@ -67,9 +67,11 @@
                     </t>
 
                     <!-- Display account footer -->
-                    <t t-call="account_financial_report.report_general_ledger_ending_cumul">
```
pedrobaeza, line 70: You can put `t-if` clause directly here

sbejaoui: yes,
Thank you.

pedrobaeza (approved):

RobinetDenisAcsone (approved):

PR #619, opened 2019-09-30
[13.0][MIG] account_tax_balance

astirpe:
Depends on:

- [x] https://github.com/OCA/server-ux/pull/95

Includes:

- [ ] https://github.com/OCA/account-financial-reporting/pull/613


feketemihai:
@astirpe There is a good report over tax directly on v13 in Odoo base that could be configured more based on tags, countries..https://github.com/odoo/odoo/blob/master/addons/account/models/account.py#L62
Can we just adapt it?

astirpe:
@feketemihai how do you propose to "adapt it"?
This module never used tags in previous versions.

feketemihai:
@astirpe I was thinking if you can use the report to get the computed fields

astirpe:
IMO this module is meant to do something different than what `account.tax.report.line` does.

pedrobaeza:
Yes, I think both are needed.

MiquelRForgeFlow:
Is it still wip?

astirpe:
This is ready for being reviewed IMO

ernestotejeda (approved):

account_tax_balance/tests/test_account_tax_balance.py
```diff
@@ -183,13 +184,14 @@ def test_tax_balance(self):
                 'name': 'Bank Fees',
                 'tax_ids': [(4, tax.id)]
             }), (0, 0, {
-                'account_id': tax.account_id.id,
+                'account_id': tax_account_id,
                 'debit': 0,
                 'credit': 10,
                 'name': 'Bank Fees',
-                'tax_line_id': tax.id,
+                'tax_repartition_line_id': tax_repartition_line.id,
```
MiquelRForgeFlow: I don't know why you change this line. I think it should be as before. 🤔

astirpe: Because [ `tax_line_id` is a compute field](https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L2454-L2455) and doesn't have any method for the inverse.

MiquelRForgeFlow: :+1:

MiquelRForgeFlow (approved):
code review LGTM

rafaelbn (approved):
Ready to merge by a PSC :+1:

pedrobaeza:
@astirpe can you please cherry-pick #647 for performance reasons on install?

pedrobaeza (changes requested):
I have put comments in #613, and for v13, Odoo has changed the way multi-companies work (now you have `self.env.company` and `self.env.companies`), and you should change according this the queries you were modifying in that PR.

astirpe:
@pedrobaeza the list of companies ids is passed by context from any localization (or custom) module extending this one. The list of companies depends on how the localization module is supposed to work.

See for example https://github.com/OCA/l10n-netherlands/issues/237 and its proposed solution https://github.com/OCA/l10n-netherlands/pull/238/files.
As you can notice, the list of companies to be reported is calculated and can be different from the classic parent-child. In our module it also depends on the country and the currency.

I would keep the code as it is.
If you don't like this way, I could remove this change from this PR and move it to a separate PR. At least we don't block the porting.

pedrobaeza (approved):

PR #626, opened 2019-10-29
[10.0][IMP] account_financial_repot_qweb: General Ledger optimization

HviorForgeFlow:
Before this change, General Ledger report was fetching a lot of data that was not used in rendered report. Filtering partners and account used only in the selected date range make this report usable in large databases.

General Ledger report still showing correctly Journal Items related transactions in the selected date range.

In case we run Trial Balance these optimizations is not applied because all history is required to display correctly initial period balances headers for account/partners.

cc ~ @Eficent


JordiBForgeFlow:
@alexis-via what do you think about this? It really boosts performance

PR #627, opened 2019-10-30
[11.0][IMP] account_financial_report: General Ledger optimization

HviorForgeFlow:
Forward Port: #626 

cc ~ @Eficent @etobella


patel1008b:
@eficent @etobella @HviorForgeFlow

HviorForgeFlow:
@bhaveshselarka This was not the proper way to handle the problem. I think I'm going to close this PR. It is known that saving huge quantity of records on a transient model does not allow auto vacuum to work properly. We are working in a new better approach.

patel1008b:
Hello @HviorForgeFlow , Thanks for sharing information.

PR #629, opened 2019-11-07
[10.0][IMP] account_financial_repot_qweb: Open Items optimization

HviorForgeFlow:
CC ~ @Eficent @jbeficent 

This PR aims to speed up Open items renderization when is launched at current date.

When Open Items is created at current date it is not required to look into all account moves lines, we just focus on not reconciled items, limiting number of account move lines to be computed, for the same result.

### DIFF
--- a/account_financial_report_qweb/report/open_items.py
+++ b/account_financial_report_qweb/report/open_items.py
@@ -27,6 +27,7 @@ class OpenItemsReport(models.TransientModel):
     company_id = fields.Many2one(comodel_name='res.company')
     filter_account_ids = fields.Many2many(comodel_name='account.account')
     filter_partner_ids = fields.Many2many(comodel_name='res.partner')
+    is_today = fields.Boolean()
 
     # Data fields, used to browse report data
     account_ids = fields.One2many(
@@ -239,6 +240,10 @@ WITH
             AND
                 p.id IN %s
             """
+        if self.is_today:
+            query_inject_account += """
+            AND ml.reconciled = False
+            """
         query_inject_account += """
             GROUP BY
                 a.id, c.id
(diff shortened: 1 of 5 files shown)


JordiBForgeFlow (approved):
:+1: This really speeds up the report for the most usual use case, that is to query for today's due amounts.

JordiBForgeFlow:
cc @rvalyi @alexis-via what do you think?

rousseldenis:
@HviorForgeFlow @JordiBForgeFlow We have the same issue for partner balance.

Shouldn't we add an option on aged balance to get only unreconiled items ?

We get from 1 hour to 4 seconds with this ....

account_financial_report_qweb/report/open_items.py
```diff
@@ -27,6 +27,7 @@ class OpenItemsReport(models.TransientModel):
     company_id = fields.Many2one(comodel_name='res.company')
     filter_account_ids = fields.Many2many(comodel_name='account.account')
     filter_partner_ids = fields.Many2many(comodel_name='res.partner')
+    is_today = fields.Boolean()
```
sbidoul, line 30: Since this is an internal optimization of the report, I would make this a computed field instead of asking the caller to set it.

sbidoul:
another possible approach that I used in a custom report is to filter move lines like this:

```sql
WHERE full_reconcile_id IS NULL OR account_full_reconcile.create_date >= %(report_end_date_plus_1_day)s
```

sbidoul:
@JordiBForgeFlow @HviorForgeFlow could you check if #655 provides the expected performance improvement? It should optimize both cases (today as well as past date).

PR #632, opened 2019-11-13
[11.0][account_financial_report] fix unaffected earnings calculation

JordiBForgeFlow:
This PR fixes the calculation of the unaffected earnings so that it computes when the reporting period is not the start of the year.

cc @pedrobaeza


pedrobaeza:
```
======== Testing test_flake8 ========

account_financial_report/report/general_ledger.py:1561:89: E501 line too long (99 > 88 characters)
```

JordiBForgeFlow:
@pedrobaeza fixed all errors.

Issue #633, opened 2019-11-13
[12.0][partner_statement] Display the payment ref instead of 'Payment' string


kh-badep:
Hello,

When printing an activity statement it seems logical that the column Description is supposed to display additional information about the transaction, which can be:
- The invoice reference (account.move,name) if it's an invoice
- The transaction communication (account.payment,communication) if it's a payment, which can be the check number, etc.

However, that column seems to be filled with a generic 'Payment' string when it's a payment and upon inspecting code, it is forced here:
https://github.com/OCA/account-financial-reporting/blob/2953498afedcc60547aaaf851eda490e162bfef2/partner_statement/report/activity_statement.py#L66

Why not just use the columns `name` and `ref` as is instead of testing on the journal type. We can just drop the `CASE` statement so that the payment reference will be printed.

gdgellatly:
Yes this was a deliberate decision. Hence why there was a special case created. To be honest I don't really remember the reason why now, but it was necessary at the time.

@metamasterplay I remember why now, it all looks fine when you have a single payment paying a single invoice. But when you have a single payment paying hundreds of invoices it is a complete messs.

kh-badep:
Hello,
The only scenario I have found matching your description is when a user select multiple invoices from the tree view and then run the action `Register Payment` and then `Group Invoices`. And even then the user can choose to input a custom memo instead of the concatenated invoice numbers:
![image](https://user-images.githubusercontent.com/12763123/71580789-6ef99200-2b02-11ea-8938-efb5ba0e9c29.png)

Anyway since it's a stored string I guess it's a presentation issue rather than a performance one. Could we maybe truncate to a predefined character number in the report, or even use the `text-overflow` css property?

gdgellatly:
@metamasterplay It is nothing to do with the string choice, we use the same string for grouping purposes. Otherwise a single payment, paying 100 invoices printed as 100 lines when reconciled through the bank statement interface or the reconciliation widget (because the process of paying created loads of seperate move lines, each with their own ref). Or at least it used to, there's actually been quite a lot of fixes to the SQL of this module since then, plus upstream changes. 

[...]

PR #634, opened 2019-11-22
[12.0][partner_statement] Display the payment ref instead of 'Payment' string

kh-badep:
See https://github.com/OCA/account-financial-reporting/issues/633

### DIFF
--- a/partner_statement/report/activity_statement.py
+++ b/partner_statement/report/activity_statement.py
@@ -63,12 +63,7 @@ class ActivityStatement(models.AbstractModel):
                 THEN l.name
                 ELSE '/'
             END as name,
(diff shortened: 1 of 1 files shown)


gdgellatly (changes requested):
This is not the solution, it causes issues when a payment pays more than 1 invoice

gdgellatly:
ok I finally had a chance to look at this properly fixing other issues. I am fairly sure it is OK, I think in the past we might have used name instead of ref which are different. I'm still not a huge fan, I think `Payment` makes more sense and is more meaningful than `STMT:00596 (24/08/21)` to a customer. Too long ago, maybe that's why I wrote it that way in first place.  But I'll remove that block and let others decide. However this query, specifically this part of the query has changed in later versions.

PR #635, opened 2019-11-26
[FIX] account_chart_report: Add ondelete=cascade to wizard

yajo:
Wizards should never have `required=True` without `ondelete="cascade"`, or they forbid deletion of the referenced model.

@Tecnativa TT18838


pedrobaeza:
Fast-tracking as obvious patch.

PR #637, opened 2019-11-27
[12.0][IMP] partner_statement: Create test user without password to avoid crash tests with other modules like as password_security

sergio-teruel:
cc @Tecnativa TT19500
Review @chienandalu @carlosdauden


chienandalu (approved):
If CIs go well :+1:

pedrobaeza (approved):
I'm going to fast-track this as straight-forward.

pedrobaeza:
After fixing translations in other module in 58ada06586c804700cf806a124a24d404171fdb8

/ocabot merge patch

PR #638, opened 2019-11-28
[FIX] cash_flow: forecast accumulate next columns

jjscarafia:
For mis_builder to compute correctly bale we need accounts with user type include_initial_balance. We use receivable type which is the more accurate for this purpose


rvalyi:
I would not have find this trick easily, but yes it makes the module work as expected, both for in our out cashflow forecast lines as you can see in the following screenshot. So thanks a lot of the fix!

image

rvalyi (approved):

pedrobaeza (approved):

PR #639, opened 2019-11-28
[11.0][account_financial_report] refactor journal ledger

JordiBForgeFlow:
This PR completely refactors the engine behind the journal ledger report in order to stop using transient models, to avoid direct SQL queries, and to minimize the number of accesses to the database.

After this refactor the report should perform much faster, and should eliminate the known problems associated to vacuuming of the transient records.

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -8,7 +8,6 @@ from odoo import models, fields, api, _
 class GeneralLedgerReport(models.TransientModel):
     """ Here, we just define class fields.
     For methods, go more bottom at this file.
-
     The class hierarchy is :
     * GeneralLedgerReport
     ** GeneralLedgerReportAccount
@@ -699,7 +698,6 @@ AND
         """Return final subquery used to compute sum amounts on partners"""
 
         subquery_sum_amounts = """
-
             SELECT
                 sub.account_id AS account_id,
                 sub.partner_id AS partner_id,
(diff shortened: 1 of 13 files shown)


account_financial_report/report/general_ledger.py
```diff
@@ -1557,6 +1557,66 @@ def _compute_analytic_tags(self):
...
+                        sum(ml.balance) as balance
+                        FROM account_move_line as ml
+                        INNER JOIN account_move as am
+                        ON am.id = ml.move_id
+                        INNER JOIN account_journal j
+                        ON am.journal_id = j.id
+                """
+        if self.filter_cost_center_ids:
+            query_select_period_unaffected_earnings += """
+                        INNER JOIN account_analytic_account aa
```
pedrobaeza: I'm still seeing here SQL... this is why I say it would be good to have an SQL view with some calculations in its definition (like debit - credit) that you can query with `read_group`

account_financial_report/report/templates/journal_ledger.xml
```diff
@@ -155,11 +158,11 @@
                  name="taxes"/>
             <div class="act_as_cell amount"
                  name="debit">
-                <b><span t-field="journal.debit"/></b>
+                <b><span t-esc="journal['debit']"/></b>
```
pedrobaeza: This is losing the widget format. Use `t-widget` for restoring it.

JordiBForgeFlow: Fixed

account_financial_report/report/general_ledger.py
JordiBForgeFlow: This was not applicable to this PR. I have removed.

etobella (approved):
It is a good improvement to remove the Transients :smile: 
Also, the results between versions are the same, so it is good for me
:+1:

PR #642, opened 2019-12-09
[ADD]account_balance_reporting module

chusamo:
Adding an existing module.


pedrobaeza:
This one is already in https://github.com/OCA/l10n-spain/tree/8.0/account_balance_reporting, and it has been replaced by `mis_builder` in upper versions, so closing this as it doesn't even has commit history.

PR #643, opened 2019-12-12
[IMP] account_tax_balance: Optimized computation of move type

yajo:
Reduced complexity based on `read_group`. This method was prefetching a lot and doing many slow operations.

Executing in a real database for 197361 recs. took 14 minutes before and 1:36 minutes now.

@Tecnativa TT18838

### DIFF
--- a/account_tax_balance/models/account_move.py
+++ b/account_tax_balance/models/account_move.py
@@ -21,22 +21,39 @@ class AccountMove(models.Model):
     @api.multi
     @api.depends('line_ids.account_id.internal_type', 'line_ids.balance')
     def _compute_move_type(self):
-        def _balance_get(line_ids, internal_type):
-            return sum(line_ids.filtered(
-                lambda x: x.account_id.internal_type == internal_type).mapped(
-                    'balance'))
-
-        for move in self:
-            internal_types = move.line_ids.mapped('account_id.internal_type')
-            if 'liquidity' in internal_types:
-                move.move_type = 'liquidity'
-            elif 'payable' in internal_types:
-                balance = _balance_get(move.line_ids, 'payable')
-                move.move_type = (
-                    'payable' if balance < 0 else 'payable_refund')
-            elif 'receivable' in internal_types:
-                balance = _balance_get(move.line_ids, 'receivable')
-                move.move_type = (
-                    'receivable' if balance > 0 else 'receivable_refund')
-            else:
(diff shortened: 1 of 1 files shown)


yajo:
What happens with travis? :dizzy_face:

account_tax_balance/models/account_move.py
```diff
@@ -21,22 +21,32 @@ class AccountMove(models.Model):
...
+                                         else 'payable_refund')),
+            ("receivable", lambda balance: ('receivable' if balance > 0
+                                            else 'receivable_refund')),
+            (False, lambda balance: "other"),
+        )
+        move_ids = set(self.ids)
+        for internal_type, criteria in sequence:
+            # Find balances of the expected type for this move
+            domain = [
+                ("move_id", "in", list(move_ids)),
```
pedrobaeza: My only concern with this optimization is with this query: if you install initially the module in a DB with millions of records, the compute is done for the whole set AFAIK, so it will generate a query with millions of IDs. Can we plan an strategy that splits in chunks (without affecting logic, that I see it removes from the set elements).

account_tax_balance/models/account_move.py
```diff
@@ -21,22 +21,32 @@ class AccountMove(models.Model):
...
+            ("liquidity", lambda balance: "liquidity"),
+            ("payable", lambda balance: ('payable' if balance < 0
+                                         else 'payable_refund')),
+            ("receivable", lambda balance: ('receivable' if balance > 0
+                                            else 'receivable_refund')),
+            (False, lambda balance: "other"),
+        )
+        move_ids = set(self.ids)
+        for internal_type, criteria in sequence:
+            # Find balances of the expected type for this move
```
pedrobaeza:
```python
if not move_ids:
    break
```

pedrobaeza (commented):
I have fixed .travis.yml in main branch

yajo:
All done.

pedrobaeza (approved):

pedrobaeza:
If you rebase, we get a green CI for better display

yajo:
Rebased.

carlosdauden:
Fix tests

yajo:
Done, thanks.

carlosdauden (approved):

pedrobaeza:
I think this improvement is as well valid for 10.0, 11.0, 12.0...

PR #647, opened 2020-01-03
[ADD] Prefill move_type

StefanRijnhart:
Precreate move_type and fill with appropriate values to prevent
a MemoryError when the ORM attempts to call its compute method on a large
amount of preexisting moves.

### DIFF
--- a/account_tax_balance/hooks.py
+++ b/account_tax_balance/hooks.py
@@ -1,0 +1,44 @@
+# coding: utf-8
+# Copyright 2020 Opener B.V. <https://opener.amsterdam>
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+from psycopg2 import sql
+
+
+def pre_init_hook(cr):
+    """ Precreate move_type and fill with appropriate values to prevent
+    a MemoryError when the ORM attempts to call its compute method on a large
+    amount of preexisting moves. Note that the order of the mapping is
+    important as one move can have move lines on accounts of multiple types
+    and the move type is set in the order of precedence. """
+    logger = logging.getLogger(__name__)
+    logger.info('Add account_move.move_type column if it does not yet exist')
+    cr.execute(
+        "ALTER TABLE account_move ADD COLUMN IF NOT EXISTS move_type VARCHAR")
+    MAPPING = [
+        ('liquidity', 'liquidity', False),
+        ('payable', 'payable', 'AND aml.balance < 0'),
+        ('payable_refund', 'payable', 'AND aml.balance >= 0'),
+        ('receivable', 'receivable', 'AND aml.balance < 0'),
+        ('receivable_refund', 'receivable', 'AND aml.balance >= 0'),
+        ('other', False, False),
+    ]
+    for move_type, internal_type, extra_where in MAPPING:
+        args = [move_type]
+        query = sql.SQL("UPDATE account_move am SET move_type = %s")
+        if internal_type:
+            query += sql.SQL(
+                """FROM account_move_line aml
+                WHERE aml.account_id IN (
+                    SELECT id FROM account_account
+                    WHERE internal_type = %s)
+                AND aml.move_id = am.id AND am.move_type IS NULL
+                """)
+            args.append(internal_type)
+        else:
+            query += sql.SQL("WHERE am.move_type IS NULL")
+        if extra_where:
+            query += sql.SQL(extra_where)
+        cr.execute(query, tuple(args))
+        logger.info('%s move set to type %s', move_type, cr.rowcount)
--- a/account_tax_balance/__init__.py
+++ b/account_tax_balance/__init__.py
@@ -4,3 +4,4 @@
 
 from . import models
 from . import wizard
+from .hooks import pre_init_hook
--- a/account_tax_balance/__openerp__.py
+++ b/account_tax_balance/__openerp__.py
@@ -24,5 +24,6 @@
     ],
     "images": [
         'images/tax_balance.png',
-    ]
+    ],
+    "pre_init_hook": "pre_init_hook",
 }


pedrobaeza:
Isn't enough with #643?

StefanRijnhart:
No, I get the MemoryError with that code (Memory limit is set to 3GB, but I only have 800K moves). This code does the work in 24 seconds for that set.

Note that chunking (in the other change) does not prevent memory errors without invalidating the cache, which you cannot do in a compute method because it erases results from the earlier iterations.

pedrobaeza (commented):
I think you can do this with a for + some lists for the mapping instead of repeating similar queries.

StefanRijnhart:
I do think that is overengineering in this case.

pedrobaeza:
I don't think so. It simplifies a lot the reading. I can perform that change if you want.

StefanRijnhart:
Go ahead.

pedrobaeza:
I'm not able to push to the branch, but here you have the diff
```diff
diff --git account_tax_balance/hooks.py account_tax_balance/hooks.py
index 583d8c54..77956cb5 100644
--- account_tax_balance/hooks.py
+++ account_tax_balance/hooks.py
@@ -1,7 +1,9 @@
 # coding: utf-8
-# © 2020 Opener B.V. <https://opener.amsterdam>
+# Copyright 2020 Opener B.V. <https://opener.amsterdam>
+# Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import logging
+from psycopg2 import sql
 
 
 def pre_init_hook(cr):
@@ -12,63 +14,29 @@ def pre_init_hook(cr):
     logger.info('Add account_move.move_type column if it does not yet exist')
     cr.execute(
         "ALTER TABLE account_move ADD COLUMN IF NOT EXISTS move_type VARCHAR")
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'liquidity'
-        FROM account_move_line aml
-        WHERE aml.account_id IN (
-            SELECT id FROM account_account
-            WHERE internal_type = 'liquidity')
-        AND aml.move_id = am.id AND am.move_type IS NULL
-        """)
-    logger.info('%s move set to type liquidity', cr.rowcount)
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'payable'
-        FROM account_move_line aml
-        WHERE aml.account_id IN (
-            SELECT id FROM account_account
-            WHERE internal_type = 'payable')
-        AND aml.move_id = am.id AND am.move_type IS NULL
-        AND aml.balance < 0
-        """)
-    logger.info('%s move set to type payable', cr.rowcount)
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'payable_refund'
-        FROM account_move_line aml
-        WHERE aml.account_id IN (
-            SELECT id FROM account_account
-            WHERE internal_type = 'payable')
-        AND aml.move_id = am.id AND am.move_type IS NULL
-        AND aml.balance >= 0
-        """)
-    logger.info('%s move set to type payable_refund', cr.rowcount)
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'receivable'
-        FROM account_move_line aml
-        WHERE aml.account_id IN (
-            SELECT id FROM account_account
-            WHERE internal_type = 'receivable')
-        AND aml.move_id = am.id AND am.move_type IS NULL
-        AND aml.balance > 0
-        """)
-    logger.info('%s move set to type receivable', cr.rowcount)
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'receivable_refund'
-        FROM account_move_line aml
-        WHERE aml.account_id IN (
-            SELECT id FROM account_account
-            WHERE internal_type = 'receivable')
-        AND aml.move_id = am.id AND am.move_type IS NULL
-        AND aml.balance <= 0
-        """)
-    logger.info('%s move set to type receivable_refund', cr.rowcount)
-    cr.execute(
-        """
-        UPDATE account_move am SET move_type = 'other'
-        WHERE am.move_type IS NULL
-        """)
-    logger.info('%s move set to type other', cr.rowcount)
+    MAPPING = [
+        ('liquidity', 'liquidity', False),
+        ('payable', 'payable', 'AND aml.balance < 0'),
+        ('payable_refund', 'payable', 'AND aml.balance >= 0'),
+        ('receivable', 'receivable', 'AND aml.balance < 0'),
+        ('receivable_refund', 'receivable', 'AND aml.balance >= 0'),
+        ('other', False, False),
+    ]
+    for move_type, internal_type, extra_where in MAPPING:
+        args = [move_type]
+        query = sql.SQL("UPDATE account_move am SET move_type = %s")
+        if internal_type:
+            query += sql.SQL(
+                """FROM account_move_line aml
+                WHERE aml.account_id IN (
+                    SELECT id FROM account_account
+                    WHERE internal_type = %s)
+                AND aml.move_id = am.id AND am.move_type IS NULL
+                """)
+            args.append(internal_type)
+        else:
+            query += sql.SQL("WHERE am.move_type IS NULL")
+        if extra_where:
+            query += sql.SQL(extra_where)
+        cr.execute(query, tuple(args))
+        logger.info('%s move set to type %s' % move_type, cr.rowcount)
```

pedrobaeza:
@StefanRijnhart any news about my diff?

account_tax_balance/hooks.py
```diff
@@ -0,0 +1,74 @@
...
+        "ALTER TABLE account_move ADD COLUMN IF NOT EXISTS move_type VARCHAR")
+    cr.execute(
+        """
+        UPDATE account_move am SET move_type = 'liquidity'
+        FROM account_move_line aml
+        WHERE aml.account_id IN (
+            SELECT id FROM account_account
+            WHERE internal_type = 'liquidity')
+        AND aml.move_id = am.id AND am.move_type IS NULL
+        """)
```
gdgellatly: Well notwithstanding @pedrobaeza comment on a loop to do 5 queries basically the same as this, you could just do a single query with a CASE statement.

Something like
```sql
UPDATE account_move am SET move_type = CASE
  WHEN aa.internal_type = 'liquidity' THEN 'liquidity'
  WHEN aa.internal_type = 'payable' and aml.balance < 0 THEN 'payable'
  WHEN aa.internal_type = 'payable' and aml.balance >= 0 THEN 'payable_refund'
  WHEN aa.internal_type = 'receivable' and aml.balance > 0 THEN 'receivable'
  WHEN aa.internal_type = 'receivable' and aml.balance <= 0 THEN 'receivable_refund'
  ELSE 'other' 
END
FROM account_move_line aml, account_account aa
WHERE aa.id = aml.account_id 
      AND aml.move_id = am.id AND am.move_type IS NULL
```

Although you'd lose the rowcount info for logging, it is just one query and self explanatory.

StefanRijnhart: Alright, thanks both. I'll go with @gdgellatly's version then.

pedrobaeza (approved):

pedrobaeza:
This can be cherry-picked to all versions.

account_tax_balance/hooks.py
```diff
@@ -0,0 +1,33 @@
...
+            WHEN aa.internal_type = 'receivable' and aml.balance > 0
+                THEN 'receivable'
+            WHEN aa.internal_type = 'receivable' and aml.balance <= 0
+                THEN 'receivable_refund'
+            ELSE 'other'
+        END
+        FROM account_move_line aml, account_account aa
+        WHERE aa.id = aml.account_id
+            AND aml.move_id = am.id AND am.move_type IS NULL
+        """)
```
gdgellatly: @StefanRijnhart I am not sure this is right now I look. Because an account_move will have multiple lines of varying account types, we cannot do this, we need to evaluate the account types in order. With this query the behaviour is indeterminate.

gdgellatly (changes requested):

account_tax_balance/hooks.py
gdgellatly: You still could do it in one query, but it would require an array_agg grouping on account_type and an in clause.

pedrobaeza: You're totally right, then my diff is better IMO, and with logging.

StefanRijnhart: Thanks for noting. Replaced by Pedro's mapping

StefanRijnhart: Thanks for noting. Replaced by Pedro's mapping

account_tax_balance/hooks.py
```diff
@@ -0,0 +1,42 @@
...
+                    WHERE internal_type = %s)
+                AND aml.move_id = am.id AND am.move_type IS NULL
+                """)
+            args.append(internal_type)
+        else:
+            query += sql.SQL("WHERE am.move_type IS NULL")
+        if extra_where:
+            query += sql.SQL(extra_where)
+        cr.execute(query, tuple(args))
+        logger.info('%s move set to type %s' % move_type, cr.rowcount)
```
pedrobaeza: Sorry, my bad here:

```suggestion
        logger.info('%s move set to type %s' % (move_type, cr.rowcount))
```

StefanRijnhart:
Thanks, fix applied.

pedrobaeza:
Not sure why is Travis now failing...

StefanRijnhart:
@pedrobaeza restart the tests half a year later and it's green!

pedrobaeza:
Can you please squash commits as it's not done by the bot?

StefanRijnhart:
Ah yes, sorry. Squashed now.

PR #648, opened 2020-01-14
Display only posted entries

LGodoo:


gdgellatly:
@LGodoo you'll need to fix those line length issues before this will be accepted. Just put them on the next line down.

HviorForgeFlow:
@MiquelRForgeFlow

MiquelRForgeFlow (approved):
The commits should be squashed

PR #649, opened 2020-01-15
[FIX] account_tax_balance: improve test resilience

yajo:
This test failed in an integration environment where the search returned more items than necessary.

I modified it to just check that the tax that's created some lines above appears in the search results.

@Tecnativa TT21419

### DIFF
--- a/account_tax_balance/tests/test_account_tax_balance.py
+++ b/account_tax_balance/tests/test_account_tax_balance.py
@@ -108,8 +108,7 @@ class TestAccountTaxBalance(HttpCase):
 
         # exercise search has_moves = True
         taxes = self.env['account.tax'].search([('has_moves', '=', True)])
-        self.assertEqual(len(taxes), 1)
-        self.assertEqual(taxes[0].name, "Tax 10.0%")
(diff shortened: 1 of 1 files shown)


account_tax_balance/tests/test_account_tax_balance.py
```diff
@@ -108,8 +108,8 @@ def test_tax_balance(self):
 
         # exercise search has_moves = True
         taxes = self.env['account.tax'].search([('has_moves', '=', True)])
-        self.assertEqual(len(taxes), 1)
-        self.assertEqual(taxes[0].name, "Tax 10.0%")
+        self.assertLessEqual(tax, taxes)
+        self.assertEqual(tax.name, "Tax 10.0%")
```
pedrobaeza: This test is useless.

yajo: That's what I thought. Let me remove it.

pedrobaeza (approved):

pedrobaeza:
I think this can be fast-tracked:

/ocabot merge patch

PR #650, opened 2020-01-16
10.0-Trial-Balance-account.group-ordering-issue - Needs review

flotho:
Backport of the issue https://github.com/OCA/account-financial-reporting/pull/615 from v12.0

### DIFF
--- a/account_financial_report_qweb/report/trial_balance.py
+++ b/account_financial_report_qweb/report/trial_balance.py
@@ -399,7 +399,7 @@ SELECT
     accgroup.parent_id,
     coalesce(accgroup.code_prefix, accgroup.name),
     accgroup.name,
-    accgroup.parent_left * 100000,
+    accgroup.code_prefix,
     accgroup.level
 FROM
     account_group accgroup"""
--- a/account_financial_report_qweb/report/trial_balance.py
+++ b/account_financial_report_qweb/report/trial_balance.py
@@ -551,7 +551,7 @@ WHERE report_trial_balance_qweb_account.account_group_id = accgroup.id
         account account."""
         query_update_account_group = """
 UPDATE report_trial_balance_qweb_account
-SET sequence = CONCAT(newline.sequence, newline.code),,
+SET sequence = CONCAT(newline.sequence, newline.code),
     level = newline.level + 1
 FROM report_trial_balance_qweb_account as newline
 WHERE newline.account_group_id = report_trial_balance_qweb_account.parent_id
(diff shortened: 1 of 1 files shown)


flotho:
@JordiBForgeFlow , please add need review tag

flotho:
Hi @pedrobaeza ,

Who am I suppose to ping here ? Jordi Balester ?

pedrobaeza:
Check who created that code.

flotho:
Hi @gaikaz , @amcor ,

Any chance to be reviewed ?

Regards

account_financial_report_qweb/report/trial_balance.py
```diff
@@ -75,7 +75,7 @@ class TrialBalanceReportAccount(models.TransientModel):
     )
     hide_line = fields.Boolean(compute='_compute_hide_line')
     # Data fields, used to keep link with real object
-    sequence = fields.Integer(index=True, default=1)
+    sequence = fields.Char(index=True, )
```
gaikaz: Redundant comma. Also, in v12 it has `default='1'` here. Personally can't see when would sequence not get set, but maybe still leave it?
```suggestion
    sequence = fields.Char(index=True)
```

gaikaz (changes requested):
The changes look ok in general apart from that one comma.
I could also do my little test case described in the original issue: https://github.com/OCA/account-financial-reporting/issues/609#issue-486261606 
But only later today, worst case - tomorrow.

gaikaz:
@flotho I've gone out of my way and during free time, I got Odoo v10 running and, like promised in my PR, tried to reproduce the inconsistent ordering issue the way I could reproduce it on v12.
I'm sorry to say, I couldn't :neutral_face: 
I used this module before the changes of this PR. If I would have reproduced it, then I would have tried it with the changes to see if they actually fix it. 

Could you maybe provide a step by step reproduction guide for a fresh install of Odoo v10 with/without demo data?

flotho:
Hi @gaikaz ,

Create groups in a reverse order for example :
* 700 : Revenues
* 600 : charges
* 500 : Bank
* ../...

So that before my PR the datas will be displayed in the created order and with my PR they will be ordered following group names

flotho:
ping @gaikaz , @JordiBForgeFlow

gaikaz (approved):
Sorry for the delay, @flotho 
I did a functional test now with your described scenario. It indeed is a problem and this PR fixes it.

PR #653, opened 2020-01-31
[12.0][account_financial_report] Improve General Ledger wizard

ivantodorovich:
This PR improves usability a bit in the General Ledger report wizard.

Instead of using a hardcoded `receivable/payable_accounts_only`, we're using `account.account.types` to offer a wider selection.

Also I took the oportunity to improve a bit the "settings" part of the view.

**Before:**
image

**After:**
image

### DIFF
--- a/account_financial_report/wizard/general_ledger_wizard.py
+++ b/account_financial_report/wizard/general_ledger_wizard.py
@@ -4,6 +4,7 @@
 # Copyright 2016 Camptocamp SA
 # Copyright 2017 Akretion - Alexis de Lattre
 # Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# Copyright 2020 Druidoo
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 
@@ -57,8 +58,10 @@ class GeneralLedgerReportWizard(models.TransientModel):
     show_analytic_tags = fields.Boolean(
         string='Show analytic tags',
     )
-    receivable_accounts_only = fields.Boolean()
-    payable_accounts_only = fields.Boolean()
+    account_type_ids = fields.Many2many(
+        'account.account.type',
+        string='Account Types',
+    )
     partner_ids = fields.Many2many(
         comodel_name='res.partner',
         string='Filter partners',
@@ -133,8 +136,8 @@ class GeneralLedgerReportWizard(models.TransientModel):
                 lambda p: p.company_id == self.company_id or
                 not p.company_id)
         if self.company_id and self.account_ids:
-            if self.receivable_accounts_only or self.payable_accounts_only:
-                self.onchange_type_accounts_only()
+            if self.account_type_ids:
+                self._onchange_account_type_ids()
             else:
                 self.account_ids = self.account_ids.filtered(
                     lambda a: a.company_id == self.company_id)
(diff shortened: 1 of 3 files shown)


account_financial_report/wizard/general_ledger_wizard.py
```diff
@@ -180,28 +183,24 @@ def _check_company_id_date_range_id(self):
...
-            elif self.receivable_accounts_only:
-                domain += [('internal_type', '=', 'receivable')]
-            elif self.payable_accounts_only:
-                domain += [('internal_type', '=', 'payable')]
-            self.account_ids = self.env['account.account'].search(domain)
+    @api.onchange('account_type_ids')
+    def _onchange_account_type_ids(self):
+        if self.account_type_ids:
+            self.account_ids = self.env['account.account'].search([
+                ('user_type_id', 'in', self.account_type_ids.ids)])
```
i-vyshnevska, line 191: to add a domain for the company as by default current user company will be used and this is not always correct

account_financial_report/wizard/general_ledger_wizard.py
```diff
@@ -180,28 +183,24 @@ def _check_company_id_date_range_id(self):
...
         else:
             self.account_ids = None
 
     @api.onchange('partner_ids')
     def onchange_partner_ids(self):
         """Handle partners change."""
         if self.partner_ids:
-            self.receivable_accounts_only = self.payable_accounts_only = True
+            self.account_type_ids = self.env['account.account.type'].search([
+                ('type', 'in', ['receivable', 'payable'])])
```
i-vyshnevska, line 200: the same remark about the company

account_financial_report/wizard/general_ledger_wizard_view.xml
```diff
@@ -91,10 +107,6 @@
                 res_model="general.ledger.report.wizard"
                 src_model="res.partner"
                 view_mode="form"
-                context="{
-                    'default_receivable_accounts_only':1,
-                    'default_payable_accounts_only':1,
-                    }"
```
i-vyshnevska: it is a pity to lost this context, what about add a hook to create two tags for account receivable and payable and set them as default

account_financial_report/wizard/general_ledger_wizard.py
ivantodorovich: Thanks!!!
It's fixed on https://github.com/OCA/account-financial-reporting/pull/653/commits/131f9dc162afae7f08a5d87253a29ba92e151e95

account_financial_report/wizard/general_ledger_wizard.py
ivantodorovich: Account Types have no company. In this case it's not necessary

account_financial_report/wizard/general_ledger_wizard_view.xml
```diff
@@ -91,10 +107,6 @@
                 res_model="general.ledger.report.wizard"
                 src_model="res.partner"
                 view_mode="form"
-                context="{
-                    'default_receivable_accounts_only':1,
-                    'default_payable_accounts_only':1,
```
i-vyshnevska, line 96: suggestion: you can set Payable, Receivable tags by default so it will be initialized in the same way as before

i-vyshnevska (approved):

account_financial_report/wizard/general_ledger_wizard_view.xml
ivantodorovich: It is already working like this, thanks to the _onchange_partner_id :)

ivantodorovich:
ping @alexis-via @simahawk

mamcode (approved):
LGTM.

simahawk:
@OCA/accounting-maintainers anyone?

JordiBForgeFlow:
@ivantodorovich @simahawk In my opinion we should stop making improvements to v12. We have completely refactored the report in  v12 here https://github.com/OCA/account-financial-reporting/pull/660

Anyone looking to improve the report shall consider using that PR, that is already merged in 13.0,  and proposing changes there instead to stable 12.9

Cc @pedrobaeza

pedrobaeza:
Yeah, I agree, as previous state is totally unmaintainable. Merging the refactor in v12 is risky, but v13 is the one to go. We can agree to merge it also in v12 with enough testing if you prefer.

JordiBForgeFlow:
@pedrobaeza the report contains tests and has been in use at two large companies (millions of move lines for the last month). The performance is outstanding.

ivantodorovich:
Well.. this PR (now merged) only contained a very small UX improvement to the wizard, not the actual report.

I didn't check but it's probably compatible with https://github.com/OCA/account-financial-reporting/pull/660

PR #655, opened 2020-02-06
[10.0][IMP] account_financial_report_qweb: Improve performances

rousseldenis:
We rely on full reconciliation and remove the HAVING clause
which could be bulky as operating after the SELECT.

### DIFF
--- a/account_financial_report_qweb/report/open_items.py
+++ b/account_financial_report_qweb/report/open_items.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # © 2016 Julien Coux (Camptocamp)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from datetime import timedelta
 
 from odoo import models, fields, api, _
 
@@ -415,10 +416,6 @@ FROM
                 account_partial_reconcile pr
                     ON ml.balance < 0 AND pr.credit_move_id = ml.id
             LEFT JOIN
-                account_move_line ml_future
-                    ON ml.balance < 0 AND pr.debit_move_id = ml_future.id
-                    AND ml_future.date > %s
-            LEFT JOIN
                 account_move_line ml_past
                     ON ml.balance < 0 AND pr.debit_move_id = ml_past.id
                     AND ml_past.date <= %s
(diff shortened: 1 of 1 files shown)


rousseldenis:
WIP

rousseldenis:
@sbidoul @HviorForgeFlow @JordiBForgeFlow @jcoux 

We experienced problems with aged balance especially with great number of unreconciled items.

account_financial_report_qweb/report/open_items.py
```diff
@@ -428,27 +425,20 @@ def _get_line_sub_query_move_lines(self,
...
         """
         sub_query += """
+            LEFT JOIN account_full_reconcile afr ON afr.id = ml.full_reconcile_id
             WHERE
                 ra.report_id = %s
+            AND ml.full_reconcile_id IS NULL OR afr.create_date >= %s
             GROUP BY
                 ml.id,
                 ml.balance,
                 ml.amount_currency
```
sbidoul, line 441: IMO, this GROUP BY can be removed. GROUP BY id does not make much sense to me, and the id should be unique across the two parts of the union.

account_financial_report_qweb/report/open_items.py
```diff
@@ -617,17 +607,19 @@ def _inject_line_values(self, only_empty_partner_line=False):
...
-             self.date_at,
+             full_reconcile_date,
              self.date_at,
              self.id,
+             full_reconcile_date,
              self.env.uid,
              self.id,
-             self.date_at,)
+             self.date_at,
+             )
```
sbidoul, line 623: Just a side note (no need to change): in such  big queries, using `%(name)s` and passing a dict to `execute` is much easier to comprehend and maintain.

account_financial_report_qweb/report/open_items.py
sbidoul: Hm, no I retract this. There are these LEFT JOIN that repeat the id.

account_financial_report_qweb/report/open_items.py
```diff
@@ -428,27 +425,20 @@ def _get_line_sub_query_move_lines(self,
...
             LEFT JOIN
                 account_move_line ml_past
                     ON ml.balance > 0 AND pr.credit_move_id = ml_past.id
                     AND ml_past.date <= %s
         """
         sub_query += """
+            LEFT JOIN account_full_reconcile afr ON afr.id = ml.full_reconcile_id
             WHERE
                 ra.report_id = %s
+            AND ml.full_reconcile_id IS NULL OR afr.create_date >= %s
```
sbidoul, line 437: nit: indent this 4 chars

sbidoul (approved):
Looking for ml that are either not fully reconciled, or have been fully reconciled after the pivot date is apparently faster way to reduce the search space. The amounts will be the same as these still consider the move lines before or on the pivot date.

HviorForgeFlow (approved):
Performance is quite better! Thanks

AaronHForgeFlow (approved):
:+1: Much faster in actual database and same result!

account_financial_report_qweb/report/open_items.py
AaronHForgeFlow: I think this is the only thing pending

sbidoul: It's a nit, non blocking for me :)

PR #660, opened 2020-02-21
[12.0] account_financial_report: refactor [DO NOT MERGE]

JoanSForgeFlow:
This PR introduces a complete refactoring of the account financial reports.

Improvements:
- No use of Transient models - using transient models resulted in the auto vacuum process for transient models process often collapsing because the AFR introduced a huge number of records every time the report was run.
- No use of direct SQL - means that the code is more readable, and can be more easily extended - we make use of ORM.
- Faster execution - We experimented with huge databases. Trial balance execution time reduced from 3 minutes to 15 secs.

This is the status of the various reports.
* [x] General Ledger - Needs review
* [x] Journal Ledger - Needs review
* [x] Open Items - Needs review
* [x] Trial Balance - Needs review
* [x] Aged Partner Balance - Needs review
* [x]  VAT report - Needs review


HaraldPanten (approved):
Tested in local environment. Functional review OK 👍

pedrobaeza:
Why are you adding natsort and pandas libraries? They are very heavy dependencies, and I don't think they are really needed for this.

JordiBForgeFlow:
@pedrobaeza the pandas library speeds up a lot the process of managing the dictionaries.
The natsort library is very useful to sort the account groups, because it just applies the natural sorting.

pedrobaeza:
I don't agree on adding such hard dependencies, and more only for "syntactic sugar", and about the natural sorting, I don't get the problem, but for sure that will mean less performance, which is the contrary at what we are looking for.

And not only that, introducing these libraries will mean that you break all the current installations that have this module installed. It's not the desirable thing but OK if it's totally justified, but in this case I don't consider this addition as essential.

rvalyi:
may be the code could be made compatible with an import try and a fallback to the classical types when the natsort and pandas libs aren't found?

JoanSForgeFlow:
I have removed pandas and natsort dependencies. Performance has been a little bit affected, not critical at all, that addition wasn't essential. Thanks for your review!

pedrobaeza:
Great, thanks.

JordiBForgeFlow:
@fclementic2c @sbidoul @rousseldenis @Cedric-Pigeon what do you think about this? It's really worth reviewing!

gva-acsone:
Hello,

Thanks for this interesting PR.
I tested (functional test) the reports on one of our databases and compared the result to the previous OCA reports.

My observations:
- I have a time out for 3 of the 6 reports. I can't open or export them anymore for a database with a large amount of journal entries. The reports that are concerned are : 

1. Aged partner balance
2. Journal ledger
3. Open Items

The loading of those 3 reports worked well with the previous version of the OCA financial reports (3 minutes loading time on average), but I now have a time out with the new version of the reports.

Did the reports work for you on a database with a large amount of data ? Am I the only one with this problem? 

- When you export the trial balance to XLS, the columns 'period balance' and 'ending balance' are not completed. However the problem doesn't appear when you export the report to PDF or open it in Odoo directly. It would be nice to also complete those two columns in the XLS export.

- The general ledger report is much faster than before, thanks! 
Small observation: when I check the 'activate centralization option', I find centralized entries in the report for accounts for which the 'Centralized' checkbox on the account is set on FALSE.  

- The numbers presented in the new VAT report are very different then the figures I have in the old report. When i compare the numbers with the journal entries (group by tax), the numbers from the new report don't seem very correct. I will try to investigate in a near future why those numbers are different.

Cheers

@sbidoul

account_financial_report/report/templates/trial_balance.xml
```diff
@@ -34,50 +36,65 @@
...
+                            <t t-if="balance['type'] == 'account_type'">
+                                <t t-call="account_financial_report.report_trial_balance_line"/>
+                            </t>
                         </t>
-                        <t t-if="line.account_group_id">
-                            <t t-set="style" t-value="style + 'font-weight: bold; color: blue;'"/>
+<!--                        <t t-set="padding" t-value="line.level * 4"/>-->
+<!--                        <t t-if="hierarchy_on != 'none'">-->
+<!--                            <t t-set="style" t-value="'font-size: ' + str(14 - line.level) + 'px; margin-left: ' + str(line.level * 4) + 'px;'"/>-->
+<!--                        </t>-->
```
gaikaz, line 57: Lot's of commented out XML code like this. And because v13 migration was based on this PR, v13 is now kind of messy.
The code commented here is also one of the better UX features of this module.
It makes the Trial Balance look like this (before PR):
image
Instead of like this (taken from v13):
image

Also in v13 clicking on different parts of the table (like group code, account name or even the sums of money) doesn't open related data. This is also one of the best "selling points" of the module.
Though, I haven't checked if this is the case with this PR.

eLBati:
Hi @JoanSForgeFlow please consider that `account_financial_report` is a dependency of some modules at https://github.com/OCA/l10n-italy and other repositories I guess.

Those dependent modules, with a simple update of `account_financial_report` after this PR, will stop working, if not refactored.

I see 2 options:
 - Don't merge this before the dependent modules have been modified
 - Do these changes for the new release, v14, allowing the dependent modules to adapt in time.


Thanks!

pedrobaeza:
FYI, the refactoring has been included in v13

JoanSForgeFlow:
Hi @elbati, 
This PR is not going to be merged in v12 because of all the dependent modules, so these dependent modules don't need to be modified in v12.

pedrobaeza:
@JoanSForgeFlow if you fix some of the commented issues, please tell for forward-porting them to v13

JoanSForgeFlow:
@pedrobaeza yes, I am forward-porting the fixed issues and improvements by making PR to v13.
I try to backport other improvements done in v13 to this PR too.
Thank you for your reminder!

JordiBForgeFlow:
@JoanSForgeFlow I have fixed an issue with foreign currency https://github.com/OCA/account-financial-reporting/pull/660/commits/d5cac4b67403a47b5d2b49ccd187dbc3c0d8806e Can you review?

account_financial_report/reports.xml
```diff
@@ -5,112 +5,122 @@
...
-      id="action_report_general_ledger_html"
-      model="report_general_ledger"
+      id="action_print_report_general_ledger_html"
+      model="general.ledger.report.wizard"
       string="General Ledger"
       report_type="qweb-html"
-      name="account_financial_report.report_general_ledger_qweb"
-      file="account_financial_report.report_general_ledger_html"
+      menu="False"
+      name="account_financial_report.general_ledger"
```
gaikaz, line 22: **The `name` attribute can not be the same between report records.**

When you print a PDF report, a JS action is triggered that later calls back to Odoo with only report_name as parameter to the controller. Reference here:
https://github.com/odoo/odoo/blob/85fe44a7298ef9883160359814e74e39b7e10873/addons/web/controllers/main.py#L1588-L1600

So if `_get_report_from_name` in that piece of code gets the "html" report record instead (which doesn't have paperformat set), a default paperformat is used.

This has already been reported on #558 yet wasn't fixed.

gaikaz (changes requested):
Sorry to bring more issues to the table, but this really should be fixed here, as you are already changing the report names anyway.

account_financial_report/reports.xml
JordiBForgeFlow: Can you send a PR to this branch?

gaikaz: Do you mean on the forked repo?
I probably could, but not this week, I guess.
Simply adding `_html` to the HTML report name would work and be quick to do, but "Coder OCD :smile:" doesn't allow me to leave the PDF one without `_pdf` at the end, which then requires some more changes elsewhere.

AaronHForgeFlow:
Conflicts here after this merge https://github.com/OCA/account-financial-reporting/pull/653 I will check.

tarteo (changes requested):
Javascript doens't work anymore. I can't drill down on monetary values anymore. Clicking on it should show the account.move.line. https://github.com/OCA/account-financial-reporting/issues/690 is the same issue because this version was used for the migration to 13.0

JordiBForgeFlow:
@JoanSForgeFlow the base branch has some changes and now we have some conflicting files. Can you rebase and resolve the conflicts?

account_financial_report/wizard/trial_balance_wizard.py
```diff
@@ -183,25 +202,45 @@ def onchange_show_partner_details(self):
...
-        model = self.env['report_trial_balance']
-        report = model.create(self._prepare_report_trial_balance())
-        report.compute_data_for_report()
-
-        context1['active_id'] = report.id
-        context1['active_ids'] = report.ids
-        vals['context'] = context1
-        return vals
+        report_type = 'qweb-html'
+        return self._export(report_type)
```
thomaspaulb, line 243: By this change, as is already in the 13.0 version of this module, the report will be shown as HTML in an `iframe` instead of being a direct widget in the screen.

For 13.0, I noticed that HTML does not include the `js` files anymore, so that clicking the values in the report has stopped working. There is an issue about it [here](https://github.com/OCA/account-financial-reporting/issues/690).

I wonder whether including those JS files helps, because they would still be restricted to the iFrame.

Have you thought of a solution for this?

account_financial_report/view/report_template.xml
```diff
@@ -6,7 +6,7 @@
         <xpath expr="." position="inside">
             <link href="/account_financial_report/static/src/css/report.css" rel="stylesheet"/>
             <script type="text/javascript"
-                    src="/account_financial_report/static/src/js/account_financial_report_backend.js"/>        
+                    src="/account_financial_report/static/src/js/account_financial_report_backend.js"/>
             <script type="text/javascript"
                     src="/account_financial_report/static/src/js/account_financial_report_widgets.js"/>
```
thomaspaulb, line 11: This is the JS file I mean in the previous comment, it contains the listeners for the click actions.

StefanRijnhart:
@pedrobaeza I'm taking the liberty to reopen this PR. We experience performance issues that we think this PR solves, and it is on our backlog to review and test the change. After all, it is included in 13 and further.

pedrobaeza:
But this is not going to be merged on 12.0. It takes too much changes for this stable version, and behavior changes, so being opened is not changing anything.

rafaelbn:
What should be done with this huge work? Do you saw it @SirTakobi @eLBati ?

StefanRijnhart:
Looks like our performance problems are solved by https://github.com/OCA/account-financial-reporting/pull/875, so for the moment we are not invested in this PR anymore.

eLBati:
@rafaelbn yes we know this and that will not be merged, so we make specific PRs when needed

Issue #661, opened 2020-02-26
[13.0] Account tax balance not working correctly in multicompany


CasVissers-360ERP:
https://drive.google.com/file/d/19Y-gqaw2XEFHdXFc9Lko30rtsoYH3vxa/view

Two issues:
1. the company is not loaded correctly in the wizard
2. if company != main company the tax balances are not loaded

PR #662, opened 2020-02-26
[13.0][FIX] account_tax_balance: select company in multicompany

astirpe:
Fixes https://github.com/OCA/account-financial-reporting/issues/661

@CasVissers

### DIFF
--- a/account_tax_balance/models/account_tax.py
+++ b/account_tax_balance/models/account_tax.py
@@ -26,10 +26,11 @@ class AccountTax(models.Model):
 
     def get_context_values(self):
         context = self.env.context
+        actual_company_id = context.get("company_id", self.env.company.id)
         return (
             context.get("from_date", fields.Date.context_today(self)),
             context.get("to_date", fields.Date.context_today(self)),
-            context.get("company_ids", [self.env.user.company_id.id]),
+            context.get("company_ids", [actual_company_id]),
             context.get("target_move", "posted"),
         )
 
--- a/account_tax_balance/wizard/open_tax_balances.py
+++ b/account_tax_balance/wizard/open_tax_balances.py
@@ -8,8 +8,12 @@ class WizardOpenTaxBalances(models.TransientModel):
     _name = "wizard.open.tax.balances"
     _description = "Wizard Open Tax Balances"
 
-    company_id = fields.Many2one(
-        "res.company", required=True, default=lambda self: self.env.user.company_id
+    company_ids = fields.Many2many(
+        comodel_name="res.company",
+        string="Companies",
+        required=True,
+        domain=lambda self: [("id", "in", self.env.companies.ids)],
+        default=lambda self: self.env.companies.ids,
     )
     from_date = fields.Date(
         required=True, store=True, readonly=False, compute="_compute_date_range"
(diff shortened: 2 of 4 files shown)


account_tax_balance/wizard/open_tax_balances.py
```diff
@@ -9,7 +9,7 @@ class WizardOpenTaxBalances(models.TransientModel):
     _description = "Wizard Open Tax Balances"
 
     company_id = fields.Many2one(
```
pedrobaeza: Shouldn't this be now a m2m?

CasVissers-360ERP (approved):
Fixes the issue! Thnx!

account_tax_balance/wizard/open_tax_balances.py
astirpe: Yes could be. In this case is not only a fix but also an implementation of a new feature.

pedrobaeza: Well, it's the multi-company implementation of v13 in fact.

astirpe: Done

account_tax_balance/models/account_tax.py
```diff
@@ -26,10 +26,11 @@ class AccountTax(models.Model):
 
     def get_context_values(self):
         context = self.env.context
+        actual_company_id = context.get("company_id", self.env.company.id)
```
pedrobaeza, line 29: Not sure about the specific code, but I think you will get the idea:

```suggestion
        actual_company_ids = context.get("company_ids", self.env.companies.ids)
```

account_tax_balance/wizard/open_tax_balances.py
```diff
@@ -8,8 +8,11 @@ class WizardOpenTaxBalances(models.TransientModel):
     _name = "wizard.open.tax.balances"
     _description = "Wizard Open Tax Balances"
 
-    company_id = fields.Many2one(
-        "res.company", required=True, default=lambda self: self.env.user.company_id
+    company_ids = fields.Many2many(
+        comodel_name="res.company",
+        string="Companies",
+        required=True,
+        default=lambda self: self.env.company.ids,
```
pedrobaeza:
```suggestion
        default=lambda self: self.env.companies.ids,
```
not sure if the variable of allowed companies is that one.

account_tax_balance/models/account_tax.py
astirpe: `actual_company_id` is for retro-compatibility, in case someone is still using `company_id` in context

pedrobaeza: Then:
```suggestion
        actual_company_id = context.get("company_id", False)
        ...
            context.get("company_ids", actual_company_id and [actual_company_id] or self.env.companies.ids),
```

CasVissers-360ERP:
Currently the wizards loads the companies activated by the user, but it also allows selecting companies not activated by the user. Which shouldn't be the case since the underlaying data is not accessible anyhow.

pedrobaeza:
Yeah, a domain over `company_ids` field with `self.env.companies.ids` restriction.

CasVissers-360ERP (approved):
👍

account_tax_balance/models/account_tax.py
astirpe: Why?

pedrobaeza: For having the same default that you have put in the field.

astirpe: I don't understand. With this line I'm keeping the same behavior as it was before, in case `company_ids` and `company_id` are not passed in the context.

pedrobaeza: OK, maybe I'm not understanding the use of this, so let's move on!

PR #670, opened 2020-04-07
[13.0][FIX] account_financial_report: general_ledger

JoanSForgeFlow:
Some fixes for General Ledger and Trial Balance reports.

General Ledger:
* Ordering move_lines of each account by date
* Centralization is only calculated on accounts where 'centralized' option is True
* Cumul. Balance of each move_line fixed. 

Fix trial_balance_xlsx when not show_partner_details

@JordiBForgeFlow @ForgeFlow

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -25,6 +25,7 @@ class GeneralLedgerReport(models.AbstractModel):
                         "group_id": account.group_id.id,
                         "currency_id": account.currency_id or False,
                         "currency_name": account.currency_id.name,
+                        "centralized": account.centralized,
                     }
                 }
             )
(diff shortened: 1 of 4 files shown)


ernestotejeda:
Why when the account is centralized, the partner is not shown for each row? is that right?

ernestotejeda (commented):
Why when the account is centralized, the partner is not shown for each row? is that right?

JoanSForgeFlow:
Thank you for your review!

When the account is centralized, amounts are grouped by period. A row shows an amount that corresponds to different partners. Because of this partner is not shown and it is shown 'centralized entries'. It seems to me this is the same behavior of current OCA v12 reports.

ernestotejeda (approved):
Thanks for the clarification :+1:

pedrobaeza (approved):

Issue #669, opened 2020-03-28
[13.0] General Ledger Report crash


theopensourcerer:
```
Traceback (most recent call last):
  File "/opt/odoo13/odoo/odoo/addons/base/models/ir_http.py", line 229, in _dispatch
    result = request.dispatch()
  File "/opt/odoo13/odoo/odoo/http.py", line 804, in dispatch
    r = self._call_function(**self.params)
  File "/opt/odoo13/odoo/odoo/http.py", line 345, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo13/odoo/odoo/service/model.py", line 93, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo13/odoo/odoo/http.py", line 338, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo13/odoo/odoo/http.py", line 910, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo13/odoo/odoo/http.py", line 510, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo13/odoo/custom-addons/reporting-engine/report_xlsx/controllers/main.py", line 49, in report_routes
    reportname, docids, converter, **data
  File "/opt/odoo13/odoo/odoo/http.py", line 510, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo13/odoo/addons/web/controllers/main.py", line 1956, in report_routes
    html = report.with_context(context).render_qweb_html(docids, data=data)[0]
  File "/opt/odoo13/odoo/odoo/addons/base/models/ir_actions_report.py", line 757, in render_qweb_html
    data = self._get_rendering_context(docids, data)
  File "/opt/odoo13/odoo/odoo/addons/base/models/ir_actions_report.py", line 774, in _get_rendering_context
    data.update(report_model._get_report_values(docids, data=data))
  File "/opt/odoo13/odoo/custom-addons/account-financial-reporting/account_financial_report/report/general_ledger.py", line 657, in _get_report_values
    cost_center_ids,
  File "/opt/odoo13/odoo/custom-addons/account-financial-reporting/account_financial_report/report/general_ledger.py", line 451, in _get_period_ml_data
    acc_id = move_line["account_id"][0]
TypeError: 'bool' object is not subscriptable
```

[...]

PR #673, opened 2020-04-09
[FIX] account_financial_report: Fix error when generating reports

ernestotejeda:
Some reports crash if account move line implicated are Sections lines or Notes lines in an invoice

Issue: https://github.com/OCA/account-financial-reporting/issues/669

Cc @Tecnativa TT19684

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -327,7 +327,11 @@ class GeneralLedgerReport(models.AbstractModel):
         analytic_tag_ids,
         cost_center_ids,
     ):
-        domain = [("date", ">=", date_from), ("date", "<=", date_to)]
+        domain = [
+            ("display_type", "=", False),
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@theopensourcerer please review

pedrobaeza (approved):

pedrobaeza:
@JoanSForgeFlow I think this affects also to your refactoring in v12

JoanSForgeFlow (approved):
Yes! Thank you! Now I add this commit to the refactoring in v12!

JoanSForgeFlow:
After checking v12, I have seen `display_type` is an invoice.move.line field and it's checked before creating the `account.move.line`.  Invoice notes and sections don't seem to create problems in v12 reports.

theopensourcerer:
Thanks all. Just updated the modules and tried the reports again. All work without crashing :-)

PR #676, opened 2020-04-23
[13.0][MIG] mis_builder_cash_flow

MiquelRForgeFlow:
Standard migration.


MiquelRForgeFlow:
@jjscarafia @etobella @LoisRForgeFlow ready to review

AdriaGForgeFlow (approved):
Code review :+1:

HaraldPanten (approved):
I tested several use cases and I didn't find any issue. Functional review OK. 👍

mis_builder_cash_flow/report/mis_cash_flow.py
```diff
@@ -77,15 +75,16 @@ def init(self):
                 aml.full_reconcile_id as full_reconcile_id,
                 aml.partner_id as partner_id,
                 aml.company_id as company_id,
-                aml.user_type_id as user_type_id,
+                aa.user_type_id as user_type_id,
```
jjscarafia: Perhaps this change can add a little of overhead because of an additional join?
Now that we have `account_internal_type` on aml. perhaps we can add this field instead of `user_type_id` and then on all the expressions (like [this one](https://github.com/OCA/account-financial-reporting/blob/4d15be2e1e592e978cee55c713a663ebf3d8ab15/mis_builder_cash_flow/data/mis_report.xml#L39)) use just `account_internal_type` instead of `account_id.internal_type` (what could improve performance).

What do you think?

mis_builder_cash_flow/report/mis_cash_flow.py
```diff
@@ -97,23 +96,22 @@ def init(self):
...
                 fl.partner_id as partner_id,
                 fl.company_id as company_id,
-                %i as user_type_id,
+                {} as user_type_id,
                 fl.name as name,
                 fl.date as date
             FROM mis_cash_flow_forecast_line as fl
-        """
-            % account_type_receivable.id
+        """.format(
```
jjscarafia: Just to know, why this change? I used to believe that in odoo the "standard" is to use % instead of format. Perhaps I was wrong and that standard is only for translatable strings.

jjscarafia:
thanks @MiquelRForgeFlow . I've send some comments.

mis_builder_cash_flow/report/mis_cash_flow.py
MiquelRForgeFlow: First, there was a `%d`. the documentation in https://www.psycopg.org/docs/usage.html says:

```
The variables placeholder must always be a %s, even if a different placeholder (such as a %d for integers or %f for floats) may look more appropriate:
```

Second, about using `%` or `format`. Both things are not recommended to be used because they may lead to sql-injection problems. That's why pylint has a sql-injection checker. Just happens, in this case, there is no problem of sql-injection, that's why there is no problem if we use `%` or `format`. I just preferred to use `format` as a saveproof, to add `NULL` in case there is no account_type_receivable.

pedrobaeza: Please use `sql.SQL`

mis_builder_cash_flow/report/mis_cash_flow.py
MiquelRForgeFlow: I think there is no overhead. But I will try what you said.

pedrobaeza: The overhead is already there if you have `account_account aa` in the `JOIN` part. If you can avoid it because there's no need of any of its fields, then you boost up performance.

MiquelRForgeFlow: Well, I just removed this field in the query, as it was not used anywhere in the report. Now, there is an error in account_financial_report/tests/test_vat_report.py, but seems unrelated, right? @HaraldPanten could you test again?

HaraldPanten:
Just to know, I have a question. Do you think that `account` dependency should be added in `"depends"` parameters?

jjscarafia:
> Just to know, I have a question. Do you think that `account` dependency should be added in `"depends"` parameters?
@HaraldPanten 
I believe it's not necessary because the [mis_builder module already has it as a dependency](https://github.com/OCA/mis-builder/blob/10.0/mis_builder/__manifest__.py#L15)

mis_builder_cash_flow/report/mis_cash_flow.py
jjscarafia: Great, thanks @MiquelRForgeFlow !
I still believe that adding `account_internal_type` and replacing `account_id.internal_type` for `account_internal_type` could help in performance but that could be achieved on a new PR.

Thanks!

HaraldPanten:
> > Just to know, I have a question. Do you think that `account` dependency should be added in `"depends"` parameters?
> > @HaraldPanten
> > I believe it's not necessary because the [mis_builder module already has it as a dependency](https://github.com/OCA/mis-builder/blob/10.0/mis_builder/__manifest__.py#L15)

@jjscarafia I agree. I was wondering why some modules include all dependencies even if other dependencies include these other modules. 🤔

As you said, this would be the situation of `mis_builder_cash_flow`. As MIS Builder includes `account` as a dependency, there's no need to include it in `mis_builder_cash_flow` `"depends"` parameters. But I've realized that in some modules all dependencies are included.

I was curious to know if there are any "rules" to include them or not, or both options are correct.

Thanks for your reply

mis_builder_cash_flow/report/mis_cash_flow.py
MiquelRForgeFlow: Done :)

MiquelRForgeFlow:
> I was curious to know if there are any "rules" to include them or not, or both options are correct.

I think there are no rules about redundant dependencies. My opinion is: avoid to put them except in cases when the dependency is profound and not obvious.

AaronHForgeFlow (approved):
Functional tests. Working as expected

jjscarafia (approved):
code review +1
travis error seems related to other module

AaronHForgeFlow:
@MiquelRForgeFlow Could you rebase?

MiquelRForgeFlow:
@AaronHForgeFlow  done, ready to merge

Issue #545, opened 2019-05-17
[11.0] Trial Balance's Journal option does nothing


Arvinman:
By changing Journal Field in Trial Balance Report. Only thing that changes in result is Profit/Loss Account. All other general account still take all entries from all journal.

Is this Journal field still valid for use? Or do you have any plan to make it obsolete?

jalzaga:
How is the status of this issu? I have tried also in version 10 and no filtering is done for journals.

eLBati:
Hi, can you check the following fix?
 - v10: https://github.com/OCA/account-financial-reporting/pull/677
 - v11: https://github.com/OCA/account-financial-reporting/pull/679
 - v12: https://github.com/OCA/account-financial-reporting/pull/678

PR #678, opened 2020-04-23
[12.0] FIX account_financial_report trial balance: consider journals set in wizard

eLBati:
See https://github.com/OCA/account-financial-reporting/issues/545


andreampiovesana (approved):
go

sergiocorato (approved):

PR #665, opened 2020-03-11
[12.0] account_financial_report: Increase columns width in general ledger

grindtildeath:
Before this commit, amounts such as "10'000.00 CHF" were exceeding their
columns width. Label column could be reduced as it's printed correctly
over two different lines.

PR #686, opened 2020-05-06
[13.0] account_financial_report: Increase columns width in general ledger

Tonow-c2c:
idem as https://github.com/OCA/account-financial-reporting/pull/665 but for v`13.0`

Before this commit, amounts such as "10'000.00 CHF" were exceeding their
columns width. Label column could be reduced as it's printed correctly
over two different lines.


TDu (approved):

guewen (approved):

guewen:
@Tonow-c2c could you squash?

Issue #672, opened 2020-04-09
[13.0] Account Financial Report - Aged Partner Balance & Open Items


CasVissers-360ERP:
Both reports use date and due_date for calculation of the reports, should this not be updated to "invoice_date or date" and "invoice_date_due or due_date". The reports should be based on invoices when possible?

@JoanSForgeFlow

pedrobaeza:
AFAIK `invoice_date_due` is the due date you put manually on invoice if no payment_term is specified. Then, when posted, the valid due date is the one on the account.move.line, which has been transferred from the other field, but subsequent modifications must be done on aml field, isn't it?

CasVissers-360ERP:
@pedrobaeza 
My main issue is invoices where account_move.date != account_move.invoice_date. The option to do this is standardly available only on supplier invoices, but with a simple invisible = 0 also possible on customer invoices. I think you are correct about the due_date , but if we can use invoice dates throughout I think we should.

pedrobaeza:
I don't agree, at least on due dates. Imagine a payment term that split the payment in several parts: you will have several aml with different due dates corresponding to that parts.

Showing the invoice date when applicable in these reports though can be important. How was done in previous versions?

CasVissers-360ERP:
I agree on the due dates. Current workflow is better. Missed the several parts use case.

From what I can see (version 12) it contains the same workflow:
```
    ml.date,
    ml.date_maturity,
```
Since the date on move line is not always equal to the invoice date I think it's incorrect there as well. Or maybe I'm missing something?

pedrobaeza:
I think both dates (aml date and invoice date) can be put in the report, but I would qualify this as improvement.

CasVissers-360ERP:
Well I think an aged partner balance is oftenly used to check state of unpaid invoices. Not showing the invoice date seem illogical to me. Anyhow if considered an improvement, not a bug, I will close the issue.

PR #692, opened 2020-05-27
[13.0] [IMP] account_financial_report - Add invoice date to aged partner balance & open items

CasVissers-360ERP:
Solves #672

Invoice date can be different as date on account.move. Companies invoicing in month B for month A use both fields. It's crucial to show the (correct) invoice date on the partner reports.


account_financial_report/models/account_move_line.py
```diff
@@ -1,11 +1,17 @@
...
+from odoo import models, fields
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    invoice_date = fields.Date(related='move_id.invoice_date',
+                               string='Invoice/Bill Date',
+                               readonly=True,
+                               store=True,
```
MiquelRForgeFlow, line 12: Is needed to be stored?

PR #705, opened 2020-06-30
[13.0][FIX] account_financial_report: missing date in invoice

glitchov:
The second invoice created has no explicit date so taking the current date could lead to failing tests because the invoice is out of search as the the setup sets date_from as time.strftime("%Y-%m-01") and date_to as time.strftime("%Y-%m-28"). So i choose the easy way by setting an explicit date for the invoice.


pedrobaeza (approved):
Seems reasonable enough

michotm (approved):
Ok for me

JoanSForgeFlow (approved):
Great, LGTM

PR #709, opened 2020-07-20
[12.0][FIX] account_financial_report: Totals alignment with foreign currency

SimoRubi:
Fix https://github.com/OCA/account-financial-reporting/issues/708

### DIFF
--- a/account_financial_report/report/templates/open_items.xml
+++ b/account_financial_report/report/templates/open_items.xml
@@ -257,11 +257,11 @@
                     </t>
                     <t t-if="not account_or_partner_object.currency_id.id">
                         <!--## currency_name-->
-                        <div class="act_as_cell"/>
+                        <div class="act_as_cell" style="min-width: 2.25%;"/>
                         <!--## amount_total_due_currency-->
-                        <div class="act_as_cell"/>
+                        <div class="act_as_cell" style="min-width: 6.57%;"/>
                         <!--## amount_residual_currency-->
-                        <div class="act_as_cell"/>
+                        <div class="act_as_cell" style="min-width: 6.57%;"/>
                     </t>
                 </t>
             </div>


HviorForgeFlow (approved):

astirpe (approved):

pedrobaeza (commented):
Why are you fixing widths? That's normally not a good idea.

marcelsavegnago (approved):

SimoRubi:
> Why are you fixing widths? That's normally not a good idea.

The template is full of fixed widths, I agree that is not a good idea but it is in the stable version of the module, so I suppose it has been OCA-approved at a certain point.

I am adding width in the three cells because those cells are place-holders for the three cells in the above rows, and having them without width causes issues with other cells in the same row, something like:

**Before PR**
<table>
<tr>
    <th>Other columns</th>
    <th>Currency name</th>
    <th>Total due currency</th>
    <th>Residual currency</th>
  </tr>
<tr>
    <td>...</td>
    <td>0.0</td>
    <td>0.0</td>
    <td>0.0</td>
  </tr> 
<tr>
    <td colspan="4" align="right">Total other columns</td>
  </tr>
</table>

**After PR**
Other columns | Currency name | Total due currency | Residual currency
-- | - | - | -
... | 0.0 | 0.0 | 0.0
Total other columns

SimoRubi:
@OCA/accounting-maintainers how can I help this PR to be merged? Thanks!

pedrobaeza (dismissed):
I don't like that fixed widths, and the fix should be to remove the other fixed ones.

StefanRijnhart (approved):
Agreed. View is littered with fixed widths so as such this is an improvement. @pedrobaeza is this blocking for you?

pedrobaeza:
The problem I see that fixing this width will make other columns to overflow or change current "status quo", and this has been conflicting in the past.

SimoRubi:
I've tried to remove all the fixed widths in the template, and the issue still occurs.
The main problem is that the 'accumulating' row is actually another table, so its cells can't 'inherit' the width from above cells/header and they (correctly) don't get any space when they're empty.
I've tried to refactor the template so that the 'accumulating' row becomes the footer but it's quite a change considering the template's structure.

The only fix I have thinked of till now is adding fixed width for the empty cells, I'll propose a new solution as soon as I have something working but I don't think I'll be able to do that anytime soon.

SimoRubi:
> I don't like that fixed widths, and the fix should be to remove the other fixed ones.

Do you like `min-width`s better?

SimoRubi:
@OCA/accounting-maintainers I have found no way to remove the fixed width in current view, any suggestion is appreciated.
Is it possible to merge this fix?

Maybe the `width` issue should be tackled in a separated refactoring PR/issue.

SimoRubi:
@OCA/accounting-maintainers I have found no way to resolve https://github.com/OCA/account-financial-reporting/pull/709#pullrequestreview-476876526, can this be merged anyway or how should I proceed? Thanks!

pedrobaeza:
What happen when you remove all fixed sizes?

SimoRubi:
> What happen when you remove all fixed sizes?

The last time I tried (https://github.com/OCA/account-financial-reporting/pull/709#issuecomment-703526880), it behaved like with all the fixed widths (current behavior in https://github.com/OCA/account-financial-reporting/issues/708): the empty currency cells do not appear in the table

pedrobaeza:
And why not putting everything in the same table?

SimoRubi:
> And why not putting everything in the same table?

Thanks for the suggestion, I had the same thought but discarded the idea:
> I've tried to refactor the template so that the 'accumulating' row becomes the footer but it's quite a change considering the template's structure.

(from https://github.com/OCA/account-financial-reporting/pull/709#issuecomment-703526880)
Moreover, there is already a heavy un-merge-able refactoring for this module in https://github.com/OCA/account-financial-reporting/pull/660, (already wrote about it from https://github.com/OCA/account-financial-reporting/issues/708#issuecomment-660100721) and I wouldn't start another one here because it would be un-merge-able for the same reasons as https://github.com/OCA/account-financial-reporting/pull/660

pedrobaeza:
Such refactor was merged in v13, so I would focus there to fix the inner problems. I remove here my requested changes for not blocking, but please take it into account for next versions.

SimoRubi:
> please take it into account for next versions.

Sure I'll try to forward-port this fix into `13.0` when I can, added to https://github.com/OCA/account-financial-reporting/issues/708.

Should I also propose the fix to https://github.com/OCA/account-financial-reporting/pull/660? I'm not sure I can see any benefit in working on something that won't be merged

pedrobaeza:
No, and in fact, I'm going to close such PR for avoiding confusions. It's too much change for this old version.

PR #714, opened 2020-07-31
[10.0] improve account_financial_report_qweb

gurneyalex:
* fix some rounding issues in the OCA reports which can happen if some
account move lines have not been correctly rounded

* improve the excel export by using decimal.Decimal to avoid writing
fractions of a cent at the 12th decimal place

### DIFF
--- a/account_financial_report_qweb/report/abstract_report_xlsx.py
+++ b/account_financial_report_qweb/report/abstract_report_xlsx.py
@@ -2,6 +2,7 @@
 # Author: Julien Coux
 # Copyright 2016 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from decimal import Decimal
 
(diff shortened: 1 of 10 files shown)


gurneyalex:
I'm not happy with the hard coded '2' in the SQL queries. Do you think it is worth computing the this based on the rounding of the currencies?

yvaucher (approved):
LGTM

The 2 decimal constant for rounding could be an issue, but I couldn't name a country or a case where that would be. This fix should handle the vast majority of the issues we have with rounding in financial reports.

StefanRijnhart:
If you replace 2 by 6, that would still prevent the Python float artifact in the 12th decimal from showing up and not cause problems with currencies using a larger number of decimals? Oh, there would still be problems using bitcoin :-P. Is it a problem to derive the number of decimals from the currency?

PR #717, opened 2020-08-10
[12.0][FIX] account_financial_report; Add format to date fields

SimoRubi:
**Steps to reproduce issue**

* In runbot, go to Invoicing > Reporting > OCA accounting reports > Open items
* Print the XLSX report

**Behavior before this PR**
The date format is not the user's date format

**Behavior after this PR**
The date format is the user's date format

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -2,8 +2,10 @@
 # Author: Julien Coux
 # Copyright 2016 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import models, fields
+from odoo import models
 import datetime
+
+from odoo.tools import format_date
 
 
 class AbstractReportXslx(models.AbstractModel):
@@ -190,7 +192,7 @@ class AbstractReportXslx(models.AbstractModel):
         for col_pos, column in self.columns.items():
             value = getattr(line_object, column['field'])
             if isinstance(value, datetime.date):
-                value = fields.Date.to_string(value)
+                value = format_date(self.env, value)
             cell_type = column.get('type', 'string')
             if cell_type == 'many2one':
                 self.sheet.write_string(


account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -187,10 +187,13 @@ def write_line(self, line_object):
         """Write a line on current line using all defined columns field name.
         Columns are defined with `_get_report_columns` method.
         """
+        lang = self.env.user.lang
+        format_date = self.env['ir.qweb.field.date'].with_context(
+            lang=lang).value_to_html
```
astirpe: Why do you use the `value_to_html()` instead of the `odoo.tools.format_date`?

SimoRubi: Because it's the first way that I found to change the format, I found it in https://github.com/odoo/enterprise/blob/3963662509db09208e42ab660cda83c93946cb11/sale_subscription/models/sale_order.py#L174.

`odoo.tools.format_date` seems cleaner and works just as well, thanks!

astirpe (approved):

stevech091 (approved):
Functional test ok.

marcelsavegnago (approved):

account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -2,9 +2,11 @@
 # Author: Julien Coux
 # Copyright 2016 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import models, fields
+from odoo import models
 import datetime
 
+from odoo.tools import format_date
```
tafaRU, line 8: Shouldn't be `from odoo.tools.misc import format_date`?

tafaRU (changes requested):

account_financial_report/report/abstract_report_xlsx.py
SimoRubi: `format_date` is both in `odoo.tools` and `odoo.tools.misc` due to https://github.com/odoo/odoo/blob/98ec5df0ccea6c22024132c31010c0dac606cc27/odoo/tools/__init__.py#L10.
I have found both in the code, for instance:
* `odoo.tools`: https://github.com/odoo/enterprise/blob/c555459062af0a90296a08dad98b66180943935f/sale_subscription/models/sale_subscription.py#L14
* `odoo.tools.misc`: https://github.com/OCA/l10n-italy/blob/9ef10fa194279ad1c7e7c8502ca13cd49ff85cad/l10n_it_dichiarazione_intento/models/account_invoice.py#L6

tafaRU: > due to https://github.com/odoo/odoo/blob/98ec5df0ccea6c22024132c31010c0dac606cc27/odoo/tools/__init__.py#L10.

ok, thanks for sharing it.

tafaRU (approved):

PR #682, opened 2020-04-30
[13.0][IMP] account_financial_report: multiple fixes and improvements

JoanSForgeFlow:
This PR implements in v13 improvements and fixes done in v12:
- #660 

Allows selecting all the accounts between two codes in General Ledger, Trial Balance, Open Items, and Aged Partner Balance reports.

Also, it changes the way to choose accounts in Open Items and Aged Partner Balance reports.
None account is selected by default, and when selecting it is only possible to select accounts that allow reconciliation.

Complete tax name appears in general Ledger.

Added option for enabling "Partners Details" in General Ledger and added complete tax name and complete Ref-Label field in general Ledger. It also appears (future) next to the matching number if the reconciliation is done after "date_to".

Refactoring for speed up Open Items and Aged Partner Balance.

Fixes Vat Report.

PR #718, opened 2020-08-11
[13.0][FIX] vat report date domain

### LINKED ISSUE #710: [13] VAT Report Date selection "=" missing for less than or equal
In vat_report.py

Instead of ("date", "<", date_to), please change to ("date", "<=", date_to), 

Equal sign after "<"

Otherwise the last day of the entered period is not considered.

image

Thanks a lot

sbidoul:
fixes #710


astirpe:
Isn't it already the case? https://github.com/OCA/account-financial-reporting/blob/13.0/account_financial_report/report/vat_report.py#L37

sbidoul:
Oh yes, it looks that was part of #682

PR #720, opened 2020-08-31
[10.0][UPT]customer_outstanding_statement standard name is Aging Statement

AaronHForgeFlow:
All English speaking companies I have seen call this report Aging Statement. They find current name confusing. Just searching on the web by Aging Statement retrieves much more results than searching by Customer Outstanding Statement or Partner Statement. As long as we try to be the most standardized as possible I think this change is necessary. I am not changing anything in the data model, just strings, in order to avoid breaking current instances.

If community agrees I will forward port this change also to newer versions (partner_statement module).

### DIFF
--- a/customer_outstanding_statement/README.rst
+++ b/customer_outstanding_statement/README.rst
@@ -2,11 +2,11 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-====================================
-Print Customer Outstanding Statement
-====================================
+=====================
+Print Aging Statement
+=====================
 
-The outstanding statement provides details of all outstanding customer receivables
+The aging statement provides details of all outstanding customer receivables
 up to a particular date. This includes all unpaid invoices, unclaimed refunds and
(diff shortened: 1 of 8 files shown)


StefanRijnhart (approved):
Agreed. Terminology is 'Aging' in my experience.

gdgellatly:
@AaronHForgeFlow In general I never heard it titled that. Normally it is just called Account Statement, Statement of Account or just Statement. We always just customise for the second.

see https://www.accountingtools.com/articles/what-is-a-statement-of-account.html

MiquelRForgeFlow (changes requested):
That's wrong. The aging statement it's just the Aging report, a subreport of this module when you click on "include aging buckets".

See:
Selection_253
Selection_254

AaronHForgeFlow:
The report is the same, but with some more information. In any case the text shown when printing the aging buckets it is still Customer Outstanding Statement, which it is something I don't want.  In any case I don't think this will ever be merged. I will close this but please keep the branch alive :)

PR #732, opened 2020-10-09
[10.0][FIX] account_tax_balance: Empty bills/invoices should not be considered as refunds

SimoRubi:
**Steps to reproduce the issue**:
1. Create an invoice
2. In the invoice, add a line with total **10**
3. In the invoice, add a line with total **-10**

**Actual behavior**:
The generated move has type `Receivable refund`

**Expected behavior**:
The generated move has type `Receivable`

### DIFF
--- a/account_tax_balance/models/account_move.py
+++ b/account_tax_balance/models/account_move.py
@@ -2,7 +2,7 @@
 # © 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import models, fields, api
+from odoo import models, fields, api
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I'm not sure this works for the other invoice type: vendor bills if you have tried with customer invoices or vice versa.

SimoRubi:
> I'm not sure this works for the other invoice type: vendor bills if you have tried with customer invoices or vice versa.

Thanks for checking! I've actually tried it with both vendor bills and customer invoices, please have a look at the new tests

pedrobaeza:
And what happens with refunds of amount 0?

SimoRubi:
> And what happens with refunds of amount 0?

I've now checked that it has type `Receivable`/`Payable`, and it doesn't seem right, does it?

So maybe I'll have to fix the type computation a bit more thoroughly, thanks for the review!

SimoRubi:
I have updated the move type computation logic and extended the tests to also check both refund cases.
@pedrobaeza can you have a look if now it looks better?

account_tax_balance/models/account_move.py
```diff
@@ -24,22 +24,28 @@ class AccountMove(models.Model):
...
-                lambda x: x.account_id.internal_type == internal_type).mapped(
-                    'balance'))
+        refund_types = ('in_refund', 'out_refund')
+
+        def _is_refund(line_ids, internal_type):
+            """Check whether all the lines of type `internal_type`
+            come from a refund."""
+            line_ids = line_ids.filtered(
+                lambda x: x.account_id.internal_type == internal_type)
+            line_types = line_ids.mapped('invoice_id.type')
```
pedrobaeza, line 34: The problem here is those moves without invoice_id.

SimoRubi: I have searched the code and found no better way to determine if 'something' is a refund.
Using the balance has the issues described in this PR that I'd summarize in: what is **0**?

The moves without `invoice_id` would be represented as `Payable`/`Receivable`, according to their account type.
That doesn't seem that bad to me because if there is no refund (that is, an invoice of type `*_refund`), why should any move be identified as a refund?

This said, I'm no functional so if there is some better way to recognize a refund `account.move` please, I'm listening :)

tafaRU: @SimoRubi according to https://github.com/OCA/account-financial-reporting/pull/732/files#r502467979 could you fallback the previous behaviour in case of moves without `invoice_id`?

tafaRU (changes requested):

account_tax_balance/models/account_move.py
SimoRubi: Added fallback to previous behavior

tafaRU: thanks!

tafaRU (approved):

account_tax_balance/models/account_move.py
SimoRubi: @pedrobaeza can you update your review?
Thanks!

pedrobaeza (approved):

SimoRubi:
@OCA/accounting-maintainers can this be merged?

SimoRubi:
> /ocabot merge patch

thanks

PR #734, opened 2020-10-21
[13.0][FIX] account_financial_report: default company id is always the lowest ID

LoisRForgeFlow:
move company_id to the abstract model and correct default definition.

@ForgeFlow

### DIFF
--- a/account_financial_report/wizard/general_ledger_wizard.py
+++ b/account_financial_report/wizard/general_ledger_wizard.py
@@ -21,12 +21,6 @@ class GeneralLedgerReportWizard(models.TransientModel):
     _description = "General Ledger Report Wizard"
     _inherit = "account_financial_report_abstract_wizard"
 
-    company_id = fields.Many2one(
-        comodel_name="res.company",
-        default=lambda self: self.env.user.company_id,
-        required=False,
-        string="Company",
-    )
     date_range_id = fields.Many2one(comodel_name="date.range", string="Date range")
     date_from = fields.Date(required=True, default=lambda self: self._init_date_from())
     date_to = fields.Date(required=True, default=fields.Date.context_today)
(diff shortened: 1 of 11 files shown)


JoanSForgeFlow (approved):
LGTM :+1:

JordiBForgeFlow (approved):

JordiBForgeFlow:
@JoanSForgeFlow can you backport to 12.0?

HaraldPanten (approved):

LoisRForgeFlow:
@JordiBForgeFlow @JoanSForgeFlow I don't think it is needed in v12, `self.env.user.company_id` is fine in v12.

LoisRForgeFlow:
I aborted the merge. I need to include some more changes.

LoisRForgeFlow:
it should be good now.

account_financial_report/wizard/general_ledger_wizard.py
```diff
@@ -119,8 +113,8 @@ def on_change_account_range(self):
     def _init_date_from(self):
         """set start date to begin of current year if fiscal year running"""
         today = fields.Date.context_today(self)
-        last_fsc_month = self.env.user.company_id.fiscalyear_last_month
-        last_fsc_day = self.env.user.company_id.fiscalyear_last_day
+        last_fsc_month = self.env.company.fiscalyear_last_month
```
pedrobaeza: Shouldn't this be self.company_id?

LoisRForgeFlow: The problem is that company_id isn't required, but maybe we could use it and fallback to self.env.company which will be more correct. I'm changing it now.

pedrobaeza (approved):

PR #737, opened 2020-10-30
[13.0][IMP] account_financial_report: vat_report "hide_tax_at_zero"

JoanSForgeFlow:
Adds feature hide_tax_at_zero. Not enabling this option all taxes appear, which means all taxes of a company appear always in the same column if no changes in taxes. This is useful for xlsx reports where accountants have their own calculation templates. 

This is a cherry-pick of a commit added here. #660 

Also fixes for increase the readability of the code


rafaelbn:
Hello! Maybe @SirTakobi you could review this PR and @JoanSForgeFlow could review yours https://github.com/OCA/account-financial-reporting/pull/885

Issue #690, opened 2020-05-15
[13.0] clicking on amount wont show you the details of the account move lines


yung-wang:
1. Go to runbot and go to reports OCA General Ledger.
2. Click on View, so you can see the report on screen.

Try to click on amounts. Nothing happens. This used to work in V12.

I see the click event in the JS file.
https://github.com/OCA/account-financial-reporting/blob/13.0/account_financial_report/static/src/js/account_financial_report_widgets.js

> "click .o_account_financial_reports_web_action": "boundLink",
            "click .o_account_financial_reports_web_action_multi": "boundLinkmulti",
            "click .o_account_financial_reports_web_action_monetary":
                "boundLinkMonetary",
            "click .o_account_financial_reports_web_action_monetary_multi":
                "boundLinkMonetarymulti",

kittiu:
We found it from today version of this report also. Any updates?

gjotten:
> We found it from today version of this report also. Any updates?

@kittiu and @yung-wang the above two PRs point towards the solution.

PR #744, opened 2020-11-26
[MIG] 13.0 Redo of account_financial_report

thomaspaulb:
Excuse the hack and slash approach; I needed a solution to https://github.com/OCA/account-financial-reporting/issues/690 quick and so I redid the migration, keeping the old approach of showing the HTML as a view so that it can be clicked.


account_financial_report/static/src/js/account_financial_report_backend.js
```diff
@@ -4,17 +4,22 @@ odoo.define("account_financial_report.account_financial_report_backend", functio
...
     var core = require("web.core");
-    var Widget = require("web.Widget");
-    var ControlPanelMixin = require("web.ControlPanelMixin");
+    var AbstractAction = require('web.AbstractAction');
     var ReportWidget = require("account_financial_report.account_financial_report_widget");
 
-    var report_backend = Widget.extend(ControlPanelMixin, {
+    var report_backend = AbstractAction.extend({
+        // TODO: why isnt the controlpanel shown?
+        hasControlPanel: true,
```
thomaspaulb, line 12: @NL66278 Despite all of my attempts here, the control panel is still not shown. Do you know what is going wrong?

account_financial_report/report/templates/trial_balance.xml
```diff
@@ -782,220 +753,238 @@
...
-                    <!--## amount_total_due_currency-->
-                    <div class="act_as_cell" style="width: 9%;" />
-                    <!--## amount_residual_currency-->
-                    <div class="act_as_cell" style="width: 9%;" />
+                    <t t-if="account.currency_id.id">
+                        <!--## currency_name-->
+                        <div class="act_as_cell" style="width: 4.43%;">
+                            <t
+                                t-set="domain"
+                                t-value="[('account_id', '=', account.account_id.id)]"
```
thomaspaulb, line 1106: These domains should take into account the state, based on the value of 'only_posted_moves'.

account_financial_report/static/src/js/account_financial_report_backend.js
NL66278: @thomaspaulb I have no idea I do have only an extremely limited experience with javascript in Odoo...

gjotten:
For context, @thomaspaulb had to do this quickly and maybe a bit drastically in order to fix #690

PR #750, opened 2021-01-22
[12.0][FIX] account_financial_report: add index to report_move_id

janverb:
When running `DELETE FROM report_journal_ledger_move` this column is checked repeatedly, even though the table is empty. On our setup that caused the query to take hours. Adding an index solved it.

### DIFF
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -722,6 +722,7 @@ class ReportJournalLedgerMoveLine(models.TransientModel):
         comodel_name='report_journal_ledger_move',
         required=True,
         ondelete='cascade',
+        index=True,
     )
     move_line_id = fields.Many2one(
         comodel_name='account.move.line',


gaikaz (approved):
Tested

pedrobaeza:
Let's go with this option:

/ocabot merge patch

pedrobaeza:
@JoanSForgeFlow @JordiBForgeFlow do you know if this is needed in 13.0?

dalonsod:
@pedrobaeza I think this PR is not needed for v13, since starting from it the report calculation is changed, and no transient models are used anymore.
Indeed, the commit that makes the major changes (https://github.com/OCA/account-financial-reporting/commit/e41e718bfcb9b1724f37077ff17c8b2e76d96400) mentions 12.0 in the commit description, but I've found that it's been never applied to v12 branch, and it should be a good improvement for v12 and even v11. Maybe @JoanSForgeFlow could better explain this, anyway.

PR #752, opened 2021-01-30
account_financial_report: Fix slow Journal Ledger vacuum

### LINKED ISSUE #727: Slow auto vacuum on journal ledger report
Hi all,

i've encountered a very slow delete statement in here:

https://github.com/OCA/account-financial-reporting/blob/a3eacc2fe34d84ce0fc0bbefc7002dcebd2544b9/account_financial_report/report/journal_ledger.py#L625

This will get called on auto vacum

In my case it took longer than the default timeout - so it restarted again and again

Wouldn't it be much faster to use:

truncate report_journal_ledger_move, report_journal_ledger_move_line;

gaikaz:
Fixes #727


pedrobaeza:
Duplicated of #750

PR #758, opened 2021-02-25
[14.0][FIX] account_financial_report +  account_tax_balance: Fix tests

joao-p-marques:
On multi-company settings, we need to make sure we are searching and creating objects with the correct company associated.
Also take into account previously created account groups

@Tecnativa
TT28423

### DIFF
--- a/account_financial_report/tests/test_general_ledger.py
+++ b/account_financial_report/tests/test_general_ledger.py
@@ -7,45 +7,34 @@ import time
 from datetime import date
 
 from odoo import api, fields
-from odoo.tests import common
+
+from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 
 
-class TestGeneralLedgerReport(common.TransactionCase):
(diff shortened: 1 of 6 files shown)


pedrobaeza:
I'm not sure why you need to restrict all the searches. You only need to make sure the user is in and only in the right company. ORM record rules should do the rest.

joao-p-marques:
> I'm not sure why you need to restrict all the searches. You only need to make sure the user is in and only in the right company. ORM record rules should do the rest.

Yes, but with the latest refactor I added, using the base tests from the `account` module, 2 companies are created associated to the same test user. So, to avoid problems, it is better to restrict the search to a single one.

joao-p-marques:
Tests were failing now because of `account_tax_balance` due to an unrelated problem. I pushed another commit to fix it.

pedrobaeza (approved):

victoralmau (approved):

PR #763, opened 2021-03-30
[9.0][FIX] account_tax_balance: init hook refund confusion

StefanRijnhart:
Regression of https://github.com/OCA/account-financial-reporting/pull/647


pedrobaeza (approved):
True... fast-tracking

PR #766, opened 2021-04-06
[14.0][FIX] account_financial_report fix default company

CasVissers-360ERP:
Company_id on res.users is used as a default company. When opening the wizard in company A it should not use company B since this is the default company of a user. It should use the current user.

@pedrobaeza


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as proper bugfix and straight-forward:

/ocabot merge patch

PR #767, opened 2021-04-11
[13.0][imp][account_tax_balance] Use direct SQL queries to improve performance

JordiBForgeFlow:
The tax balance report searches using a m2m field, which kills performance in large databases. We need direct SQL queries in order to work around this issue. See https://github.com/odoo/odoo/issues/30350

cc @Yajo @pedrobaeza @JoanSForgeFlow

### DIFF
--- a/account_tax_balance/models/account_tax.py
+++ b/account_tax_balance/models/account_tax.py
@@ -2,7 +2,7 @@
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, fields, models
+from odoo import api, fields, models
 
 
 class AccountTax(models.Model):
@@ -18,11 +18,6 @@ class AccountTax(models.Model):
     base_balance_refund = fields.Float(
         string="Base Balance Refund", compute="_compute_balance"
     )
-    has_moves = fields.Boolean(
-        string="Has balance in period",
-        compute="_compute_has_moves",
-        search="_search_has_moves",
-    )
 
     def get_context_values(self):
         context = self.env.context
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I don't like to use SQL in the module, so other solution should be found.

LoisRForgeFlow:
@pedrobaeza Wouldn't this have a chance if the performance improvement is significant? What other options would there be? A setting option to use SQL at the risk of the implementator? It would be great if we don't kill the discussion this quickly :grin:

pedrobaeza:
The option is to do the searches/domains for not having such drawback, but not to go again to the SQL-hell that we had on `account_financial_report`.

JoanSForgeFlow:
Hi @pedrobaeza ,

The current approach of the account_tax_balance runs first this direct SQL for selecting ids of the used taxes:

```
SELECT id
FROM account_tax at
WHERE
company_id = 1 AND
EXISTS (
SELECT 1 FROM account_move_Line aml
WHERE
date >= '2021-03-01' AND
date <= '2021-03-31' AND
company_id = 1 AND (
  tax_line_id = at.id OR
  EXISTS (
    SELECT 1 FROM account_move_line_account_tax_rel
    WHERE account_move_line_id = aml.id AND
      account_tax_id = at.id
  )
)
)
```
Then python code ends in 4 queries for each tax whit this structure:

```
SELECT min("account_move_line".id) AS id, 
count("account_move_line".id) AS "__count" , 
sum("account_move_line"."balance") AS "balance"
FROM "account_move" as "account_move_line__move_id",
"account_move_line"
WHERE ("account_move_line"."move_id"="account_move_line__move_id"."id") 
AND ((((((("account_move_line"."date" <= '2021-03-31')  
AND  ("account_move_line"."date" >= '2021-03-30'))  
AND  ("account_move_line"."company_id" = 1))  
AND  ("account_move_line__move_id"."state" in ('posted')))  
AND  ("account_move_line"."id" in (
SELECT "account_move_line_id" FROM "account_move_line_account_tax_rel" WHERE "account_tax_id" IN (5))))  
AND  ("account_move_line"."tax_exigible" = true))  
AND  ("account_move_line__move_id"."move_type" in ('receivable','payable','liquidity','other'))) 
AND ("account_move_line"."company_id" IS NULL OR  ("account_move_line"."company_id" in (1)))
```
With the new approach, we remove the first direct SQL query, and then we reduce the number of queries to 4 for calculating all taxes. 

```
SELECT SUM(balance), 
rel.account_tax_id as tax_id 
FROM account_move_line AS aml 
INNER JOIN account_move AS am 
ON aml.move_id = am.id 
INNER JOIN account_move_line_account_tax_rel AS rel 
ON aml.id = rel.account_move_line_id 
WHERE aml.date <= '2021-03-31' 
AND aml.date >= '2021-03-01' 
AND aml.company_id = 1 
AND am.state IN ('posted') 
AND aml.tax_exigible = True  
AND am.move_type IN ('receivable', 'payable', 'liquidity', 'other') 
GROUP BY rel.account_tax_id
```
But as tax_ids is an m2m field we are not able to end in this type of query using python as there is no possibility of use read_group() method.
Do you have an idea of how we can set this type of query not using python SQL code?

pedrobaeza:
I will investigate and tell you back.

JordiBForgeFlow:
@JoanSForgeFlow We can even increase speed by doing just one query if we group by move_type in the query, and we do just one call to compute_balance. So, in one query you get the full results.

@pedrobaeza this report, in large databases, is completely unusable, as it is right now.

JoanSForgeFlow:
With this new approach, we have reduced the time for reporting a month from 5 minutes 30 secs to only 15 secs.

Issue #770, opened 2021-04-15
Tax balance migration to 14.0


SimoRubi:
Hi all,
I am migrating an enterprise DB from 10.0 to 14.0, I have migrated the core with https://upgrade.odoo.com/upload and in the upgrade report they mention:
> The field 'move_type' from the model 'account.move' is now a standard Odoo field, but it already existed in the database (coming from a non-standard module) and thus has been renamed to 'move_type_custom'

I am experiencing a _small_ issue with this code:
https://github.com/OCA/account-financial-reporting/blob/a648e2c7482ff56a41fc7bde598d00fdfb8a43a6/account_tax_balance/migrations/14.0.1.0.0/pre-migration.py#L17
Before this is executed, account moves have these types:
```
# select move_type from account_move group by move_type;
  move_type  
-------------
 entry
 in_invoice
 in_refund
 out_invoice
 out_refund
(5 rows)
# select move_type_custom from account_move group by move_type_custom;
 move_type_custom  
-------------------
 payable_refund
 
 payable
 receivable_refund
 receivable
 other
 liquidity
(7 rows)
```
As the mentioned report says, they have correctly moved the old `move_type` to `move_type_custom`.

After the execution of linked code, account move types are like this:
```
# select move_type from account_move group by move_type;
 move_type 
-----------
 entry
(1 row)
# select move_type_custom from account_move group by move_type_custom;
 move_type_custom  
-------------------
 payable_refund
 
 payable
 receivable_refund
 receivable
 other
 liquidity
(7 rows)
# select financial_type from account_move group by financial_type;
 financial_type 
----------------
 out_invoice
 in_invoice
 in_refund
 entry
 out_refund
(5 rows)
```

The issue is that now none of the account moves is presented as an invoice, so there are _apparently_ no invoices.
In order to solve this, I'll simply copy `financial_type` to `move_type` and `move_type_custom` to `financial_type`, but the question is: could this issue be prevented?

The problem here is that the migration resets the `move_type` field but I can't see how that can happen in https://github.com/OCA/openupgradelib/blob/01cc9a1a8933376dcf6516b9a38030c6d0f16072/openupgradelib/openupgrade.py#L574
So I don't really know if 

- this is expected from openupgradelib, so the upgrade script has to be fixed
- this is not expected from openupgradelib, so it has to be fixed over there
- I had to do something before migrating the module

pedrobaeza:
[...]

PR #771, opened 2021-04-16
Manage renamed move_type column

SimoRubi:
Fix https://github.com/OCA/account-financial-reporting/issues/770

### DIFF
--- a/account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
+++ b/account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
@@ -1,17 +1,31 @@
 # Copyright 2020 Ozono Multimedia S.L.L.
+# Copyright 2021 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
-
-field_renames = [
-    (
-        "account.move",
-        "account_move",
-        "move_type",
-        "financial_type",
-    ),
-]
 
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.rename_fields(env, field_renames)
+    """
+    Column `move_type` of table `account_move` has been renamed to `financial_type`
+    because `move_type` is now used by the core,
+    so the column is moved during migration of module `account` in the core.
+    Enterprise renames it to `move_type_custom`;
+    OpenUpgrade renames it to its legacy name.
+    Move data from the renamed column to the new `financial_type` column.
+    """
+    old_move_type_column = "move_type"
+    new_move_type_column = "financial_type"
+    move_table_name = "account_move"
+    enterprise_move_type_rename = "move_type_custom"
+    ou_move_type_rename = openupgrade.get_legacy_name(old_move_type_column)
+
+    for move_type_rename in (enterprise_move_type_rename, ou_move_type_rename):
+        if openupgrade.column_exists(env.cr, move_table_name, move_type_rename):
+            openupgrade.rename_columns(
+                env.cr,
+                {
+                    move_table_name: [(move_type_rename, new_move_type_column)],
+                },
+            )
+            break


account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
```diff
@@ -1,17 +1,26 @@
...
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.rename_fields(env, field_renames)
+    move_table_name = "account_move"
+    enterprise_move_type_rename = "move_type_custom"
+    if openupgrade.column_exists(env.cr, move_table_name, enterprise_move_type_rename):
+        old_move_type_col = enterprise_move_type_rename
+    else:
+        old_move_type_col = openupgrade.get_legacy_name("move_type")
```
pedrobaeza: Please don't do this part. It will be done on OpenUpgrade itself.

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
```diff
@@ -1,17 +1,26 @@
...
+
+    openupgrade.rename_fields(
```
pedrobaeza: This doesn't serve for enterprise upgrade, as I don't think filters and the rest of the stuff is renamed from `move_type` to `move_type_custom`.

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
SimoRubi: Thanks for having a look!

Shouldn't each module take care of its own fields?

I don't know OpenUpgrade that well, but I thought it only existed for managing migration of core modules; if it also takes care of fields in any OCA module, that might be confusing.

For instance, if there is a module in v14 that has no migration script I think that it does not require any, but maybe there is something in OpenUpgrade repo that takes care of everything?.

In a situation (my current situation) where core has been migrated by Odoo, how should the above example be managed?

pedrobaeza: As we have to rename anyway to another field name in OpenUpgrade, it doesn't makes sense to rename to something different from the target name and make twice the renaming operation.

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
SimoRubi: Thanks, I've separated the two approaches, I'll remove the OpenUpgrade branch depending on how https://github.com/OCA/account-financial-reporting/pull/771#discussion_r614902420 goes

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
SimoRubi: So in general migration scripts should only manage changes coming from enterprise migration and everything else should be implemented in OpenUpgrade, or is this a special case for not renaming twice the same data?

I also have in mind this old comment (https://github.com/OCA/manufacture/pull/595#issuecomment-779052897) where the renaming of the module has been done in OpenUpgrade; isn't this centralization causing issues when core is not migrated by OpenUpgrade? Or am I missing something?

Not to mention that distributing migration scripts in each module could ease OpenUpgrade's scripts development.

pedrobaeza: No, this is a special case, as the previous field name is occupied by a core field, and not renaming in OpenUpgrade will lead to problems in update process. In general, we can do the needed operations in the migration scripts of the module itself.

pedrobaeza (approved):

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
pedrobaeza: The module renaming/merging can only be done on OpenUpgrade in a centralized way, as they need to be handled at the very beginning of the upgrade process, or it can be conflicting.

For more things about the host of the migration scripts (which is not the same of where to perform the operations in the update pipeline), I have opened https://github.com/OCA/OpenUpgrade/issues/2661

SimoRubi: Big thanks! I'll be following that

SimoRubi:
Rebased to see if Travis gets 🟢

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
StefanRijnhart: I would be happy to have the renaming done here based on the OpenUpgrade legacy name. There is the possibility that the column does not belong to this module (which would then not be installed) but to an unrelated module which might want to rename to a different new column name.

SimoRubi: Thanks @StefanRijnhart for having a look :)
@pedrobaeza would the change https://github.com/OCA/account-financial-reporting/pull/771#discussion_r641520009 be ok with you?
That is kind of different from what you wrote in https://github.com/OCA/account-financial-reporting/pull/771#discussion_r614923746:
> As we have to rename anyway to another field name in OpenUpgrade, it doesn't makes sense to rename to something different from the target name and make twice the renaming operation.

We have to be aligned on this with OpenUpgrade scripts.
If you agree I can do the implementation and then let's discuss on that new code

marcelsavegnago (approved):

account_tax_balance/migrations/14.0.1.0.0/pre-migration.py
pedrobaeza: By the nature of OpenUpgrade, the rename must be done there on the pre-migration of `account` module, or the old column will be used for the new purpose with no warning nor change. We can add to OpenUpgrade the same layer as in enterprise for auto-renaming conflicting columns, but while this happens, we don't have other choice.

StefanRijnhart: Yes, OpenUpgrade renames to the legacy name, and here the legacy named column is renamed to the new column name.

pedrobaeza: I don't remember such automatic mechanism.

StefanRijnhart: I don't know how to explain this otherwise than just with code. My suggestion is to have in OpenUpgrade 14.0 account pre-migration:
```
if openupgrade.column_exists(env.cr, "account_move", "move_type"):
    openupgrade.rename_columns(cr, [("account_move",` [("move_type", None)])])
```
And then have here in the module migration:
```
for old in ("move_type_custom", openupgrade.get_legacy_name("move_type"):
    if openupgrade.column_exists(env.cr, "account_move", old):
        openupgrade.rename_columns(
            env.cr, [("account_move", [(old, "finance_type")])])
        break
```
I understand from the discussion above that you preferred renaming the column once, having this in OpenUpgrade:
```
if openupgrade.column_exists(env.cr, "account_move", "move_type"):
    openupgrade.rename_columns(cr, [("account_move",` [("move_type", "finance_type")])])
```
I prefer the version above so that this module (or perhaps an unrelated module also defining move_type in 13.0) takes ownership of its own columns.

pedrobaeza: Thanks for the extra explanations. The question is that this can only be executed on a migration from 13 to 14, so there are 2 options:

- You have this module installed, so the rename takes place on OpenUpgrade and nothing more to be done.
- You don't have this module installed, so no collapse at the end.

The same bad is that you have other module installed and not this one with such column, as OpenUpgrade will rename to the `openupgrade_legacy...` or `financial_type` name.

Do you get my point?

StefanRijnhart: Not sure I understand
> You don't have this module installed, so no collapse at the end.

Do you mean, if the module is not installed there is no column `move_type` in the 13.0 database? That is the assumption I'm challenging here.

> The same bad is that you have other module installed and not this one with such column, as OpenUpgrade will rename to the openupgrade_legacy... or financial_type name

This is my point. If `move_type` in 13.0 is owned by an unrelated module, renaming to openupgrade legacy format is predictable according to the openupgrade standard where as renaming to `financial_type` is random from the perspective of that module.

pedrobaeza: I still prefer to rename to `financial_type` in one shot, than having to rename 2 times, but it's a matter of taste. My point is that if we force here such renaming from `openupgrade_legacy_...` to `financial_type`, then we will need to know that when doing the OpenUpgrade part, so 2 works to do. If we leave this neutral on that part, only one work is needed.

StefanRijnhart: Don't worry, me and other people will be keeping an eye on that.
@SimoRubi do I understand correctly that you prefer the renaming to be done in this module too?

SimoRubi: Yes for the same reasons you explained in https://github.com/OCA/account-financial-reporting/pull/771#discussion_r641520009: OpenUpgrade should not assume that column `move_type` comes from this module and rename it to `financial_type`, it should instead rename the column to a neutral (legacy) name.
You can PR https://github.com/SimoRubi/account-financial-reporting/tree/14.0-imp-account_tax_balance-migration_enterprise proposing the change in https://github.com/OCA/account-financial-reporting/pull/771#discussion_r641765326 to keep authorship or I can implement it.

StefanRijnhart: Please go ahead. I think the suggestions are very much along the lines of your initial implementation so credits to you!

SimoRubi:
Re-requested reviews because I added some changes after the discussion that started in https://github.com/OCA/account-financial-reporting/pull/771#discussion_r614902420.

StefanRijnhart (approved):

pedrobaeza (approved):

PR #772, opened 2021-04-17
[13.0][account_financial_reporting] fix errors related to multicompany

JordiBForgeFlow:
- Display the currency of the company that was selected by the user, not
  the company that the user is currently active on.
- In the wizards, show only the companies that the user has chosen to work on,
  instead of all companies.

cc @JoanSForgeFlow @MateuGForgeFlow @LoisRForgeFlow @MiquelRForgeFlow


MateuGForgeFlow (approved):
Rebased to 13.0 and tested locally.

MiquelRForgeFlow (approved):
Makes sense.

PR #775, opened 2021-04-27
[13.0][FIX] account_financial_reporting: Pass res_id correctly to lines

joao-p-marques:
In the open items ledger, the res_id attribute of the invoice
in the report was getting: (id, move_name)
The result was that, when clicking the line, it would redirect
to a new record, instead of the existing.
This passes only the id to the line, solving that issue.

@Tecnativa
TT29371

ping @pedrobaeza @victoralmau


pedrobaeza (approved):

HviorForgeFlow (approved):
:+1:

marcelsavegnago (approved):

PR #776, opened 2021-04-27
[13.0][FIX] account_financial_report: allow navigation on all fields

joao-p-marques:
As with #775 , do the same for the **journal**, **account** and **partner** columns, to allow navigation to the related records.
This was present in v12, but was lost in the migration process. 
In this case, there isn't any new data being passed to the report, we are just reusing data already present, so no performance issues should come up.

@Tecnativa
TT29371

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_financial_report/report/templates/open_items.xml
+++ b/account_financial_report/report/templates/open_items.xml
@@ -186,16 +186,34 @@
             </div>
             <!--## journal-->
             <div class="act_as_cell left">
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
cc @JoanSForgeFlow

victoralmau (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port both to v14

PR #784, opened 2021-05-21
Revert "[10.0][IMP] account_financial_report_qweb: Improve performances"

JordiBForgeFlow:
This reverts commit 3d5e66991892cec9995401e50c9018f5bbb0e127.

PR https://github.com/OCA/account-financial-reporting/pull/655/files was incorrect due to this line:
https://github.com/OCA/account-financial-reporting/pull/655/files#diff-c87a48df934f9645d23d82970b5a5c51c0fbf14b00a1ea48d8a9e64e57d4ae41R437

This will drive incorrect results in the open items reports when you report on a past date.

For example create the following journal entries that involve the account 'Accounts receivable':

- Entry 1 (01/01/2021)
- Entry 2 (03/01/2021)

Then match them today (21/05/2021).

If you run the report on 10/01/2021 the report will show Entry 1 and Entry 2 as open.
Only when you run the open items report with date 22/05/2021 will the entries disappear from the open items report because they have been matched.

### DIFF
--- a/account_financial_report_qweb/report/open_items.py
+++ b/account_financial_report_qweb/report/open_items.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2016 Julien Coux (Camptocamp)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from datetime import timedelta
 
 from odoo import models, fields, api, _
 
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@rousseldenis this will reduce the performance of the report, but will be accurate. If you run the same report in odoo 13, it will follow the rule that in 01/03/2021 Entry 1 and Entry 2 are matched and the entries won't appear in the financial report.

MiquelRForgeFlow (approved):

rousseldenis:
Hi, I'm not at all in the context anymore...

Maybe @Cedric-Pigeon @sbidoul

JordiBForgeFlow:
I realized that Odoo 10 standard financial reports actually did it wrongly as well. So, after all we won't need this. I'm closing the PR.

PR #666, opened 2020-03-25
[13.0][ADD]account_move_line_report_xls

luc-demeyer:

PR #786, opened 2021-05-25
[14.0] [MIG]  account move_line_report_xls

PieterPaulussen:
Forward port of https://github.com/OCA/account-financial-reporting/pull/666#pullrequestreview-667647747


StefanRijnhart:
Some boilerplate here:
* please squash the earlier commits by the original author
* readd the underscores in the module name in this PR's title that Github so helpfully replaced by spaces so that we cannot find this PR back when we such by module name
* This is a [MIG] not an [ADD], and also otherwise respect specified commit message format from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.
* check Travis (see https://github.com/OCA/reporting-engine/commit/97e530adeb7267a49b18d1634d67b1d270f635fb#diff-f2686ed828c24d988d12b8f50d114dd6e61577f796f272d18682c4bd8d44c0f3 for a relevant refactoring related to the fact that Odoo 14 does not allow you anymore to store random data as class attributes).

PR #788, opened 2021-05-27
[FIX] account_financial_report: filter by correct data in trial balance

igallart:
When on a Trial Balance report and filtering by hierarchical accounts or for a foreing currency, an error was produced when generating the report.

The report fields were dependant on data that is no longer present, such as line data.

This uses the correct data provided when generating the report to avoid this errors

TT29261


pedroortega-punt:
@pedrobaeza está funcional para mergear. Gracias

pedrobaeza (approved):

pedrobaeza:
Fast-tracking it:

/ocabot merge patch

PR #790, opened 2021-05-28
[12.0][FIX] account_tax_balance: prevent multicompany glitch in tests

StefanRijnhart:
Prevent selecting a journal from another company, resulting in
UserError(_("Cannot create moves for different companies.")).

### DIFF
--- a/account_tax_balance/tests/test_account_tax_balance.py
+++ b/account_tax_balance/tests/test_account_tax_balance.py
@@ -168,7 +168,8 @@ class TestAccountTaxBalance(HttpCase):
         self.env['account.move'].create({
             'date': Date.context_today(self.env.user),
             'journal_id': self.env['account.journal'].search(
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

astirpe (approved):

astirpe:
Seems to me that the merge process is stuck

astirpe:
I'm noticing that the bumping of the version is not taking care of the latest version. In this case: from 12.0.1.4.5 to 12.0.1.1.2
See https://github.com/OCA/account-financial-reporting/commits/12.0

StefanRijnhart:
@astirpe but it's not the same module, so it looks alright to me.

Commits:

image

Version bump due to this PR:

image

astirpe:
Right, sorry :see_no_evil: I need holidays!

PR #791, opened 2021-05-28
[13.0][account_financial_report][fix] define XLSX currency format based on the currency of the company selected for the report.

JordiBForgeFlow:
Otherwise the system will use the currency associated to the user's default company, which is not necessarily the company that he's trying to get the report for.

cc  @LoisRForgeFlow  @JoanSForgeFlow

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -39,7 +39,7 @@ class AbstractReportXslx(models.AbstractModel):
 
         self.row_pos = 0
 
-        self._define_formats(workbook)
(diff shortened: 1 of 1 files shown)


JoanSForgeFlow (approved):
Tested locally, LGTM!

HaraldPanten (approved):

pedrobaeza:
Is there the possibility of mixing several companies in the same `data` set?

JordiBForgeFlow:
@pedrobaeza not to my knowledge. When you run the financial reports you must choose one and only one company.

pedrobaeza:
OK then:

/ocabot merge patch

Please fw-port it to v14

PR #792, opened 2021-06-01
[13.0][account_financial_report][fix] The aged and open items reports should ignore move lines with a zero balance

JordiBForgeFlow:
Even when they have an amount residual.
Those entries can exist as a result of an exchange rate difference.

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -140,6 +140,9 @@ class AgedPartnerBalanceReport(models.AbstractModel):
             move_line
             for move_line in move_lines
             if move_line["date"] <= date_at_object
+            and not float_is_zero(
+                move_line["debit"] - move_line["credit"], precision_digits=2
+            )
             and not float_is_zero(move_line["amount_residual"], precision_digits=2)
         ]
         for move_line in move_lines:
--- a/account_financial_report/report/open_items.py
+++ b/account_financial_report/report/open_items.py
@@ -85,6 +85,9 @@ class OpenItemsReport(models.AbstractModel):
             move_line
             for move_line in move_lines
             if move_line["date"] <= date_at_object
+            and not float_is_zero(
+                move_line["debit"] - move_line["credit"], precision_digits=2
+            )
             and not float_is_zero(move_line["amount_residual"], precision_digits=2)
         ]
 
--- a/account_financial_report/report/open_items.py
+++ b/account_financial_report/report/open_items.py
@@ -88,7 +88,13 @@ class OpenItemsReport(models.AbstractModel):
             and not float_is_zero(
                 move_line["debit"] - move_line["credit"], precision_digits=2
             )
-            and not float_is_zero(move_line["amount_residual"], precision_digits=2)
+            and (
+                not float_is_zero(move_line["amount_residual"], precision_digits=2)
+                or move_line["currency_id"]
+                and not float_is_zero(
+                    move_line["amount_residual_currency"], precision_digits=2
+                )
+            )
         ]
 
         open_items_move_lines_data = {}


pedrobaeza:
I don't see this use case. Please specify exact steps to reproduce the problem.

JordiBForgeFlow:
@pedrobaeza we have seen cases where the debit and credit is zero, and there's an amount_residual stated. Typically this happens in journal entries posting exchange rate differences, and causes the aging report and the open items report to show incorrect numbers. The odoo standard aging report calculation excludes those entries https://github.com/odoo/odoo/blob/13.0/addons/account/report/account_aged_partner_balance.py#L147 and when you compare the two reports an increpancy arises due to this.

pedrobaeza:
I think that should be patched at Odoo level, not here, as there's no sense of having amount_residual in a 0 line.

JordiBForgeFlow:
@pedrobaeza since this does not make sense, it won't be harmful to you, and solves my problem :)

pedrobaeza:
Well, it supposes more computation to be done for each move line.

StefanRijnhart:
Aligning with Odoo behaviour seems alright to me, but it would be good to have an Odoo issue number for  the problem of the zero balance move line with a non-zero residual amount, and then add a comment in this code referring to that issue. Do you know, by any chance, how that issue can be reproduced?

MiquelRForgeFlow (approved):
This issue won't be fixed in odoo (in v13), so it's ok to me.

JordiBForgeFlow:
@StefanRijnhart @pedrobaeza this is yet another example that these exchange rate journal entries with zero debit and credit well exist https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L3275

How to reproduce:

Assuming that company currency is USD.

Create journal entry 1:

Dr. Bank 100 USD (Amount currency 200 EUR)
Cr. Account Receivable 100 USD (Amount currency -200 EUR)

Journal entry 2:
Cr. Bank 100 USD
Dr. Account Receivable 100 USD

Reconcile the Account Receivable for the journals 1 and 2. You will see that the system will create a journal entry with label "Exchange Rate Difference"

Dr. Account Receivable 0 USD (Amount currency 134.59 EUR)
Cr. Foreign Exchange Gain 0 USD  (Amount currency -134.59 EUR)


image

StefanRijnhart:
So the issue is fixed in Odoo 14? And there is no Odoo issue for 13?

JordiBForgeFlow:
@StefanRijnhart it's not fixed in v14. I just reproduced the same scenario in runbot v14. This looks to me more like the intended behaviour. I don't think that they will consider it as a bug

image

StefanRijnhart:
I'm looking at those entries in the 14.0 runbot database (https://8472239-14-0-all.runbot37.odoo.com) right now, and these zero balance entries do not have a residual amount (which is what I would consider the bug):
image

If they don't have a residual amount, then it is not a valid reproduction of the case that this PR tries to deal with.

JordiBForgeFlow:
@StefanRijnhart I used this database https://8472240-14-0-all.runbot54.odoo.com/web#action=446&model=account.move.line&view_type=list&cids=1&menu_id=579

JordiBForgeFlow:
image

StefanRijnhart:
Sorry for the confusion, I must have been using that database as well, because the zero balance forgeflow entries in that database *also* don't have a residual amount.

pedrobaeza:
Then what is the final verdict?

StefanRijnhart:
This PR proposes to exclude lines with a zero balance even if the residual amount is not zero. Without this change, lines with a zero residual amount are already excluded. I checked @JordiBForgeFlow's journal entries in the 14.0 runbot database and they don't have a residual amount so this change does not apply to those entries. I have not checked the entries in the 13.0 runbot database and I can't find them back now, but if they also don't have a residual amount then the entries that this PR attempts to deal with are not reproducible on standard Odoo and then we don't know if we are covering up for some freak accident or a custom module gone wrong.

But if Jordi can reproduce on 13.0 again and check the residual amounts, that would be good. If the behaviour occurs there, we know it's a bug that was fixed in 14.0 and then we can attempt to have it fixed in 13.0 and meanwhile consider merging this cover-up.

pedrobaeza:
Ok, thanks for the explanation. Waiting then for Jordi's answer.

JordiBForgeFlow:
@StefanRijnhart find attached a video that reproduces the behaviour in 13.0. 
Tested here: https://8514978-13-0-all.runbot54.odoo.com/web?debug=1#action=547&model=account.move.line&view_type=list&cids=1&menu_id=384

Peek 2021-06-25 10-49

StefanRijnhart:
@JordiBForgeFlow Here too, you generate exchange rate lines with a zero balance but without a residual amount (I checked an export of these lines in the runbot instance you linked to), while this PR addresses lines with zero balance but *with* a residual amount (because all lines without a residual amount are already excluded regardless of the balance).

PR #762, opened 2021-03-29
[14.0][MIG] partner_statement: Migration to 14.0

RujiaL:

PR #793, opened 2021-06-03
[14.0][MIG] partner_statement: Migration to 14.0

victoralmau:
Migration to 14.0

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT30008


partner_statement/__manifest__.py
```diff
@@ -3,16 +3,17 @@
 
 {
     "name": "Partner Statement",
-    "version": "13.0.1.2.0",
+    "version": "14.0.1.0.0",
     "category": "Accounting & Finance",
     "summary": "OCA Financial Reports",
-    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "author": "ForgeFlow, Tecnativa, Odoo Community Association (OCA)",
```
pedrobaeza, line 9: No co-authorship should be granted in this simple migration

partner_statement/tests/test_outstanding_statement.py
```diff
@@ -51,14 +51,13 @@ def test_customer_outstanding_statement(self):
 
         statement = wiz_id.button_export_pdf()
 
-        self.assertDictContainsSubset(
+        self.assertGreaterEqual(
```
pedrobaeza, line 54: Please don't change this, as it doesn't work the same way.

partner_statement/tests/test_activity_statement.py
```diff
@@ -55,14 +55,13 @@ def test_customer_activity_statement(self):
 
         statement = wiz_id.button_export_pdf()
 
-        self.assertDictContainsSubset(
+        self.assertGreaterEqual(
```
pedrobaeza, line 58: Please don't change this, as it doesn't work the same way.

partner_statement/report/report_statement_common.py
```diff
@@ -66,7 +69,7 @@ def _show_buckets_sql_q1(self, partners, date_end, account_type):
                 WHERE l2.date <= %(date_end)s
             ) as pc ON pc.credit_move_id = l.id
             WHERE l.partner_id IN %(partners)s
-                                AND l.account_internal_type = %(account_type)s
+                                AND l_account_type.type = %(account_type)s
```
pedrobaeza, line 72: Why adding extra table? I think the field is still available, at least on `account_account`, isn't it?

partner_statement/report/outstanding_statement.py
```diff
@@ -57,7 +60,7 @@ def _display_lines_sql_q1(self, partners, date_end, account_type):
                 WHERE l2.date <= %(date_end)s
             ) as pc ON pc.credit_move_id = l.id
             WHERE l.partner_id IN %(partners)s
-                                AND l.account_internal_type = %(account_type)s
+                                AND l_account_type.type = %(account_type)s
```
pedrobaeza, line 63: The same as the other query

partner_statement/report/activity_statement.py
```diff
@@ -107,9 +110,11 @@ def _display_lines_sql_q1(self, partners, date_start, date_end, account_type):
                 END as date_maturity
             FROM account_move_line l
             JOIN account_move m ON (l.move_id = m.id)
+            JOIN account_account l_account ON l.account_id = l_account.id
+            JOIN account_account_type l_account_type ON l_account.user_type_id = l_account_type.id
             JOIN account_journal aj ON (l.journal_id = aj.id)
             WHERE l.partner_id IN %(partners)s
-                AND l.account_internal_type = %(account_type)s
+                AND l_account_type.type = %(account_type)s
```
pedrobaeza, line 117: The same as the other query

pedrobaeza (changes requested):

pedrobaeza:
And please squash together the 3 commits for Spanish (Argentina) translation.

gdgellatly:
Why do this and ignore the open pull request?

pedrobaeza:
We didn't know about the other. Is it in better condition then this one? Can you point to it?

gdgellatly:
https://github.com/OCA/account-financial-reporting/pull/762 As for better condition, well it has a lot more manual testing but still I feel we haven't quite got the currency part correct. There are some minor fixes needed as well but for single currency it works well.

We have a load of downstream dependent modules which automate a lot of the reports so as those are migrated this kind of gets extra tested.

pedrobaeza:
Closing in favor of #762

PR #799, opened 2021-06-16
[13.0][FIX] partner_statement: consider a multicurrency case

MiquelRForgeFlow:
Detected cases where account_partial_reconcile has null currency_id, thus giving falsy amount_currency, and thus making some reconciled invoices appearing in the outstanding. Used [_amount_residual](https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L3289) of account move lines as reference to fix it.

Hope you enjoy :)


JordiBForgeFlow (approved):
:+1: Tested. Works as expected.

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: Functional review

MiquelRForgeFlow:
This is ready, finally.

PR #802, opened 2021-07-07
[13.0][FIX+IMP] mis_builder_cash_flow: Proper SQL expression + default for v13 + target moves

pedrobaeza:
- On v13, invoices are journal entries, so we have to explicitly exclude cancelled invoices from the expression. We leave draft ones, as they are good for forecasting.
- Default value for the company of the forecast line doesn't follow new specification, putting an incorrect value.
- Extra improvement: avoid a non needed casting of the fixed column type.

@Tecnativa TT30828


pedrobaeza:
Included improvement for taking into account "Target Moves" option.

rvalyi (approved):

joao-p-marques (approved):

victoralmau (approved):

PR #804, opened 2021-07-13
[13.0][partner_statement][fix] group by move line id in outstanding partner balance

JordiBForgeFlow:
Otherwise in some cases the results are incorrect. In reality we do the grouping because a move line can have multiple partial reconciliations, and we sum them up into each move line.


MiquelRForgeFlow (approved):
LGTM

StefanRijnhart (approved):

PR #805, opened 2021-07-15
[12.0][FIX] account_financial_report: limit scope of Trial Balance query within allowed account groups

astirpe:
This proposal is to fix a problem encountered while working on a customization project that required field company_id being added on account.group. Notice that such solution was also included in standard Odoo 14.0.
The problem is an access rights error that occurs because of the Trial Balance report query that actually reads all the existing account groups without respecting the read rights permissions.
This proposed solution first gets all the accessible records of account.group model and then use that id list in the query.

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -391,6 +391,9 @@ AND ra.report_id = %s
 
     def _inject_account_group_values(self):
         """Inject report values for report_trial_balance_account"""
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

HaraldPanten (approved):
Seems reasonable

StefanRijnhart (approved):
Disclaimer: approving as a team mate of Andrea on this particular project.

PR #807, opened 2021-08-09
[14.0][FIX] account_financial_report: fix method parameter order

ferran-S73:
Fixes the order of the values passed to write_ending_balance_from_dict()

### DIFF
--- a/account_financial_report/report/open_items_xlsx.py
+++ b/account_financial_report/report/open_items_xlsx.py
@@ -185,8 +185,8 @@ class OpenItemsXslx(models.AbstractModel):
                 self.write_ending_balance_from_dict(
                     accounts_data[account_id],
                     type_object,
-                    report_data,
                     total_amount,
+                    report_data,
                     account_id,
                 )
 
--- a/account_financial_report/report/open_items_xlsx.py
+++ b/account_financial_report/report/open_items_xlsx.py
@@ -160,8 +160,8 @@ class OpenItemsXslx(models.AbstractModel):
                             type_object,
                             total_amount,
                             report_data,
-                            account_id,
-                            partner_id,
+                            account_id=account_id,
+                            partner_id=partner_id,
                         )
 
                         # Line break
@@ -187,7 +187,7 @@ class OpenItemsXslx(models.AbstractModel):
                     type_object,
                     total_amount,
                     report_data,
-                    account_id,
+                    account_id=account_id,
                 )
 
                 # 2 lines break


pedrobaeza (commented):
I see that the overwriting of `write_ending_balance_from_dict` method is not passing upstream the keyword arguments. Can you fix that as well?

ferran-S73:
> I see that the overwriting of `write_ending_balance_from_dict` method is not passing upstream the keyword arguments. Can you fix that as well?

I've added keywords to both calls of this function and to it's super call too on a second commit.

account_financial_report/report/open_items_xlsx.py
```diff
@@ -212,5 +212,5 @@ def write_ending_balance_from_dict(
             my_object["residual"] = total_amount[account_id]["residual"]
             label = _("Ending balance")
         super(OpenItemsXslx, self).write_ending_balance_from_dict(
-            my_object, name, label, report_data
+            my_object=my_object, name=name, label=label, report_data=report_data
```
pedrobaeza: First four arguments are not keyword ones, so you have to pass them without name. But this is still lacking the 2 keyword ones:
```suggestion
            my_object, name, label, report_data, account_id=account_id, partner_id=partner_id,
```

ferran-S73: I don't think it's super function is expecting any keyword arguments. Keep in mind that it's firstly declared with only 4 arguments and that they're not the same as the inherited one.
On AbstractReportXslx it's declared as `def write_ending_balance_from_dict(self, my_object, name, label, report_data):` whereas on OpenItemsXslx it is declared as `def write_ending_balance_from_dict(self, my_object, type_object, total_amount, report_data, account_id=False, partner_id=False):`

pedrobaeza: Are you telling me that this is breaking inheritance adding extra kw-args to the method? (not your fault of course) That is going to fail when having other module inheriting the same method and not passing such keyword arguments. Method arguments should be constant on all inheritance chain, or have a fallback `**kwargs` for that.

ferran-S73: This function is inheriting the one declared on the `Abstract` class of the same module [AbstractReportXslx](https://github.com/OCA/account-financial-reporting/blob/14.0/account_financial_report/report/abstract_report_xlsx.py#L447).

The original function is expecting 4 arguments and every class that inherits this one on the module (there's quite a few of them) overwrite the expected arguments. In order to not cause any error each implementation calls the parent function with only the four original arguments regardless of how many arguments the child function implements.

I'm not sure this is the best practice either but that's how is done in this module and is why I didn't add the extra keyword arguments in the first place.

pedrobaeza: OK, then let's keep it for now, but it's not correct to pass positional arguments as keyword arguments, so please revert that change and let's revert to the original change you do. Sorry for the generated noise, that came from that awkward situation with the asymmetric arguments.

ferran-S73: I've removed the keyword for non-keyword arguments but left it for the ones that are keyword.

pedrobaeza (approved):
Thanks for doing the changes

pedrobaeza:
Let's fast-track the obvious patch:

/ocabot merge patch

Issue #808, opened 2021-08-21
[13.0][BUG] account_financial_report, error Aged Partner Balance with Show move line details


kittiu:
Just install module with demo database, and run Aged partner Balance report.

Affected version: 13.0.1.8.0 (latest)

![image](https://user-images.githubusercontent.com/1973598/130309539-fd22107c-e875-4ec9-97cf-7e32f298d86f.png)

![image](https://user-images.githubusercontent.com/1973598/130309546-5c6a899f-dc7a-424a-b9b7-249f48f323af.png)

![image](https://user-images.githubusercontent.com/1973598/130309336-23f60816-7959-4627-a204-6d581cee2d83.png)

SimoRubi:
Found the same issue on `14.0` and working on a fix, stay tuned :)

SimoRubi:
@kittiu can you check if https://github.com/OCA/account-financial-reporting/pull/811 fixes this issue and leave a review there? Thanks!

PR #809, opened 2021-08-24
[14.0][FIX] account_financial_report: Print report with details

SimoRubi:
**Steps to reproduce**
Same as https://github.com/OCA/account-financial-reporting/issues/808 but for `14.0`
1. Open Invoicing > Reporting > Aged Partner Balance
2. Check:
    - Show Move Line Details
    - Receivable Accounts Only
3. Click on View

**Actual behavior**
Error:
```
Error message:

Error to render compiling AST
AttributeError: 'dict' object has no attribute 'move_line_id'
Template: account_financial_report.report_aged_partner_balance_move_lines
Path: /t/div/t/div/div[1]/span
Node: <span t-att-res-id="line.move_line_id.id" res-model="account.move.line" view-type="form">
                            <!--## We don't use t-field because it throws an error on click -->
                            <t t-esc="line['date']" t-options="{'widget': 'date'}"/>
                        </span>
                    
The error occured while rendering the template account_financial_report.report_aged_partner_balance_move_lines and evaluating the following expression: <span t-att-res-id="line.move_line_id.id" res-model="account.move.line" view-type="form"> <!--## We don't use t-field because it throws an error on click --> <t t-esc="line['date']" t-options="{'widget': 'date'}"/> </span>
```

**Desired behavior**
Show the Aged partner balance

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -218,9 +218,8 @@ class AgedPartnerBalanceReport(models.AbstractModel):
             "ref",
             "reconciled",
         ]
-        move_lines = self.env["account.move.line"].search_read(
-            domain=domain, fields=ml_fields
-        )
+        line_model = self.env["account.move.line"]
+        move_lines = line_model.search_read(domain=domain, fields=ml_fields)
         journals_ids = set()
         partners_ids = set()
         partners_data = {}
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why not fixing it in 13.0?

SimoRubi:
> Why not fixing it in 13.0?

I found the bug in one of our `14.0` instances so I need this fix.
Then I noticed the bug existed in `13.0` too thanks to https://github.com/OCA/account-financial-reporting/issues/808.

I don't like to have the same code implemented in two open PRs because any comment/fix implies double work on both PRs.
I'll wait for this to be merged and then I'll be happy to cherry-pick the merged commit from `14.0` into `13.0` (in a new PR).

pedrobaeza:
OK, good for me.

pedrobaeza (approved):

tafaRU (approved):

HaraldPanten (approved):
Functional review

Issue #815, opened 2021-09-08
[14.0][BUG] account_financial_report : issue with columns of first line


tv-openbig:
There is an issue with the first line representing the column sums of the fields debit/credit.
As you can see in the following screenshot there is a wrong offset of a colum, breaking the table of the report. 
![Bildschirmfoto 2021-09-08 um 18 36 42](https://user-images.githubusercontent.com/8348598/132549852-53bc48a5-59a2-4770-a8ca-1ecd1f155bbe.png)

This happens with the Journal Ledger and Trial Balance.

pedrobaeza:
Steps to reproduce it?

tv-openbig:
> Steps to reproduce it?

1.) Invoicing App / Reporting / OCA accounting reports / Journal Ledger
2.) Select a start/end date or a pre-defined date_range (with account moves for sure)
3.) Click on "View"

PR #818, opened 2021-09-22
[14.0][FIX] account_financial_report: Filter column creation to have correct offset in Journal Ledger

joao-p-marques:
Fixes https://github.com/OCA/account-financial-reporting/issues/815
Seems like a missing change from https://github.com/OCA/account-financial-reporting/commit/0e3b4d29722b9018b58d611b16af945279790265

@tv-openbig I could only reproduce the error in the Journal Ledger. Let me now the appropriate steps to reproduce it in other reports if that is the case.

@Tecnativa
ping @pedrobaeza @ValentinVinagre


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this:

/ocabot merge minor

PR #819, opened 2021-10-04
[14.0][IMP] account_tax_balance: use rename_fields instead of rename_columns

MiquelRForgeFlow:
The rename_fields method is cleaner, because it handles ir_model_fields and translations.


JordiBForgeFlow (approved):
:+1: Tested. Works as expected

pedrobaeza (approved):

PR #768, opened 2021-04-13
14.0 mig mis builder cash flow

dreispt:
- [12.0][ADD] mis_builder_cash_flow module
- [IMP] mis_builder_cash_flow: Add tests
- [UPD] Update mis_builder_cash_flow.pot
- [UPD] README.rst
- [FIX] cash_flow: forecast accumulate next columns
- Added translation using Weblate (Portuguese)
- Added translation using Weblate (Catalan)
- Added translation using Weblate (Spanish)
- [IMP] include partner_id in mis builder cash flow
- [IMP] partner_id in forecastline
- [UPD] Update mis_builder_cash_flow.pot
- [UPD] README.rst
- mis_builder_cash_flow 12.0.1.3.0
- Update translation files
- Translated using Weblate (Portuguese)
- [IMP] mis_builder_cash_flow: black, isort, prettier
- [MIG] mis_builder_cash_flow: Migration to 13.0
- [UPD] Update mis_builder_cash_flow.pot
- [UPD] README.rst
- Update translation files
- Added translation using Weblate (Spanish (Argentina))
- Translated using Weblate (Spanish (Argentina))
- Update translation files
- Added translation using Weblate (Dutch)
- Translated using Weblate (Dutch)
- Translated using Weblate (Spanish)
- [IMP] mis_builder_cash_flow: black, isort, prettier
- [MIG] mis_builder_cash_flow: Migration to 14.0

PR #821, opened 2021-10-16
[14.0][mis_builder_cash_flow] mig mis_builder_cash_flow

rvalyi:
An newer version of https://github.com/OCA/account-financial-reporting/pull/768 with:

* the latest v13.0 changes from Tecnativa https://github.com/OCA/account-financial-reporting/pull/802
* a fix for the offending mis_cash_flow_search_view of #768: assuming the balance filter never worked I simply removed it. In the future one could bother ensuring debit and credit are not null, but as worked without such filter so far I preferred drop the offending balance filter.
* a fix to make tests pass: type -> move_type for the account.move object.


pedrobaeza (approved):
If presumably goes green in Travis

marcelsavegnago (approved):
LGTM

PR #823, opened 2021-10-19
[14.0][ADD] mis_template_financial_report

StefanRijnhart:
Migrated from https://github.com/OCA/l10n-netherlands/tree/12.0/l10n_nl_mis_reports by @hbrunn as there is nothing specific to the Dutch localization in this module, apart from the horizontal rendering which I've made optional.


mis_template_financial_report/hooks.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Opener B.V. <stefan@opener.amsterdam>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+def pre_init_hook(cr):
+    cr.execute(
+        """
+        UPDATE ir_model_data SET module = 'mis_template_financial_report'
```
pedrobaeza: Shouldn't this be a migration script?

StefanRijnhart: When using OpenUpgrade, no script is necessary at all once we've put the rename in apriori.py. But this might help people who do not use OpenUpgrade, for which this would be a fresh install (on which Odoo does not run the migration scripts).

pedrobaeza: Well, I think you should add it the same to pre-migration script instead of pre-init hook. This will work the same with OpenUpgrade or with enterprise upgrade, but we are not mixing tools.

pedrobaeza: OK, I think you are trying to do people to install this module without renaming? But that is leading to failure, as it depends if they haven't uninstalled the previous module yet, and even that, they may have errors.

StefanRijnhart: Running with a missing module like this without uninstalling does not lead to errors. But as the number of people who would benefit from the script would be quite low, I don't mind removing it entirely. @hbrunn do you have an opinion on the matter?

hbrunn: I see OpenUpgrade as the de facto standard for OCA based migrations, so I agree with @pedrobaeza on this one

hbrunn (approved):
yes, makes sense to put this here. In some later iteration we'll pull the horizontalness to the base module as discussed in https://github.com/OCA/mis-builder/issues/307, https://github.com/OCA/mis-builder/issues/109, but I don't see a problem with having it here for the time being

mis_template_financial_report/models/mis_report_instance.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class MisReportInstance(models.Model):
+    _inherit = "mis.report.instance"
+
+    allow_horizontal = fields.Boolean(compute="_compute_allow_horizontal")
+    horizontal = fields.Boolean()
+
+    @api.depends("report_id")
+    def _compute_allow_horizontal(self):
```
pedrobaeza, line 16: Shouldn't this be in an isolated module or part of the main one?

mis_template_financial_report/hooks.py
StefanRijnhart: Thanks for your input. Removed the pre-init hook.

mis_template_financial_report/models/mis_report_instance.py
StefanRijnhart: As apparent from the issues that @hbrunn linked to, that is a complex topic. I've mentioned it here now https://github.com/OCA/mis-builder/issues/389 in the context of the 15.0 migration of mis_builder. Out of scope for this version I believe.

pedrobaeza: OK then

PR #842, opened 2021-12-28
[FIX] account_tax_balance: Fix development_status

ValentinVinagre:


ValentinVinagre:
@pedrobaeza @HaraldPanten  :P

HaraldPanten (approved):

pedrobaeza (approved):
Module is there since 9.0 and have tests, so let's promote it as deserved.

PR #846, opened 2021-12-30
[FIX][14.0] partner-statement: Remove need to groupby on amount_currency in Activity Statement

gdgellatly:
By aggregating before the case statement there is no need to group by amount_currency.

In earlier versions of Odoo this bug only presented in multicurrency transactions as company currency transactions
would have an amount_currency of 0.0 and group by 0.0 just fine.

In later versions of Odoo amount_currency is always set and this causes otherwise identical move lines
that should be summed to display seperately. However, amount_currency was required in groupby because it
was part of a non aggregated case statement with the sum inside.

By reversing the nesting we remove the need to group by amount_currency. This is safe because we already group by
currency_id and aggregates are not evaluated until after group by completes.

### DIFF
--- a/partner_statement/report/activity_statement.py
+++ b/partner_statement/report/activity_statement.py
@@ -18,14 +18,14 @@ class ActivityStatement(models.AbstractModel):
             self._cr.mogrify(
                 """
             SELECT l.partner_id, l.currency_id, l.company_id,
-            CASE WHEN l.currency_id is not null AND l.amount_currency > 0.0
-                THEN sum(l.amount_currency)
-                ELSE sum(l.debit)
-            END as debit,
-            CASE WHEN l.currency_id is not null AND l.amount_currency < 0.0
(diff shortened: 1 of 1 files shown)


gdgellatly:
In further testing of side by side comparisons of a v14.0 and v12.0 database (migrated via enterprise), initial balances were also incorrect due to the group by clause. Force pushed to update with the same rationale. With this change both statements are now identical.

gdgellatly:
This has gone through over 40 statement runs now across multiple companies and not skipped a beat. Now that I think about, when I rewrote partner-statement back in v12 I added that group by to solve an issue (the must appear in GROUP BY error), it was never necessary, it was just my bad SQL.

AaronHForgeFlow (approved):
Code review LG!

MiquelRForgeFlow (approved):
Shouldn't this be backported?

MiquelRForgeFlow:
Can we merge this?

PR #850, opened 2022-01-15
[14.0][FIX] account_financial_report: domain detail move line on target move

Saran440:
Step to error:
1. Reset to draft 1 invoice and filter trail balance report with target move = all posted. it show amount is correct

Selection_001

2. click on line `debit` or else. it will related on account.move.line BUT it show all entries not filter state. So, it not equal report and details
Selection_002

### DIFF
--- a/account_financial_report/readme/CONTRIBUTORS.rst
+++ b/account_financial_report/readme/CONTRIBUTORS.rst
@@ -30,6 +30,7 @@
   * Valentin Vinagre
 
 * Lois Rilo <lois.rilo@forgeflow.com>
+* Saran Lim. <saranl@ecosoft.co.th>
 
 Much of the work in this module was done at a sprint in Sorrento, Italy in
 April 2016.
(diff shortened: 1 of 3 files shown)


rafaelbn:
Hello @LoisRForgeFlow @ValentinVinagre could you please take a look in this one?

Shide (approved):

rafaelbn (approved):
Thanks @Saran440 👍

Saran440:
Updated travis is red (no change code.)

pedrobaeza:
This seems an ugly patch. Why not doing this in a centralized place? Anyways, this has a problem: cancelled moved should never be included, no matter the selected option.

Saran440:
This module generates a domain from an xml file and passes it to the java script.
On centralized, I'm not sure how to fix it, please suggest me.

pedrobaeza:
OK if it can be centralized, but please include about the cancel state.

PR #853, opened 2022-01-27
[13.0][FIX] account_financial_report: open_items

AaronHForgeFlow:
Fixed the amount_residual_currency on partial reconcile

The amount_residual_currency shown is now is the actual amount_residual_currency, it is not taking into account the partial reconciled records.

@ForgeFlow

Updated:

In order to reproduce the error please do:

1. Create an invoice for a different currency in 2021
2. Create a payment in 2022
3. Run the open items report at the end of 2021

The amount residual is correct but the amount residual currency is incorrect

### DIFF
--- a/account_financial_report/report/abstract_report.py
+++ b/account_financial_report/report/abstract_report.py
@@ -90,11 +90,29 @@ class AbstractReport(models.AbstractModel):
         for move_line in move_lines:
             ml_id = move_line["id"]
             if ml_id in debit_ids:
-                move_line["amount_residual"] += debit_amount[ml_id]
-                move_line["amount_residual_currency"] += debit_amount_currency[ml_id]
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Need this or the aged_partner_balance will may fail: https://github.com/ForgeFlow/account-financial-reporting/pull/5 Will include that PR here soon.

rafaelbn:
Hello @AaronHForgeFlow , please could you update this PR?

JordiMForgeFlow (approved):

pedrobaeza (changes requested):
You are modifying the arguments of a function in a stable version. This will lead to problems in dependent modules. And there's already `amount_residual_currency` field in standard AFAIK

PR #857, opened 2022-02-01
[13.0][FIX] partner_statement: activity statement initial balance

AaronHForgeFlow:
Currently, the initial balance taken is the last balance of the last line for the currency.

cc @ForgeFlow


MiquelRForgeFlow:
@gdgellatly

gdgellatly:
@MiquelRForgeFlow I have a suspicion that the fix I did for 14.0 that we merged yesterday actually solves this issue.

MiquelRForgeFlow:
@AaronHForgeFlow could you check if in branch 14.0 do you have same error?

PR #858, opened 2022-02-04
[15.0][FIX] account_tax_balance: disable readme pylint in migrations folder

mariadforgeflow:
Disable missing-readme pylint check so other modules' pre-commit doesn't fail.


pedrobaeza:
I think you have an old version of pylint in your local, and that's why you get that problem, but the CI was green before this.

mariadforgeflow:
Hi @pedrobaeza! I've just rebased in my local and I still get the missing-readme error in pre-commit. Is there any other thing I can do? Thank you for your time:)

pedrobaeza:
But the CI was previously red here on GitHub before this?

MiquelRForgeFlow:
@pedrobaeza @AaronHForgeFlow ready to merge

MiquelRForgeFlow (approved):

PR #862, opened 2022-02-15
[14.0][FIX] account_financial_report: Run compute in multi

hildickethan:
- Run computes in batch
- Added recursive to the compute

### DIFF
--- a/account_financial_report/models/account_group.py
+++ b/account_financial_report/models/account_group.py
@@ -20,26 +20,34 @@ class AccountGroup(models.Model):
         string="Compute accounts",
         store=True,
     )
-    complete_name = fields.Char("Full Name", compute="_compute_complete_name")
-    complete_code = fields.Char("Full Code", compute="_compute_complete_code")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
There's a new attribute `recursive=True` in the ORM that would benefit these fields computation. Can you please check?

hildickethan:
> There's a new attribute `recursive=True` in the ORM that would benefit these fields computation. Can you please check?

Added it, it seemed to work fine. I presume it's to compute the parents `complete_name` too?

pedrobaeza (approved):
`recursive` helps to process the compute in the best order for performance and completeness.

hildickethan:
> `recursive` helps to process the compute in the best order for performance and completeness.

Nice, is there any documentation for this new attribute? I didn't see it in either the 14.0 or 15.0 docs

pedrobaeza:
I'm afraid not.

MiquelRForgeFlow (approved):

PR #863, opened 2022-02-17
14.0 fix null error partner statement xlsx

gdgellatly:
The excel logic in determining name_to_show is wrong. 

By using get everywhere, but with the value gaurenteed to be in the dict we can get NULL values. This cause the report to fail. Moreover the prior reference to name in ref and then setting to name is also incorrect. If the name was in the ref we should use the ref. Furthermore, the name is simply the move_name in default Odoo invoices and this is included 2 columns prior. It therefore makes sense to display the ref if available, or else the name or else the empty string. This hugely simplifies the code as well.

Depends #846 - once that is merged I can tidy up this commit history.

### DIFF
--- a/partner_statement/report/outstanting_statement_xlsx.py
+++ b/partner_statement/report/outstanting_statement_xlsx.py
@@ -60,11 +60,11 @@ class OutstandingStatementXslx(models.AbstractModel):
                 if not line.get("ref", ""):
                     name_to_show = line.get("name", "")
                 else:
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@gdgellatly can you rebase to rebuid commit history?

AaronHForgeFlow (approved):

MiquelRForgeFlow (approved):

rafaelbn (approved):

PR #866, opened 2022-02-23
[14.0][FIX] account_financial_report: Trial Balance with 'hide account at 0'

ernestotejeda:
Cc @Tecnativa TT34490 TT35127

Fix 'Trial Balance' report: The 'Trial Balance' report shows accounts at 0 even if 'Hide accounts at 0' option is checked.

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -5,6 +5,7 @@
 
 
 from odoo import api, models
+from odoo.tools.float_utils import float_is_zero
 
 
 class TrialBalanceReport(models.AbstractModel):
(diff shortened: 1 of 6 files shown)


ernestotejeda:
ping @pedrobaeza @sergio-teruel @carlosdauden  @JoanSForgeFlow

ernestotejeda:
ping @pedrobaeza

account_financial_report/report/trial_balance.py
```diff
@@ -300,6 +300,29 @@ def _compute_partner_amount(
                         ] += round(tb["amount_currency"], 2)
         return total_amount, partners_data
 
+    def _remove_accounts_at_cero(self, total_amount, show_partner_details):
+        def is_removable(d):
+            return (
+                d["initial_balance"] == 0
```
pedrobaeza: This should be used with `float_is_zero` (and if not possible to get decimal precision, put 8 digits).

account_financial_report/report/trial_balance.py
```diff
@@ -300,6 +300,29 @@ def _compute_partner_amount(
...
+            )
+
+        accounts_to_remove = []
+        for acc_id, ta_data in total_amount.items():
+            if is_removable(ta_data):
+                accounts_to_remove.append(acc_id)
+            elif show_partner_details:
+                partner_to_remove = []
+                for key, value in ta_data.items():
+                    if isinstance(key, int) and is_removable(value):
```
pedrobaeza, line 324: Why do you need to do this `isinstance` thing? Always explain as comment the weird things.

ernestotejeda: I have put a comment in the code, although it is difficult for me to explain. The image below shows the value of the total_amount variable (the main dict used for printing the report) while debugging the code in a use case. If the show_partner_details option is checked, in addition to the "initial_balance" ... "ending_balance" keys (marked in red), the same account data dictionary will also have the partner IDs as key (marked in blue) with their respective data. That's why i used isinstance method. Check the comment i have put.

image

And this is the result with this data and without my code:

image

Shide (approved):

Shide:
@pedrobaeza are you ok with the new changes?

rafaelbn (approved):
Tested functionally 👍 

https://github.com/OCA/account-financial-reporting/blob/14.0/account_financial_report/wizard/trial_balance_wizard.py#L55-L57

```
    hide_account_at_0 = fields.Boolean(
        string="Hide accounts at 0",
        default=True,
        help="When this option is enabled, the trial balance will "
        "not display accounts that have initial balance = "
        "debit = credit = end balance = 0",
    )
```

Shide:
@pedrobaeza are you ok with the changes (comments) of @ernestotejeda ?

pedrobaeza (changes requested):
It's not working properly yet.

ernestotejeda:
@pedrobaeza could please you check again with this new commit?

pedrobaeza:
@ernestotejeda can you remove the "computed hierarchy" option as now the hierarchy is already computed in the records and no need to compute them on the fly?

ernestotejeda:
> @ernestotejeda can you remove the "computed hierarchy" option as now the hierarchy is already computed in the records and no need to compute them on the fly?

@pedrobaeza change done

ernestotejeda:
Travis is red, i will check it tomorrow

ernestotejeda:
@pedrobaeza travis is green now

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to v15

PR #867, opened 2022-02-25
[FIX] [14.0] Removed external dependencies already defined in Odoo base

Shide:
Removed external dependencies already defined in Odoo base

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L40

MT-282 @moduon


yajo (approved):

pedrobaeza (changes requested):
Remove it instead of changing it, as it's part of Odoo standard requirements

pedrobaeza (approved):

PR #875, opened 2022-04-26
[12.0][FIX] account_financial_report: Speed up open items

hbrunn:
before, we got the Cartesian product of move lines and partners in the `only_empty_partner_line=True` case, which is potentially huge and here grinds the database to a halt. Thanks @StefanRijnhart for pushing me in the right direction.

The sorting also cost a few percent performance, so I removed that too as I don't see how it's helpful.

### DIFF
--- a/account_financial_report/report/open_items.py
+++ b/account_financial_report/report/open_items.py
@@ -394,6 +394,9 @@ FROM
                 ) AS partial_amount_currency,
                 ml.currency_id
             FROM
+        """
+        if not only_empty_partner_line:
+            sub_query += """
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):

rafaelbn (approved):

LuqueDaniel (approved):

rafaelbn:
Hello @StefanRijnhart , I think you could merge this one.

Just relate this PR with another about "performance": https://github.com/OCA/account-financial-reporting/pull/660

PR #798, opened 2021-06-14
[14.0][IMP] account_financial_report: added 'auto sequence' option in Journal Ledger

ValentinVinagre:
@pedrobaeza  forward-port in v14 from https://github.com/OCA/account-financial-reporting/pull/787

PR #787, opened 2021-05-26
[13.0][IMP] account_financial_report: added 'auto sequence' option in Journal Ledger.

ValentinVinagre:
Added 'auto sequence' option in Ledger.
This option is necessary because in v.13 the Journal Ledger only shows the invoice number, but does not have a correlative sequence with versions prior to v.13.
The size of columns has also been changed for this new value.

PR #876, opened 2022-04-29
[IMP] account_financial_report: let store auto-generated legal sequence number

yajo:
- Let accountants choose to store auto-generated sequence.
- Let them use that sequence number in search, form and tree views of account.move.
- Fix one bug that was forbidding rendering report.

With this improvement, accountants will be able to find data when referenced by the official legal number.

@moduon MT-672


JordiBForgeFlow:
@yajo thanks for the proposal. I suggest to add a module "account_financial_report_auto_sequence". This feature is not used anywhere but in Spain, as far as I can see, and many countries use the financial reports nowadays. They will be very confused with this feature.

pedrobaeza:
I don't think this is needed at all, as it's for a very specific requirement. You can have the PDF sent to the "Registro Mercantil" in case there's any audit to match one number with the other, but most of the checks can be done with the real number and ignore the other. Take into account that the progressive and consecutive numbering is a requirement from the old 1887 merchant code.

yajo:
OK, I'll add it in a separate module.

rafaelbn:
@JordiBForgeFlow , do you mean this PR should be also taken out from this module? Thank you

[[13.0][IMP] account_financial_report: added 'auto sequence' option in Journal Ledger. #787](https://github.com/OCA/account-financial-reporting/pull/787)

[[14.0][IMP] account_financial_report: added 'auto sequence' option in Journal Ledger](https://github.com/OCA/account-financial-reporting/pull/798) #798

JordiBForgeFlow:
@yajo I did not notice those PR's... I would have preferred to have those also exist in a separate module, but they are now merged.

I can see that in your case the split is rather simple

pedrobaeza:
Please don't ignore my comments. IMO everything should stay as is. No change is needed.

Shide:
I think this change makes Accountants job easy when searching for specific account moves on an audit.
The audit person may only have the "legal" number (000123) and not the Odoo one. And avoids to store the Excel file or the PDF file forever. You are sure that the presented "legal" number is stored on the database.

Shide (approved):

yajo:
I have made the split, now all is in a separate module (except the bugfix). There's still some work to do so I'll set it as draft for now.

rafaelbn:
Hello! 😄 

We don't ignore anyone, in fact we have split it as @yajo  said in https://github.com/OCA/account-financial-reporting/pull/876#issuecomment-1113231274

- a FIX of the module
- a new module `account_financial_report_remember_sequence` (name was temporally and that is why **this PR is Draft**)

In fact, I agree with @JordiBForgeFlow https://github.com/OCA/account-financial-reporting/pull/876#issuecomment-1113150421

About if is needed or not, there are many modules in OCA with the need of different clients and any one can choose the module the use.

Said that, and as Jordi said. Including [these PR](https://github.com/OCA/account-financial-reporting/pull/876#issuecomment-1113154733) on a global module is done and we cannot change that, but an externa module `account_financial_report_auto_sequence` would be better (only) in my opinion.

Thank you for all you opinions 🙏 👍 . We sleep it over (and ask again the auditors are requesting it) to develop the best option.

Best regards

pedrobaeza:
Ok if you want to go ahead with the storage not breaking existing things. I tried to explain why I think it's not needed to do it, but of course you decide.

account_financial_report/report/templates/journal_ledger.xml
```diff
@@ -306,10 +306,7 @@
                 </t>
             </div>
             <div class="act_as_cell amount" name="amount_currency">
-                <t
-                    t-if="move_line['amount_currency']"
-                    t-options="{'widget': 'float', 'precision': 2}"
-                >
+                <t t-if="move_line['amount_currency']">
```
ValentinVinagre, line 309: Why did you delete `t-option`?

account_financial_report_remember_sequence/report/journal_ledger.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "report.account_financial_report.journal_ledger"
+
+    def _get_move_lines_data(self, ml, wizard, ml_taxes, auto_sequence, exigible):
+        """Remember auto_sequence if required."""
+        result = super()._get_move_lines_data(
+            ml, wizard, ml_taxes, auto_sequence, exigible
+        )
+        if wizard.with_auto_sequence and wizard.store_auto_sequence:
+            move = self.env["account.move"].browse(result["move_id"])
+            move.legal_sequence = auto_sequence
```
ValentinVinagre, line 17: If an "account.move" has N "account.move.line", aren't you overwriting "legal_sequence" field N times?

ValentinVinagre (commented):
Some Comments.

HaraldPanten (commented):
Do you want to keep it as a sequence? Don't you prefer to keep the number as it is? I mean... "00003" instead of just "3"?


Captura de Pantalla 2022-04-29 a las 17 02 30

account_financial_report/report/templates/journal_ledger.xml
yajo: It was giving a qweb exception while printing the report. It didn't have any sense, because it's just used for formatting, but no output is ever made by a `<t t-if` node, so it was an obvious bug.

account_financial_report_remember_sequence/report/journal_ledger.py
yajo: Yes, but since v13 it's a cheap operation.

yajo:
We proposed this as a MVP to fix the problem expressed in https://github.com/OCA/account-financial-reporting/pull/876#issuecomment-1113199549. However we examined the issue with more depth and found out that the MVP is just not enough for our use case, so I'll close this PR while we get to a better solution.

We initially thought this was a global need due to https://github.com/OCA/account-financial-reporting/pull/787 being included in the base module, but after what's been talked here, we'll do a separate module to avoid polluting the main one, so whoever don't need it won't have a problem.

Sorry for the noise and thanks everyone for your reviews!

PR #878, opened 2022-05-03
[13.0][FIX] *: Switch license + website

pedrobaeza:
To comply with constraints and have green CI

@Tecnativa


rafaelbn:
Should we change licenses of `mis_builder_cash_flow` from LGPL to AGPL in previous and future versions?

I don't know if the OCA bot detects errors with licenses dependencies like this one.

pedrobaeza:
It has to, as the base module mis_builder is AGPL

PR #879, opened 2022-05-05
[IMP] mis_template_financal_report: Create detailed PL/BS

ghost:
This PR will add a menuitem: Invoicing / Configuration / MIS Reporting / Generate detailed PL/BS
The method will create:
(mis.report) Profit & Loss, detailed
(mis.report) Balance Sheet, detailed
(mis.report.instance) Profit & Loss, last 2 years
(mis.report.instance) Balance Sheet, last 2 years

The report templates will have a KPI per account type and KPIs for totals.


hbrunn (changes requested):
I think you can better add those reports in a second module depending on this, something like `mis_template_financal_report_detailed`

PR #883, opened 2022-05-27
[14.0][FIX] account_financial_report: trial balance hierarchy

ernestotejeda:
Cc @Tecnativa TT35127

When you generate a Trial Balance report with account hierarchy, it may happen that the values of some accounts are not equal to the sum of their subaccounts. This PR aims to fix that.


pedrobaeza:
What is the relation of this with #866 ?

ernestotejeda:
Each one tries to fix different problem in the Trial Balance, but since they refer to the same report, I'm going to close that PR and put these changes in https://github.com/OCA/account-financial-reporting/pull/866.

PR #890, opened 2022-07-01
[15.0][FIX] Trial Balance Report: style could not be applied when i select hierarchical view

ieski:
Hi,

Trial Balance Report style could not be applied when i select hierarchical view.


ieski:
image

PR #891, opened 2022-07-05
[IMP] account_tax_balance

xavier-bouquiaux:
when generating an account.move on account.internal_type
and if the balance was = 0
the financial_type was receivale_refund in place of receivable


acsonefho (approved):

pedrobaeza:
I think that it doesn't matter, as the balance is 0, but being coherent, you have to change the same on payable.

xavier-bouquiaux:
@pedrobaeza  -> done <br/>
could we merge that ?

adrienpeiffer (approved):

PR #894, opened 2022-08-01
[13.0] [FIX] account_bank_reconciliation_summary: inconsistent string format in fr.po

alexAubin:
As a result, Odoo was crashing (with an empty stacktrace ?) when triggering the export

(This is my first PR on OCA, don't hesitate to complain if I'm missing anything :sweat_smile: )


pedrobaeza:
Although this is done normally on the translation platform, let's fast-track as evident crash problem:

/ocabot merge patch

PR #906, opened 2022-09-05
[13.0][FIX] account_financial_report: Add sorting to move line records in journal ledger.

victoralmau:
It is necessary to add the `move_id` field to make sure that the order of the records is correct (especially if we use auto-sequence).

@Tecnativa TT38775

### DIFF
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -84,7 +84,10 @@ class JournalLedgerReport(models.AbstractModel):
         return [("display_type", "=", False), ("move_id", "in", move_ids)]
 
     def _get_move_lines_order(self, move_ids, wizard, journal_ids):
-        return ""
+        """Add `move_id` to make sure the order of the records is correct
+        (especially if we use auto-sequence).
+        """
+        return "move_id"
 
     def _get_move_lines_data(self, ml, wizard, ml_taxes, auto_sequence):
         base_debit = (


account_financial_report/report/journal_ledger.py
```diff
@@ -179,6 +179,12 @@ def _get_move_lines(self, move_ids, wizard, journal_ids):
             self._get_move_lines_domain(move_ids, wizard, journal_ids),
             order=self._get_move_lines_order(move_ids, wizard, journal_ids),
         )
+
```
pedrobaeza: Don't add empty lines in the middle of a method.

pedrobaeza (changes requested):
Instead of this, you should simply order also by `move_id`

victoralmau:
Changes done and tested.

pedrobaeza (approved):

pedrobaeza:
@victoralmau please always indicate why something is done, not the what, or by default, the steps to reproduce the problem. In this case, why we are sorting by move_id, because if not, the auto-sequence is not being correctly displayed, having gaps and not sorted order.

sergio-teruel (approved):

victoralmau:
Commit message updated and added a docstring in the code to explain why.

pedrobaeza:
The first line of the commit message is still a bit long. I would put this:

```
[FIX] account_financial_report: Sort journal ledger by move_id

Without this sort, and using auto-sequence, the numbers of the journal
entries may appear unsorted on the journal ledger. Forcing at least this
minimum sorting, we assure the items are sorted.

TT38775
```

Note also the empty line above the TT for better viewing in IDEs.

About the docstrings, if they are multi-line, you have to put the closing quotes in a separate line, and be more concise:
```python
        """Add `move_id` to make sure the order of the records is correct
        (especially if we use auto-sequence).
        """
```

pedrobaeza (approved):

PR #898, opened 2022-09-01
[13.0][IMP] account_financial_report: Amount currency columns from general ledger.

victoralmau:
Changes done to general ledger:
- [x] Remove `Cur.` column (currency is now added in `Amount cur.`).
- [x] Add `Cumul cur.` column.

despues

Please @pedrobaeza and @sergio-teruel could you review it?

@Tecnativa TT38722

PR #910, opened 2022-09-08
[14.0][IMP] account_financial_report: Amount currency columns from general ledger.

victoralmau:
FWP from 13.0: https://github.com/OCA/account-financial-reporting/pull/898

Changes done to general ledger:
- [x] Remove `Cur.` column (currency is now added in `Amount cur.`).
- [x] Add `Cumul cur.` column.

divisa-pdf
divisa-excel

Please @pedrobaeza and @sergio-teruel could you review it?

@Tecnativa TT38722

### DIFF
--- a/account_financial_report/README.rst
+++ b/account_financial_report/README.rst
@@ -130,6 +130,7 @@ Contributors
   * Ernesto Tejeda
   * João Marques
   * Alexandre D. Díaz
+  * Víctor Martínez
 
 * `Sygel <https://www.sygel.es>`__:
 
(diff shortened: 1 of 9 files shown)


victoralmau:
In v14, the display of the data in the new columns changes:
ejemplo-v14

The currency is always set, therefore, the cell value is filled in, what changes do you think should be applied?

pedrobaeza:
What about displaying content only if the currency of the aml is different from the company currency?

victoralmau:
Added a new commit to show the value if the currency is different from the company's currency.

pedrobaeza (changes requested):
The accumulated currency column is summing everything, not only those with different currency.

victoralmau:
> The accumulated currency column is summing everything, not only those with different currency.

Ok, I see, that was only happening in PDF report, it's changed now.
divisa-pdf

pedrobaeza (approved):

alexis-via (approved):

PR #915, opened 2022-09-19
[13.0][IMP] account_financial_report: Add tax detail to the general ledger report

victoralmau:
Changes done in general ledger:
- [x] Add tax detail to the general ledger report.
- [x] Refactor code to add grouping
- [x] Add `tax_line_id` field to taxes column in general ledger.

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT38721

PR #920, opened 2022-09-21
[13.0][FIX] account_financial_report: Display the correct name of initial balance and ending balance in general ledger

victoralmau:
Display the correct name of initial balance and ending balance in general ledger (forgotten in https://github.com/OCA/account-financial-reporting/pull/915)

@Tecnativa TT38721

### DIFF
--- a/account_financial_report/i18n/account_financial_report.pot
+++ b/account_financial_report/i18n/account_financial_report.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 13.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please change "Contactos" > "Empresas"

victoralmau:
I had changed it before in weblate but OK, changed.

pedrobaeza:
Don't do such things, or conflicts will arise between both. The only way to avoid it is to push the translations immediately.

pedrobaeza:
I have pushed it, so please solve now the conflict

victoralmau:
> Don't do such things, or conflicts will arise between both. The only way to avoid it is to push the translations immediately.

Ok sorry, I will keep this in mind in the future.

PR #926, opened 2022-10-10
[14.0][FIX] account_financial_report: open_items

are-agilebg:
Fixed the amount_residual_currency on partial reconcile.

This PR is derived from https://github.com/OCA/account-financial-reporting/pull/853, the only differences are in the adaptation to the _accounts_partial_reconcile_ model differences between 13.0 and 14.0.


stevech091 (approved):
Functional test OK

AaronHForgeFlow (approved):
Just tested. I think this is still needed. Thanks.

PR #930, opened 2022-10-20
[14.0][FIX] partner_statement: avoid showing lines with 0.00001 open amounts

MiquelRForgeFlow:
:)

### DIFF
--- a/partner_statement/report/outstanding_statement.py
+++ b/partner_statement/report/outstanding_statement.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, models
+from odoo.tools.float_utils import float_is_zero
 
 
 class OutstandingStatement(models.AbstractModel):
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
@gdgellatly @JordiBForgeFlow @AaronHForgeFlow @ChrisOForgeFlow

MiquelRForgeFlow:
@ChrisOForgeFlow

AaronHForgeFlow (approved):
Code LGTM. The name of the method would be _check_line_significant_residual, or something similar

MiquelRForgeFlow:
@AaronHForgeFlow no, because other custom extensions could add other additional checks in that method. I wanted a generic name on purpose. The method only serves to check if the line should be shown or not. It could also be named `_check_show_line` or similar...

gdgellatly:
> @AaronHForgeFlow no, because other custom extensions could add other additional checks in that method. I wanted a generic name on purpose. The method only serves to check if the line should be shown or not. It could also be named `_check_show_line` or similar...

Hi,

I'm OK with this PR. But I think it could be improved. check_line_is_correct either appends or does not append. It is quite limited and restrictive and doesn't really have much of an extensibility use case. I think if we really want to make extensible then we should really do something more like

`line['currency'].extend(self._prepare_currency_line(line, currency_id))`

then 
```
def _prepare_currency_line(self, line, currency_id)
  if <wrong>:
    return []
  return line
```
which allows
```
def _prepare_currency_line(self, line, currency_id):
  res = super()
  if res and <almost right>:
    res['foo'] = 'bar'
  return res
```

MiquelRForgeFlow:
@gdgellatly done

gdgellatly (approved):
LGTM

JordiBForgeFlow (approved):

MiquelRForgeFlow:
Forward-port in https://github.com/OCA/account-financial-reporting/pull/936.

PR #931, opened 2022-10-21
[14.0][IMP] add filtering for journals on general ledger

ghost:


account_financial_report/wizard/general_ledger_wizard_view.xml
```diff
@@ -91,6 +91,14 @@
                                 options="{'no_create': True}"
                             />
                         </page>
+                        <page string="Filter journals">
+                            <field
```
luc-demeyer: Since many companies have lots of journals : please make the field span the page (like the account_ids in the first notebook page.

image

luc-demeyer (changes requested):

account_financial_report/wizard/general_ledger_wizard_view.xml
ghost: The view is changed and with the same layout as the account filtering

luc-demeyer (approved):

PR #933, opened 2022-10-26
[13.0][FIX] account_financial_report: condition where a journal item considered reconciled

AaronHForgeFlow:
cc @ForgeFlow


JordiBForgeFlow:
@AaronHForgeFlow I have changed the criteria and now I can see that the Trial Balance always matches with the Open Items & Aged Partner Balance.

MiquelRForgeFlow (approved):

PR #938, opened 2022-10-27
[14.0][IMP] add current_statement

dnplkndll:


gdgellatly:
Hi,

You can't just open a pull request of nearly 1000 lines on an existing stable module and leave the explanation blank and expect people to just get it.

Why? References? How it differs etc

Honestly from what I could tell (from the 1 line of explanation buried in settings), it is just another name for an outstanding statement.

PR #939, opened 2022-10-28
[16.0][MIG] account_financial_report: Migration to 16.0 

ramiadavid:
I am migrating this module and it requires some changes:

**General ledger:**

- The Analytic account field has been modified by Analytic distribution which is a JSON that allows assign a % in several analytic accounts: 
  - I modify the report so that it shows this information. If you don't like the appearance both in the report and in the xlsx, tell me what you think is better
  - As the new Analytic distribution field is a JSON, I can't find a way to filter the move lines by analytic account in the domain, so I have added a many2many computed field to store the analytic accounts from the JSON to be able to filter. If you know a better way tell me how I can do it
- The account.analytic.tag model has been deleted: 
  - It is removed from the report
 
This depends on https://github.com/OCA/reporting-engine/pull/654 which is pending to be merged


RodrigoBM:
Hi @ramiadavid ,
Thanks for this pr.

Can you fix red pre-commit and add report_xlsx to test-requirements.txt to fix runboat and tests?

Inspiration can be found [there](https://github.com/OCA/reporting-engine/pull/655/commits/503fda43373bcef4f138adb19e74aa4f06c53a96)

Thanks!

ramiadavid:
@RodrigoBM Ok, runboad and tests fixed but pre-commit  gives an error

`An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: could not read Username for 'https://gitlab.com/': No such device or address`

pedrobaeza:
pre-commit fixed in #957 

You can rebase. The temporary PR put for report_xlsx seems to not be the definitive one though.

victoralmau:
Please, cherry-pick https://github.com/OCA/account-financial-reporting/pull/935 and https://github.com/OCA/account-financial-reporting/pull/958 to commit history.

RodrigoBM:
Hi @ramiadavid,

Delete last commit "[[TO RMV] add test dependency](https://github.com/OCA/account-financial-reporting/pull/939/commits/4164fcd70ae7f68b8821dbc9ab9bc5f34d57b963)". The pull has been merge  https://github.com/OCA/reporting-engine/pull/690

Thanks

victoralmau:
Please, cherry-pick https://github.com/OCA/account-financial-reporting/pull/962, https://github.com/OCA/account-financial-reporting/pull/965, https://github.com/OCA/account-financial-reporting/pull/964 and https://github.com/OCA/account-financial-reporting/pull/963 to commit history.

pedrobaeza:
There are more PRs to come, so maybe better to wait and get again whole story...

account_financial_report/report/general_ledger.py
```diff
@@ -365,11 +360,10 @@ def _get_period_domain(
         only_posted_moves,
         date_to,
         date_from,
-        analytic_tag_ids,
         cost_center_ids,
     ):
         domain = [
-            ("display_type", "=", False),
+            ("display_type", "not in", ["line_note", "line_section"]),
```
victoralmau: Sorry, what is the reason for this change? (appears in several places)

ramiadavid: v16 has changes in this field usage, and has more values than old versions

ramiadavid: v16

```
 display_type = fields.Selection(
        selection=[
            ('product', 'Product'),
            ('cogs', 'Cost of Goods Sold'),
            ('tax', 'Tax'),
            ('rounding', "Rounding"),
            ('payment_term', 'Payment Term'),
            ('line_section', 'Section'),
            ('line_note', 'Note'),
            ('epd', 'Early Payment Discount')
        ],
        compute='_compute_display_type', store=True, readonly=False, precompute=True,
        required=True,
    )
```

v15
```
display_type = fields.Selection([
        ('line_section', 'Section'),
        ('line_note', 'Note'),
    ], default=False, help="Technical field for UX purpose.")
```

victoralmau: Ok, sorry, I didn't know it, thank you very much.

account_financial_report/models/account_move_line.py
```diff
@@ -0,0 +1,58 @@
...
+            ON account_move_line (account_id, partner_id)"""
+            )
+
+    @api.model
+    def search_count(self, args):
+        # In Big DataBase every time you change the domain widget this method
+        # takes a lot of time. This improves performance
+        if self.env.context.get("skip_search_count"):
+            return 0
+        return super(AccountMoveLine, self).search_count(args)
```
florian-dacosta:
```suggestion
    def search_count(self, domain, limit=None):
        # In Big DataBase every time you change the domain widget this method
        # takes a lot of time. This improves performance
        if self.env.context.get("skip_search_count"):
            return 0
        return super().search_count(domain, limit=limit)
```

LESTRAT21:
I did a partial functionnal review : customer invoice + vendor invoice + manual payment. The partner ledger is ok

victoralmau:
Please cherry-pick https://github.com/OCA/account-financial-reporting/pull/966 to historial commits

account_financial_report/models/account_move_line.py
ramiadavid: Updated

ramiadavid:
> Please cherry-pick #966 to historial commits

Done

victoralmau (approved):
Code review OK

account_financial_report/models/account_move_line.py
```diff
@@ -1,11 +1,29 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).-
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    analytic_account_ids = fields.Many2many(
+        "account.analytic.account", compute="_compute_analytic_account_ids", store=True
```
pedrobaeza, line 10: You must avoid this new stored field that will be triggered on module installation. Imagine if you already have millions of journal items...

Why is this needed?

account_financial_report/models/account_move_line.py
```diff
@@ -32,9 +50,9 @@ def init(self):
...
 
     @api.model
-    def search_count(self, args):
+    def search_count(self, domain, limit=None):
         # In Big DataBase every time you change the domain widget this method
         # takes a lot of time. This improves performance
         if self.env.context.get("skip_search_count"):
             return 0
-        return super(AccountMoveLine, self).search_count(args)
+        return super().search_count(domain, limit)
```
pedrobaeza: Always respect keyword arguments:

```suggestion
        return super().search_count(domain, limit=limit)
```

account_financial_report/models/ir_actions_report.py
```diff
@@ -13,13 +13,13 @@ def _prepare_account_financial_report_context(self, data):
         return dict(self.env.context or {}, lang=lang) if lang else False
 
     @api.model
-    def _render_qweb_html(self, docids, data=None):
+    def _render_qweb_html(self, report_ref, docids, data=None):
         context = self._prepare_account_financial_report_context(data)
         obj = self.with_context(**context) if context else self
-        return super(IrActionsReport, obj)._render_qweb_html(docids, data)
+        return super(IrActionsReport, obj)._render_qweb_html(report_ref, docids, data)
```
pedrobaeza:
```suggestion
        return super(IrActionsReport, obj)._render_qweb_html(report_ref, docids, data=data)
```

account_financial_report/models/ir_actions_report.py
```diff
@@ -13,13 +13,13 @@ def _prepare_account_financial_report_context(self, data):
...
 
     @api.model
-    def _render_xlsx(self, docids, data):
+    def _render_xlsx(self, report_ref, docids, data=None):
         context = self._prepare_account_financial_report_context(data)
         obj = self.with_context(**context) if context else self
-        return super(IrActionsReport, obj)._render_xlsx(docids, data)
+        return super(IrActionsReport, obj)._render_xlsx(report_ref, docids, data)
```
pedrobaeza:
```suggestion
        return super(IrActionsReport, obj)._render_xlsx(report_ref, docids, data=data)
```

account_financial_report/report/general_ledger.py
```diff
@@ -99,12 +103,12 @@ def _get_initial_balances_pl_ml_domain(
     def _get_accounts_initial_balance(self, initial_domain_bs, initial_domain_pl):
         gl_initial_acc_bs = self.env["account.move.line"].read_group(
             domain=initial_domain_bs,
-            fields=["account_id", "debit", "credit", "balance", "amount_currency"],
+            fields=["account_id", "debit", "credit", "balance", "amount_currency:sum"],
```
pedrobaeza, line 106: Why adding `:sum`? Isn't the default operation?

account_financial_report/report/general_ledger.py
```diff
@@ -335,14 +336,8 @@ def _get_move_line_data(self, move_line):
             "rec_name": move_line["full_reconcile_id"][1]
             if move_line["full_reconcile_id"]
             else "",
-            "tag_ids": move_line["analytic_tag_ids"],
```
pedrobaeza: Why no tags now?

pedrobaeza (changes requested):
The analytic conversion doesn't seem correct to me, as it involves an stored m2m field that is going to provoke a bottleneck both on install and on daily work. Data should be extracted on the fly from the new data model. And tags equivalent is lost.

account_financial_report/report/general_ledger.py
ramiadavid: account.analytic.tag has been removed in v16

account_financial_report/models/account_move_line.py
ramiadavid: Ok

account_financial_report/models/ir_actions_report.py
ramiadavid: Ok

account_financial_report/models/ir_actions_report.py
ramiadavid: Ok

account_financial_report/models/account_move_line.py
ramiadavid: @pedrobaeza 
The analytic_account_id field has been changed to analytic_distribution which is a JSON.
I've created this new field because, in order to filter by Analytical Account, I don't know any other way to do it with a JSON field in a domain. And that created field has to be stored to be able to use it in a domain.

This is used in [reports/general_ledger.py#L278](https://github.com/OCA/account-financial-reporting/blob/117d2e5976f414b5b41768e933f71c545c1d5a7f/account_financial_report/report/general_ledger.py#L278)

You know some way?

Thinking now, i think that we could use the analytic lines `analytic_line_ids` field to filter using the domain `('analytic_line_ids.account_id', 'in, 'xxx')`. But these lines are not created until the invoice is published, and I don't know if this is a problem for a report

account_financial_report/report/general_ledger.py
ramiadavid: In v16 in the `amount_currency` field they have added the parameter `group_operator=None`, I don't know if that's why, but if I don't add the `:sum` it gives an error.

pedrobaeza (commented):
Thanks for the efforts done, David.

2 things I have found reviewing the final result:

- I see good to have a different color for the clickable elements, but the blue is too light and difficult a bit the reading. Can you put a darker blue?
- This is outside of the migration work, but I see awkward (and also risky) that an invoicing user is able to see all the financial reports, so I think we should restrict them to "show full account" group. This can be made in a separate commit to be backported to previous versions.

ramiadavid:
> Thanks for the efforts done, David.
> 
> 2 things I have found reviewing the final result:
> 
> * I see good to have a different color for the clickable elements, but the blue is too light and difficult a bit the reading. Can you put a darker blue?
> * This is outside of the migration work, but I see awkward (and also risky) that an invoicing user is able to see all the financial reports, so I think we should restrict them to "show full account" group. This can be made in a separate commit to be backported to previous versions.

Hello @pedrobaeza  I have not made any changes regarding the colors, it is the default color for the links in the Odoo reports, maybe they have changed it for v16. I assign a darker color by modifying the report.css. tell me how you see it now

Regarding the change of permissions, and in order not to make a mistake, it would only be to delete the first group, correct?

`    <menuitem
        parent="account.menu_finance_reports"
        id="menu_oca_reports"
        name="OCA accounting reports"
        groups="account.group_account_manager,account.group_account_user"
    />
`

pedrobaeza:
Yes, David, the naming of the groups is terrible, as `group_account_user` gives more permissions than `group_account_manager`!

ramiadavid:
@pedrobaeza access group changed, but I don't know what is the procedure to make a backport to other versions, is it necessary to do a PR to each one of them with that commit? I guess when this PR is merged

It remains to be seen what to do with the analytic_account_ids field

pedrobaeza (approved):
Let's keep it for now as is with the analytic thing and others can do further pull requests to make more improvements.

About the backport, you have to cherry-pick the commit in a new branch and make the pull request.

/ocabot merge nobump

rousseldenis:
@pedrobaeza @ramiadavid The introduction of the new field should have at least been set with an init hook as on existing database, it is too bulky ! I will do it.

Moreover, that field IMHO should have been set in another module...

pedrobaeza:
Well, sometimes the side effects can't be anticipated. I didn't realize about it when reviewing.

PR #945, opened 2022-11-14
[13.0][IMP] account_*_stock_report_non_billed: Allow use dates as a range

CarlosRoca13:
By doing this change we can use dates (threshold and check) as a range to discart moves and invoices that are out of this range.

cc @Tecnativa TT40443

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/account_purchase_stock_report_non_billed/models/stock_move.py
+++ b/account_purchase_stock_report_non_billed/models/stock_move.py
@@ -63,9 +63,12 @@ class StockMove(models.Model):
         self.ensure_one()
         if self.purchase_line_id:
             self.quantity_not_invoiced = qty_to_invoice - invoiced_qty
-            self.price_not_invoiced = (
-                qty_to_invoice - invoiced_qty
-            ) * self.purchase_line_id.price_unit
+            price_unit = self.purchase_line_id.price_unit
+            if "discount" in self.purchase_line_id._fields:
+                price_unit = self.purchase_line_id.price_unit * (
+                    1 - self.purchase_line_id.discount / 100
+                )
+            self.price_not_invoiced = (qty_to_invoice - invoiced_qty) * price_unit
         else:
             super()._set_not_invoiced_values(qty_to_invoice, invoiced_qty)
 
--- a/account_purchase_stock_report_non_billed/models/stock_move.py
+++ b/account_purchase_stock_report_non_billed/models/stock_move.py
@@ -31,6 +31,12 @@ class StockMove(models.Model):
             )
             # Check when grouping different moves in an invoice line
             moves = invoice_lines.mapped("move_line_ids")
+            date_start = self.env.context.get("moves_date_start")
+            date_end = self.env.context.get("moves_date_end")
+            if date_start and date_end:
+                moves = moves.filtered(
+                    lambda ml: ml.date_done >= date_start and ml.date_done <= date_end
+                )
             total_qty = moves.get_total_devolution_moves()
             if qty_invoiced != total_qty:
                 invoiced = 0.0
--- a/account_purchase_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
+++ b/account_purchase_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
@@ -14,6 +14,7 @@ class AccountSaleStockReportNonBilledWiz(models.TransientModel):
             [
                 domain,
                 [
+                    ("date_done", ">=", self.stock_move_non_billed_threshold),
                     ("date_done", "<=", self.date_check),
                     ("purchase_line_id", "!=", False),
                     ("state", "=", "done"),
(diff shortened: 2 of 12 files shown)


account_purchase_stock_report_non_billed/models/stock_move.py
```diff
@@ -60,6 +60,15 @@ def _set_not_invoiced_values(self, qty_to_invoice, invoiced_qty):
...
             ) * self.purchase_line_id.price_unit
+            if "discount" in self.purchase_line_id._fields:
+                self.price_not_invoiced = (qty_to_invoice - invoiced_qty) * (
+                    self.purchase_line_id.price_unit
+                    - (
+                        self.purchase_line_id.discount
+                        * self.purchase_line_id.price_unit
+                        / 100
+                    )
+                )
```
pedrobaeza: Formula is not correct and better to not repeat operations:

```suggestion
            price_unit = self.purchase_line_id.price_unit
            if "discount" in self.purchase_line_id._fields:
                price_unit = self.purchase_line_id.price_unit * (1 - self.purchase_line_id.discount / 100)
            self.price_not_invoiced = (
                qty_to_invoice - invoiced_qty
            ) * price_unit
```

account_purchase_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
```diff
@@ -14,7 +14,6 @@ def _get_search_domain(self):
             [
                 domain,
                 [
-                    ("date_done", "<=", self.date_check),
```
pedrobaeza, line 17: Why removing this? Both date intervals should be active.

account_sale_stock_report_non_billed/models/res_company.py
```diff
@@ -10,3 +10,4 @@ class ResCompany(models.Model):
     stock_move_non_billed_threshold = fields.Date(
         string="Non Billed Threshold Date", default=fields.Date.today
     )
+    use_dates_as_deadlines = fields.Boolean()
```
pedrobaeza: This should be put at wizard level, not at company level.

account_sale_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
```diff
@@ -12,14 +12,19 @@ class AccountSaleStockReportNonBilledWiz(models.TransientModel):
...
         default=lambda self: self._default_stock_move_non_billed_threshold()
     )
     date_check = fields.Date(string="Date", default=fields.Date.today)
+    use_dates_as_deadlines = fields.Boolean(
+        default=lambda self: self._default_use_dates_as_deadlines()
+    )
 
     def _get_search_domain(self):
         return [
-            ("date_done", "<=", self.date_check),
```
pedrobaeza, line 22: I think you shouldn't remove this.

account_sale_stock_report_non_billed/models/res_company.py
pedrobaeza: And a better name is: `interval_restrict_invoices`

pedrobaeza (changes requested):

account_purchase_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
CarlosRoca13: Because dates are being filtered at this level: https://github.com/OCA/account-financial-reporting/pull/945/files#diff-6c1eabc073e5d845d186043523e0ffcd942e59098613bee69886297ef8fbd937R77-R83

account_sale_stock_report_non_billed/models/res_company.py
CarlosRoca13: Okay, I will change the name, I just put this at company level to allow set this value as default if the client wants. Obviusly, it can be changed at wizard level

pedrobaeza: OK, not bad idea, but please use the existing res.config.settings mechanism for that: putting `default_` + field name + an attribute indicating the model, you make this automatic using default values. Not sure if it can be company related though.

account_purchase_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
pedrobaeza: And why 2 steps? Can't it be done everything in one shot?

account_sale_stock_report_non_billed/models/res_company.py
CarlosRoca13: I didn't know about this mechanism! Thanks for the info :smile:

CarlosRoca13:
Changes done @pedrobaeza

account_sale_stock_report_non_billed/models/res_config_settings.py
```diff
@@ -10,3 +10,7 @@ class ResConfigSettings(models.TransientModel):
     stock_move_non_billed_threshold = fields.Date(
         related="company_id.stock_move_non_billed_threshold", readonly=False
     )
+    default_interval_restrict_invoices = fields.Boolean(
+        string="Interval Restrict Invoices",
```
pedrobaeza:
```suggestion
        string="Restrict invoices using the date interval",
```

pedrobaeza (approved):

sergio-teruel (approved):

LudLaf (approved):

PR #947, opened 2022-11-14
Fix line "name" None breaking comparison

dannyadair:
Same fix as https://github.com/OCA/account-financial-reporting/pull/863 but for the activity statement


gdgellatly (approved):
LGTM, tested, error resolved.

dannyadair:
What further activity is required

dreispt (approved):

PR #951, opened 2022-11-22
[13.0][IMP] account_financial_report: extensibility: ml fields

AaronHForgeFlow:
I think the account move line fields gotten should be available to be extended or modified. Some examples:
- One customer may need the analytic account info in an open items report
- Some other customer may need information about purchase orders that were appearing in an open items report last month

bonus track: Put a proper class name for the abstract report base model

cc @ForgeFlow

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -103,19 +103,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         domain = self._get_move_lines_domain_not_reconciled(
             company_id, account_ids, partner_ids, only_posted_moves, date_from
         )
-        ml_fields = [
-            "id",
-            "name",
-            "date",
-            "move_id",
-            "journal_id",
-            "account_id",
-            "partner_id",
-            "amount_residual",
-            "date_maturity",
-            "ref",
-            "reconciled",
-        ]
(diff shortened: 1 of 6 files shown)


MiquelRForgeFlow (approved):
So much win.

ChrisOForgeFlow (approved):
LGTM

account_financial_report/report/aged_partner_balance.py
```diff
@@ -103,19 +103,7 @@ def _get_move_lines_data(
...
-            "move_id",
-            "journal_id",
-            "account_id",
-            "partner_id",
-            "amount_residual",
-            "date_maturity",
-            "ref",
-            "reconciled",
-        ]
+        ml_fields = self._get_ml_fields()
```
pedrobaeza, line 106: Instead of putting all fields again, please put the common ones in abstract, and here only the extra ones.

pedrobaeza (changes requested):
Check my comment

account_financial_report/report/aged_partner_balance.py
AaronHForgeFlow: Hi @pedrobaeza I just did that :) Except for the vat report, because it has no common fields, just id field.

pedrobaeza (approved):
/ocabot merge patch

PR #952, opened 2022-11-22
[13.0][FIX] mis_builder_cash_flow: Change to the value of the corresponding fields to avoid incorrect values in some use cases.

victoralmau:
Change to the value of the corresponding fields to avoid incorrect values in some use cases (https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L3344-L3348).

@Tecnativa TT40533


pedrobaeza:
As commented internally, this is not correct. The problem is the type / reconcile flag of the account, so closing.

PR #954, opened 2022-11-25
[13.0][FIX] account_financial_report: Prevent error in general_ledger if foreign_currency is not set.

victoralmau:
Prevent error in general_ledger if foreign_currency is not set.

@Tecnativa TT40647


carlosdauden (approved):

pedrobaeza (approved):

pedrobaeza:
This shouldn't be needed in 14+, as foreign currency is always set.

/ocabot merge patch

PR #949, opened 2022-11-21
[14.0][FIX] account_financial_report: Set the correct value of column 'Cumul cur.' in general ledger

victoralmau:
FWP from 13.0: https://github.com/OCA/account-financial-reporting/pull/942

Changes done:
- [x]  Improve domain to reduce code in general ledger.
- [x] Set the correct value of column 'Cumul cur.' in general ledger.
- [x] Change `partner_domain` to `misc_grouped_domain` in general_ledger.

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT39754

PR #966, opened 2022-12-16
[15.0][FIX] account_financial_report: Set the correct value of column 'Cumul cur.' in general ledger

victoralmau:
FWP from 14.0: https://github.com/OCA/account-financial-reporting/pull/949

Changes done:
- [x]  Improve domain to reduce code in general ledger.
- [x] Set the correct value of column 'Cumul cur.' in general ledger.
- [x] Change `partner_domain` to `misc_grouped_domain` in general_ledger.

@Tecnativa TT39754


pedrobaeza (approved):
And another one fast-tracked to fix the issue in all the branches:

/ocabot merge minor

PR #967, opened 2022-12-16
[14.0][IMP] partner_statement: add Detailed Activity report

MiquelRForgeFlow:
This PR includes:

- Minor fixes.
- New report: Detailed Activity Statement, which is as follows:

The report starts with a "Prior Balance" section summarizing the open items prior to the Date Range of the Statement, with a resulting total.
Then a section "Activity during the period" follows, that intends to explain the transactions that have happened during the period, both debits and credits. It starts with initial balance being the result of the Prior Balance and then shows all the transactions, indicating where they have been applied. The report ends with an ending balance.
A final section "Ending Balance" intends to provide details about the open items at the end of the period, and the total open amount should match with the ending balance in the preceding section.

With this kind of report I think that any user will have a comprehensive understanding of how their payments have been applied, and what is left open.

NOTE: If any of you have a customization of this module, it will not be difficult to adapt to this changes, don't worry.

image


JordiBForgeFlow:
@gdgellatly I think that you will like this report.

ChrisOForgeFlow (approved):
LGTM

JordiBForgeFlow (approved):

gdgellatly:
@JordiBForgeFlow I do, I will run by my finance team in staging. I just have wee concern with one of the queries, it look like it might be reintroducing a bug I introduced and subsequently resolved. Will try and find and verify.

EDIT: No it isn't reintroduced

dnplkndll:
@MiquelRForgeFlow 
the  balance up to is the end versus the start date to summarize the opening value. 

image

PR #969, opened 2022-12-21
[14.0][FIX] account_financial_report: generate ledger document

Reyes4711-S73:

In a runbot I test this:
**view general ledger**
image
image
**export xlsx**
image
image


pedrobaeza (approved):
Fast-tracking this to fix the problem:

/ocabot merge patch

Is this applicable to other versions?

Reyes4711-S73:
@pedrobaeza I'm going to review if the problem is in other versions and generate the corresponding PR

PR #970, opened 2022-12-23
[14.0][IMP][FWD] account_financial_report: extensibility: ml fields 

AaronHForgeFlow:
Forward port of https://github.com/OCA/account-financial-reporting/pull/951

cc @ForgeFlow

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -297,6 +297,8 @@ class AbstractReportXslx(models.AbstractModel):
                     value or "",
                     report_data["formats"]["format_right"],
                 )
+            else:
+                self.write_non_standard_column(cell_type, col_pos, value)
         report_data["row_pos"] += 1
 
     def write_initial_balance(self, my_object, label, report_data):
@@ -698,5 +700,11 @@ class AbstractReportXslx(models.AbstractModel):
     def _get_col_pos_final_balance_label(self):
         """
         :return: the columns position used for final balance label.
+        """
+        raise NotImplementedError()
+
+    def write_non_standard_column(self, cell_type, col_pos, value):
+        """
+        Write columns out of the columns type defined here.
         """
         raise NotImplementedError()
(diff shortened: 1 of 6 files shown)


DavidJForgeFlow (approved):
LGTM

mariadforgeflow (approved):
LGTM!

JoanSForgeFlow (approved):

account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -297,7 +297,9 @@ def write_line_from_dict(self, line_dict, report_data):
                     value or "",
                     report_data["formats"]["format_right"],
                 )
-        report_data["row_pos"] += 1
+            else:
+                self.write_non_standard_column(cell_type, col_pos, value)
+        self.row_pos += 1
```
pedrobaeza: Why changing this? It seems incorrect

AaronHForgeFlow: In my use case. I am adding the purchase order information to the report. When I query the purchase_id field in the method "_get_report_columns" I do it this way:

```

            6: {
                "header": _("Purchase Order"),
                "field": "purchase_id",
                "width": 11,
                "type": "purchase",
            },
```

With that hook I can then write the value I need. Putting type "many2one" in the columns declaration was not working in my case (at least in v13). Then I decided to put the hook so I can take the value when printing the line. I will check in this version if it works without the hook.

pedrobaeza: I was just talking about changing `report_data["row_pos"] += 1` with `self.row_pos += 1`

AaronHForgeFlow: I overlooked this. Yes, that is non intentional and seems an incorrect merge conflict resolution.

AaronHForgeFlow: Fixed and it is working now. Sorry all.

pedrobaeza (approved):
/ocabot merge patch

PR #975, opened 2022-12-24
[12.0][IMP] account_financial_report: change report access groups to only 'Show Full Accounting Features'

ramiadavid:
Backport of #939 

@pedrobaeza

### DIFF
--- a/account_financial_report/menuitems.xml
+++ b/account_financial_report/menuitems.xml
@@ -5,7 +5,7 @@
         parent="account.menu_finance_reports"
         id="menu_oca_reports"
         name="OCA accounting reports"
-        groups="account.group_account_manager,account.group_account_user"
+        groups="account.group_account_user"
         />
 
     <menuitem


pedrobaeza (approved):

ramiadavid:
Hi @pedrobaeza What is it codecov/project and why it fails in many PR's is it not important?

pedrobaeza:
It's a measure of the test coverage, but I'm not sure is working correctly on all of the cases. It's ignored most of the times as it's not critical for the module to work, but just a quality measure.

In this PR the problem may be the Travis failing, that don't pass to Codecov all the data.

Anyway, the backport to this older versions is not convenient unless you have explicit interest on it. To cover the currently supported versions is enough as a good contributor :)

ramiadavid:
@pedrobaeza Ok, so I cancel the PR for the previous versions? v11 v10 v9

pedrobaeza:
What you prefer. As you have done them already, maybe someone wants it, so let's see the community reaction.

PR #981, opened 2022-12-26
[14.0][FIX] account_financial_report: foregin_currency check in general_ledger_xlsx

GonzaloVich:
Change done:
- [x] Added a missing if in line 237. It was causing the Excel report to fail if "Show foreign currency" was not selected on the Wizard.

Please, @pedrobaeza could you review it? Thank you for your time.


pedrobaeza:
Maybe the patch should go in this direction instead: #969

PR #982, opened 2022-12-28
[16.0][FIX] account_financial_report: generate ledger document

Reyes4711-S73:


Reyes4711-S73:
@pedrobaeza This is the PR fixing the bug appeared in versions 14.0 and 15.0 that you merged before.

pedrobaeza (approved):
Thanks for finishing the chain:

/ocabot merge patch

PR #985, opened 2023-01-04
[14.0][FIX] account_financial_report: General ledger changes

victoralmau:
General ledger changes:
- Fix total_bal_curr value from General ledger to prevent wrong values
- Prevent wrong values

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT41041

### DIFF
--- a/account_financial_report/report/general_ledger_xlsx.py
+++ b/account_financial_report/report/general_ledger_xlsx.py
@@ -153,7 +153,7 @@ class GeneralLedgerXslx(models.AbstractModel):
         # For each account
         for account in general_ledger:
             # Write account title
-            total_bal_curr = account["init_bal"].get("bal_curr", 0)
+            total_bal_curr = 0
             self.write_array_title(
                 account["code"] + " - " + accounts_data[account["id"]]["name"],
                 report_data,
@@ -234,7 +234,7 @@ class GeneralLedgerXslx(models.AbstractModel):
 
             else:
                 # For each partner
-                total_bal_curr = account["init_bal"]["bal_curr"]
+                total_bal_curr = 0
                 for group_item in account["list_grouped"]:
                     # Write partner title
                     self.write_array_title(group_item["name"], report_data)
--- a/account_financial_report/report/templates/general_ledger.xml
+++ b/account_financial_report/report/templates/general_ledger.xml
@@ -369,10 +369,7 @@
                 </t>
             </div>
             <!-- Display each lines -->
-            <t
-                t-set="total_bal_curr"
-                t-value="account_or_group_item_object['init_bal'].get('bal_curr', 0)"
-            />
+            <t t-set="total_bal_curr" t-value="0" />
             <t t-foreach="account_or_group_item_object['move_lines']" t-as="line">
                 <!-- # lines or centralized lines -->
                 <div class="act_as_row lines">


account_financial_report/report/trial_balance.py
```diff
@@ -226,21 +226,37 @@ def _compute_account_amount(
...
         self, total_amount, tb, acc_id, prt_id, foreign_currency
     ):
-        total_amount[acc_id][prt_id] = {}
-        total_amount[acc_id][prt_id]["credit"] = 0.0
-        total_amount[acc_id][prt_id]["debit"] = 0.0
-        total_amount[acc_id][prt_id]["balance"] = 0.0
-        total_amount[acc_id][prt_id]["initial_balance"] = tb["balance"]
-        total_amount[acc_id][prt_id]["ending_balance"] = tb["balance"]
+        # Add keys to dict if not exists
+        if acc_id not in total_amount.keys():
```
chienandalu: The `in` expression over a `dict` already check if the object is a key.
```suggestion
        if acc_id not in total_amount:
```

account_financial_report/report/trial_balance.py
```diff
@@ -226,21 +226,37 @@ def _compute_account_amount(
...
+            total_amount[acc_id] = self._prepare_total_amount(foreign_currency)
+        if prt_id not in total_amount[acc_id].keys():
```
chienandalu: Same here
```suggestion
        if prt_id not in total_amount[acc_id]:
```

chienandalu (approved):
Just two little comments. Aside from that LGTM :+1:

account_financial_report/report/trial_balance.py
```diff
@@ -226,21 +226,37 @@ def _compute_account_amount(
...
+    @api.model
+    def _prepare_total_amount(self, foreign_currency):
+        res = {
+            "credit": 0.0,
+            "debit": 0.0,
+            "balance": 0.0,
+            "initial_balance": 0.0,
+            "ending_balance": 0.0,
+        }
+        if foreign_currency:
```
pedrobaeza: Shouldn't this be initialized always? On v14, `amount_currency` and `currency_id` is always filled in the `account.move.line`.

victoralmau: Ok, changes done.

pedrobaeza: I was just asking. What I don't want also is to have filled such columns when no different currency is set. You tell me...

victoralmau: Before this PR and now, the value of the corresponding columns are filled in (I didn't want to make any "extra" changes to avoid side effects).

Taking into account `amount_currency` and `currency_id` fields is probably interesting, but I prefer not to change the current behaviour.

pedrobaeza: But I'm not saying that. I'm asking if current behavior is correct after the previous changes you did are correct. Is foreign currency being shown in the PDF/XLSX only when it's different from the main currency?

victoralmau: Ok sorry, now I understand what you are saying.

Before if the initial and final balance was shown (in the corresponding currency) but NOT the currency column, I have added a change to show this data.

pedrobaeza:
Testing on runbot, I'm afraid this is not working properly.

On Excel export:

Selección_016

On direct view:

Selección_017

victoralmau:
> Testing on runbot, I'm afraid this is not working properly.
> 
> On Excel export:
> 
> Selección_016
> 
> On direct view:
> 
> Selección_017

This is General Ledger isn't it (I was checking trial balance)?
I'll check it anyway now.

pedrobaeza:
On general ledger, I still find an error on accumulated amounts:

Selección_020

and on trial balance, no foreign currency is shown at all:

Selección_021

victoralmau:
Problems related to General ledger is because `bal_curr` = `amount_currency` is set (`account.move.line `field) but in v14 that value is always set even if currency is not set, and that causes the "inconsistency" you are seeing.

pedrobaeza:
Yes, but you need to fix them. It's not correct as it is.

victoralmau:
> On direct view:
> 
> Selección_017

Previously, the value was:
ending balance = initial balance - balance
175.74 = 541.10 - 378.89

I don't understand why that was the case, but I BELIEVE it is incorrect and should be similar to General ledger.

Changes have been applied to show the correct values (with currency) and to remove the currency column.

victoralmau:
@pedrobaeza can you review it again?

Now, only minimal changes related to general ledger are added. 
Trial balance changes are at https://github.com/OCA/account-financial-reporting/pull/992 (it is a FWP of v13).

pedrobaeza (approved):
/ocabot merge patch

PR #986, opened 2023-01-10
[13.0][IMP+FIX] account_financial_report: Trial balance changes

victoralmau:
Trial balance changes:
- [FIX] Prevent wrong values

Related to https://github.com/OCA/account-financial-reporting/pull/985

@Tecnativa TT41158

### DIFF
--- a/account_financial_report/report/templates/trial_balance.xml
+++ b/account_financial_report/report/templates/trial_balance.xml
@@ -222,12 +222,10 @@
                 <!--## Ending balance-->
                 <div class="act_as_cell" style="width: 9%;">Ending balance</div>
                 <t t-if="foreign_currency">
-                    <!--## currency_name-->
-                    <div class="act_as_cell" style="width: 4%;">Cur.</div>
                     <!--## amount_currency-->
-                    <div class="act_as_cell" style="width: 9%;">Initial
+                    <div class="act_as_cell" style="width: 11%;">Initial
                         balance cur.</div>
-                    <div class="act_as_cell" style="width: 9%;">Ending balance
+                    <div class="act_as_cell" style="width: 11%;">Ending balance
                         cur.</div>
                 </t>
             </div>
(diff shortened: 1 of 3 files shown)


account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -245,6 +245,25 @@ def write_line_from_dict(self, line_dict):
...
+            # the value empty.
+            # - If the currency is different from the company's currency, we will
+            # show the value.
+            if cell_type == "amount_different_company_currency":
+                if line_dict.get("currency_id") and line_dict.get(
+                    "company_currency_id"
+                ):
+                    if line_dict["currency_id"] == line_dict["company_currency_id"]:
+                        value = ""
+                        cell_type = "string"
```
pedrobaeza: This value shouldn't be modified across the process, as this makes very confusing which value it has each time you access it. Please find another way without modifying this. The same with `value`.

victoralmau: The `amount_different_company_currency` option was added in v14 with https://github.com/OCA/account-financial-reporting/commit/422b9be13ed68315eed561137a67e717635bb573 and is not in v13 yet (I have added it in this commit but it could go in another separate commit or even another PR).

In this particular case the goal is just NOT to show the value being sent when it doesn't match (when there is no currency set, or when it is the same as the company's).

victoralmau: Finally I have removed `amount_different_company_currency` option because it is not needed in v13.
In v14 I will also remove it using another technique (set `amount_currency` value to 0, just as it is in v13 https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L1231 to simplify everything).

pedrobaeza (approved):

carlosdauden (approved):
Code LGTM

PR #988, opened 2023-01-12
[14.0][FIX] partner_statement: several fixes

MiquelRForgeFlow:
Fixes:

 - Initial balance for Activity Statement: Consider moves reconciled with moves that have different partner (use open amounts)
 - Reconciled moves for Detailed Activity Statement: Consider moves reconciled with moves that have different partner (they should appear).
 - Ending amounts should be correct in all cases.
 - Other minor fixes.

Improvements:

 - Refactor of Activity Statement: include columns of "Applied Amount" and "Open Amount", and removed "Balance" column.
 - Removed option "Show Balance column" for Detailed Activity Statement.

### DIFF
--- a/partner_statement/report/activity_statement.py
+++ b/partner_statement/report/activity_statement.py
@@ -5,6 +5,8 @@ from collections import defaultdict
 
 from odoo import api, models
 
+from .outstanding_statement import OutstandingStatement
+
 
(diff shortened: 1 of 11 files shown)


JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
Functional test in runbot

PR #971, opened 2022-12-23
[15.0][IMP][FWD] account_financial_report: extensibility: ml fields

AaronHForgeFlow:
Forward port of https://github.com/OCA/account-financial-reporting/pull/951

cc https://github.com/ForgeFlow

PR #990, opened 2023-01-16
[16.0][IMP] account_financial_report: extensibility xlsx: allow to use other column types (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/account-financial-reporting/pull/971.


AaronHForgeFlow (approved):
Tested. LGTM

PR #991, opened 2023-01-16
[13.0][FIX] partner_statement: minor fixes (partial backport)

MiquelRForgeFlow:
Backport of some minor fixes done in https://github.com/OCA/account-financial-reporting/pull/967 and https://github.com/OCA/account-financial-reporting/pull/988.

### DIFF
--- a/partner_statement/__manifest__.py
+++ b/partner_statement/__manifest__.py
@@ -7,6 +7,7 @@
     "category": "Accounting & Finance",
     "summary": "OCA Financial Reports",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "maintainers": ["MiquelRForgeFlow"],
     "website": "https://github.com/OCA/account-financial-reporting",
     "license": "AGPL-3",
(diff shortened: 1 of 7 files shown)


gdgellatly (changes requested):
The PR that this is backporting seems to now produce wrong opening balances. Suggest we hold until those are resolved.

MiquelRForgeFlow:
Thanks, fixed.

MiquelRForgeFlow:
@gdgellatly I would like to merge these backport fixes. Is it fine?

gdgellatly:
@MiquelRForgeFlow I don't know, we are still stuck on a December 2022 release of partner statement because a whole load of changes landed between then and Feb and they produce wrong numbers and in some cases complete nonsense. I haven't checked since about July whether these things have been fixed, but I know the first round of fixes didn't resolve.

MiquelRForgeFlow:
Replaced by https://github.com/OCA/account-financial-reporting/pull/1179.

PR #992, opened 2023-01-17
[14.0][IMP+FIX] account_financial_report: Trial balance changes

victoralmau:
FWP from 13.0: https://github.com/OCA/account-financial-reporting/pull/986

Trial balance changes:

- [FIX] Prevent wrong values

@Tecnativa TT41158

### DIFF
--- a/account_financial_report/report/templates/trial_balance.xml
+++ b/account_financial_report/report/templates/trial_balance.xml
@@ -191,17 +191,11 @@
                 <!--## Ending balance-->
                 <div class="act_as_cell" style="width: 9%;">Ending balance</div>
                 <t t-if="foreign_currency">
-                    <!--## currency_name-->
-                    <div class="act_as_cell" style="width: 4%;">Cur.</div>
                     <!--## amount_currency-->
-                    <div class="act_as_cell" style="width: 9%;">
-                        Initial
-                        balance cur.
-                    </div>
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I have created an invoice in other currency:

http://oca-account-financial-reporting-14-0-pr992-2970bd8c95f6.runboat.odoo-community.org/web?debug=1#id=15&action=202&model=account.move&view_type=form&cids=1&menu_id=

and it's not being shown in the trial balance:

Selección_015

victoralmau:
> I have created an invoice in other currency:
> 
> http://oca-account-financial-reporting-14-0-pr992-2970bd8c95f6.runboat.odoo-community.org/web?debug=1#id=15&action=202&model=account.move&view_type=form&cids=1&menu_id=
> 
> and it's not being shown in the trial balance:
> 
> Selección_015

I understand the use case you indicate,  but according to the current behaviour everything seems to be correct.

In this report, these columns will only be shown if account has currency set: https://github.com/OCA/account-financial-reporting/blob/14.0/account_financial_report/report/templates/trial_balance.xml#L893-L908

pedrobaeza (approved):
OK, it seems a bit awkward, but being the trial balance and not the general ledger, it can be consistent

carlosdauden (approved):

PR #998, opened 2023-01-25
[16.0][IMP+FIX] account_financial_report: Trial balance changes

victoralmau:
FWP from 15.0: https://github.com/OCA/account-financial-reporting/pull/997

Trial balance changes:

- [FIX] Prevent wrong values

@Tecnativa TT41158


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #999, opened 2023-01-26
[15.0][MIG] account_sale_stock_report_non_billed: Migration to version 15.0

stefan-tecnativa:
cc @Tecnativa TT40223

@chienandalu @CarlosRoca13 please review!


account_sale_stock_report_non_billed/tests/test_account_sale_stock_report_non_billed.py
```diff
@@ -21,7 +22,7 @@ def get_picking_done_so(self):
             and x.state in ("confirmed", "assigned", "partially_available")
         )
         picking.move_line_ids.write({"qty_done": 2})
-        picking.action_done()
+        picking._action_done()
```
chienandalu: change it to `button_validate` in all the cases

stefan-tecnativa: Change done

stefan-tecnativa: Changing all the cases for `button_validate` causes a test to fail. If in that case if I put `_action_done` the test no longer fails. Can it be a problem of my local installation, that there is something that causes conflicts and that is why it fails in something so specific?

chienandalu: Well, that's interesting as that's the action that the users will be triggering.  You have to find out what's the difference that makes the tests fail. Do the same test case work if you try it out by UI?

stefan-tecnativa: Now it should be correct.

chienandalu (approved):
Great!

CarlosRoca13 (approved):

PR #1000, opened 2023-01-30
[14.0][IMP] mis_builder_cash_flow: Improve treatment of forecast and layout (WIP)

JordiBForgeFlow:
The problem with the existing MIS Cash Flow report is that the forecasted amount is always considered. For example, when you forecast to receive 100 and then have an invoice for 100, you don't want to see incoming 200, but 100 instead.

Also, when you have forecasted 100 with date 1/1/2023, and now you are in 1/2/2023, the 100 that you forecasted should be gone, and they are now considered in the "Current" column. By 1/2/2023 you are supposed to have the 100 in your bank.

We avoid counting two times for the same amount. One when the amount is forecasted, and then when you either have a receivable due in this date, or the money has already been received in the bank.

With this PR, the layout of the MIS report now separates IN and OUT regardless if it is receivable and payable.

The forecasted amount presented is the original forecasted amount **substracted** with the actual amount from receivables and payables.

Now the report ignores the forecasts that are prior to today's date, as including them would cause an incorrect initial balance, which needs to be your current financial situation without any forecast.

In the example below we created a forecasted receivable in +1W of 600.000 and a forecasted payable in +1W of -200.000.

image

As you can see the IN TOTAL shows 600.000, even when the "In receivable" is 313,375. This is because we forecasted 600.000 for +1W, and we have received already 313.375 of it. We still estimate that we will end up receiving 600.000. That is why the row "In forecast (remaining)" shows 286.625 (that is, the original forecast of 600.000 minus the amount in receivables).

image

### DIFF
--- a/mis_builder_cash_flow/data/mis_report.xml
+++ b/mis_builder_cash_flow/data/mis_report.xml
@@ -6,25 +6,27 @@
         <field name="name">Cash Flow</field>
         <field name="style_id" ref="mis_style_cash_flow" />
     </record>
-    <record id="mis_kpi_aliquidity" model="mis.report.kpi">
-        <field name="report_id" ref="mis_report_cash_flow" />
-        <field name="name">liquidity</field>
-        <field name="description">LIQUIDITY</field>
-        <field name="style_id" ref="mis_style_account_sub_total" />
-        <field name="auto_expand_accounts" eval="True" />
-        <field name="auto_expand_accounts_style_id" ref="mis_style_account_detail" />
-        <field name="sequence">20</field>
-        <field
-            name="expression"
-        >bal[][('account_internal_type', '=', 'liquidity'), ('line_type', '=', 'move_line'), ('account_id.hide_in_cash_flow', '=', False)]</field>
-    </record>
     <record id="mis_kpi_in_total" model="mis.report.kpi">
         <field name="report_id" ref="mis_report_cash_flow" />
         <field name="name">in_total</field>
         <field name="description">IN TOTAL</field>
         <field name="style_id" ref="mis_style_account_sub_total" />
         <field name="sequence">30</field>
-        <field name="expression">in_receivable + in_forecast</field>
+        <field
+            name="expression"
+        >in_liquidity + in_receivable + in_payable + in_forecast</field>
+    </record>
+    <record id="mis_kpi_in_liquidity" model="mis.report.kpi">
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please split the changes in several commits according the fix / imp. Even in different PRs for discussing them separately and faster merge.

JordiBForgeFlow:
@pedrobaeza All changes are related with the same concept. They cannot be split, because it is intended to change a behaviour that is currently wrong.

pedrobaeza:
To separate the layout in IN and OUT is not a fix and it's not related. In fact, this PR should be [FIX], not [IMP].

JordiBForgeFlow:
@pedrobaeza Separating IN and OUT is related in fact. Because it is a necessary prerequisite to computing correctly the "In forecasted (remaining)" amount. (and out)

pedrobaeza:
OK, but you can still separate that "pre-requisite" in a `[IMP]` commit, and then put the `[FIX]` commit with each code correctly separated to be analyzed.

JordiBForgeFlow:
cc @enricostano you may be interested in this

JordiBForgeFlow:
I have now split into separate commits, and also added the liquidity part inside the IN TOTAL and OUT TOTAL

image

However I can see a fundamental flay with the "BALANCE" row. If you have forecasts, the amount will never be correct as it will double count the amount. So, now the amount from the forecast is not double counted in the "IN TOTAL" AND "OUT TOTAL", but the BALANCE is incorrect.

So, we need a way to make sure that the 'BALANCE' column starts with the total bank amount, and in each period it shows the total of the prior period + the period balance. But this is not supported in MIS Builder, AFAIK

pedrobaeza:
I don't get why it counts twice in the balance row. Can you please explain?

JordiBForgeFlow:
@pedrobaeza It counts now twice, and it counted twice before this change. The fact is that the forecasted lines, when added, they will always count twice in the existing report. In this PR I managed to remove them in the IN and OUT sections, but not in the BALANCE.

I'm afraid that MIS Builder is not the right tool to build a cash flow report, for this reason.

enricostano:
Hi @JordiBForgeFlow !

> I'm afraid that MIS Builder is not the right tool to build a cash flow report, for this reason.

Yeah, we're still struggling with this... how Odoo Enterprise solves this?

JordiBForgeFlow:
Closing, as I'm planning to implement a liquidity forecast report.

PR #1003, opened 2023-02-06
[16.0][FIX] account_financial_report: endless installation

yajo:
Installing the module in a DB with more than 50000 accounts and groups made the install stall.

It turns out this method's implementation produced almost endless recursion. it was also depending on some fields that never were used.

Now it depends on the parent path, so when one group is moved to another parent, its computed accounts are recomputed, and parents' too, recursively.

Now, the method is much more performant, and the module gets installed in the same DB in 30s.

@moduon MT-1900

### DIFF
--- a/account_financial_report/models/account_group.py
+++ b/account_financial_report/models/account_group.py
@@ -16,6 +16,7 @@ class AccountGroup(models.Model):
     )
     compute_account_ids = fields.Many2many(
         "account.account",
+        recursive=True,
         compute="_compute_group_accounts",
         string="Compute accounts",
         store=True,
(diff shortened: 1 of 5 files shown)


pedrobaeza:
I think this can benefit from the new attribute `recursive=True`. I also wonder if we need this to be stored.

rafaelbn (approved):
Tested in production 👍 

- Before this patch impossible to install 😢 
- After this patch installation is complete! 😄 

Thank you! ❤️

Shide (approved):

yajo:
> I think this can benefit from the new attribute `recursive=True`. 

AFAICS this is not the case because the computation is bubbling up and not down.

> I also wonder if we need this to be stored.

I wondered the same, but I just wanted to keep the patch minimal as I'm not very familiar with the inner plumbing of the module.

account_financial_report/models/account_group.py
```diff
@@ -57,17 +57,9 @@ def _compute_level(self):
...
+    @api.depends("parent_path", "account_ids")
     def _compute_group_accounts(self):
-        account_obj = self.env["account.account"]
-        accounts = account_obj.search([])
-        for group in self:
-            prefix = group.code_prefix_start if group.code_prefix_start else group.name
-            gr_acc = accounts.filtered(lambda a: a.code.startswith(prefix)).ids
-            group.compute_account_ids = [(6, 0, gr_acc)]
+        for one in self:
+            one.compute_account_ids = one.account_ids | one.group_child_ids.account_ids
```
pedrobaeza: If I understand correctly the purpose of the field, is to pre-compute all the accounts that belong this group or one of its descendants. As such, why not simply search for all the groups which `parent_path` contains the `parent_path` of this group, and join the `account_ids` of them?

If not, I see your code is not complete for 3 levels or more.

yajo: Hmm true... But I wanted to avoid calling to `search()` to be able to leverage prefetching. Let me check if I can use the recursive solution you mentioned earlier.

yajo: Thanks, fixed now. I tested it and it still gives almost the same performance as the original fix.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
This can be backported to v15

Issue #1005, opened 2023-02-09
14.0 Balance Forward now incorrect since latest merges in partner_statement


gdgellatly:
@MiquelRForgeFlow I see you have been involved in these PR's, maybe you know better what has caused this rather major bug.
## Partner Statement
The name of the module that has a bug.

## Describe the bug
Recently Activity Statements received a number of PR's adjusting SQL. Somewhere this has broken and the balance forward now aggregates the balance forward for all companies, not just the company sending the statement.

## To Reproduce
**Affected versions**: 14.0

1. Have a partner shared between 2 companies with old invoices on each.
2. Print Activity Statement
3. Balance Forward is now the total, rather than being specific to company sending

**Expected behavior**
The old behaviour - Balance Forward is company specific

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

gdgellatly:
https://github.com/OCA/account-financial-reporting/blob/9d213e465d8a739c76ea769425d12d13583c10f7/partner_statement/report/activity_statement.py#L88

Also why this JOIN? we have company_id already, why join just to access the same id?

gdgellatly:
https://github.com/OCA/account-financial-reporting/blob/629dd6c74e94aee43a45e2b540daf27ea8ad2a87/partner_statement/report/activity_statement.py#L106

Shouldn't this be Q3?

Also why the function re-assigning from Outstanding statement into Activity, couldn't we just have moved that into common?

gdgellatly:
ref #988

MiquelRForgeFlow:
> why this JOIN?

Yeah, I don't know. Maybe it was done this was due to a private customization, but not sure. It's true it could be simplified by doing:

```sql
FROM {sub}
WHERE {sub}.id = %(company_id)s
```

> Shouldn't this be Q3?

Yeah, that should fix this issue. Done in https://github.com/OCA/account-financial-reporting/pull/1006. Thanks

> couldn't we just have moved that into common?

No, I did that consciously. I didn't want too much methods in the common to be later replaced in the specific statement, or too much duplicated sqls. That's why I renamed them. Now ones are def _display_activity_xxx and the others are _display_outstanding_xxx, and both are in their corresponding file.

gdgellatly:
[...]

PR #1007, opened 2023-02-10
[14.0][FIX] partner_statement: activity aggregations

MiquelRForgeFlow:
Revert the change in the group by.

Solving aggregation issue in https://github.com/OCA/account-financial-reporting/issues/1005.


AaronHForgeFlow (approved):
Fasttracking

PR #1009, opened 2023-02-14
[14.0][FIX] mis_builder_cash_flow: Call the correct method (action_post) to avoid the warning message in tests.

victoralmau:
Call the correct method (`action_post`) to avoid the warning message in tests.

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #1016, opened 2023-03-09
[14.0][FIX] account_financial_report: Invalid field 'amount_currency'

igallart:
https://github.com/OCA/account-financial-reporting/issues/1015

PR #1025, opened 2023-03-21
[14.0] [FIX] account_financial_report: Fix wrong field `amount_currency` in `account.partial.reconcile`

RachidAlassir:
Replace the wrong field `amount_currency` with two fields `debit_amount_currency` and `credit_amount_currency`.

How to reproduce the bug: 
In a database with data **view** the **Aged Partner Balance** (with a date in the past) and you will get this error:

```
 File ".../account/account_financial_report/report/aged_partner_balance.py", line 74, in _get_account_partial_reconciled
    accounts_partial_reconcile = self.env["account.partial.reconcile"].search_read(
  File "/home/rachid/dev/odoo/odoo/models.py", line 4876, in search_read
    result = records.read(fields)
  File "/home/rachid/dev/odoo/odoo/models.py", line 3011, in read
    raise ValueError("Invalid field %r on model %r" % (name, self._name))
ValueError: Invalid field 'amount_currency' on model 'account.partial.reconcile'
```

In odoo 13 there was this field but not in odoo 14

This bug happened after this MR [926](https://github.com/OCA/account-financial-reporting/pull/926)

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -70,7 +70,13 @@ class AgedPartnerBalanceReport(models.AbstractModel):
 
     def _get_account_partial_reconciled(self, company_id, date_at_object):
         domain = [("max_date", ">", date_at_object), ("company_id", "=", company_id)]
-        fields = ["debit_move_id", "credit_move_id", "amount", "amount_currency"]
+        fields = [
+            "debit_move_id",
+            "credit_move_id",
+            "amount",
+            "debit_amount_currency",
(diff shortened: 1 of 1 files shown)


RachidAlassir:
@pedrobaeza I'm sorry I've just noticed that there's a similar PR [1016](https://github.com/OCA/account-financial-reporting/pull/1016)

pedrobaeza (approved):
This one seems the correct code:

/ocabot merge patch

MiquelRForgeFlow (approved):

PR #1028, opened 2023-03-22
[16.0][mig] partner_statement: Migration to 16.0

dsolanki-initos:


jado95:
Hi @dsolanki-initos , in upgrade from v14 to v16, there is error on table detailed.activity.statement.wizard because in this PR not exist, is correct?

dsolanki-initos:
> Hi @dsolanki-initos , in upgrade from v14 to v16, there is error on table detailed.activity.statement.wizard because in this PR not exist, is correct?

Yes, detailed.activity.statement.wizard was available in v14 but deprecated from v15.

jado95:
> > Hi @dsolanki-initos , in upgrade from v14 to v16, there is error on table detailed.activity.statement.wizard because in this PR not exist, is correct?
> 
> Yes, detailed.activity.statement.wizard was available in v14 but deprecated from v15.

Ok Thanks, but at this point, don't we need an upgrade script or something to avoid the error?

gdgellatly:
> > > Hi @dsolanki-initos , in upgrade from v14 to v16, there is error on table detailed.activity.statement.wizard because in this PR not exist, is correct?
> > 
> > 
> > Yes, detailed.activity.statement.wizard was available in v14 but deprecated from v15.
> 
> Ok Thanks, but at this point, don't we need an upgrade script or something to avoid the error?

Maybe we do, maybe we don't. But if we do, it is not for a v16 PR, it is for V15. However all the info stated here is incorrect. detailed.activity.statement has not been deprecated. It is a recently added feature which is yet to be forward ported. Frankly it should be forward ported to this PR. I would also seriously consider basing this migration off v14 which is more mature than v15, or at the very least pulling in the unapproved PR's.

MiquelRForgeFlow:
@dsolanki-initos We have merged several commits in v15.0, could you rework this commit history to add them?

AaronHForgeFlow:
Superseded here: https://github.com/OCA/account-financial-reporting/pull/1071

PR #1031, opened 2023-04-20
[14.0][IMP] account_financial_report, added invoice date on aged partner balance, open items

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -119,6 +119,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         partner_ids,
         date_at_object,
         date_from,
+        date_by,
         only_posted_moves,
         show_move_line_details,
     ):
@@ -165,7 +166,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         move_lines = [
             move_line
             for move_line in move_lines
-            if move_line["date"] <= date_at_object
+            if (move_line[date_by] or move_line["date"]) <= date_at_object
             and not float_is_zero(move_line["amount_residual"], precision_digits=2)
         ]
         for move_line in move_lines:
@@ -352,6 +353,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         account_ids = data["account_ids"]
         partner_ids = data["partner_ids"]
         date_at = data["date_at"]
+        date_by = data["date_by"]
         date_at_object = datetime.strptime(date_at, "%Y-%m-%d").date()
         date_from = data["date_from"]
         only_posted_moves = data["only_posted_moves"]
@@ -367,6 +369,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
             partner_ids,
             date_at_object,
             date_from,
+            date_by,
             only_posted_moves,
             show_move_line_details,
         )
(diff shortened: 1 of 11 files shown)


JuanyDForgeflow (approved):

MiquelRForgeFlow:
Why is needed this field in the move lines?

ChrisOForgeFlow:
> Why is needed this field in the move lines?

Reference to this PR https://github.com/OCA/account-financial-reporting/pull/692

date and invoice date can be different on some cases

AaronHForgeFlow:
For me it is fine. Do you think it is useful to include a boolean field so users decide to get the invoice date or not?

ChrisOForgeFlow:
> For me it is fine. Do you think it is useful to include a boolean field so users decide to get the invoice date or not?

It is something that I thought to show it optionally, because I have also seen that the filters work with the date of the journal entry, and that way I could decide with which date field to filter

AaronHForgeFlow:
Functional tests LGTM. However I would like the user to have an option to decide whether to show the invoice date or not. For companies that have same invoice date than accounting date it is like too much information having to columns with the same date:



Is that complex to implement?

ChrisOForgeFlow:
> Functional tests LGTM. However I would like the user to have an option to decide whether to show the invoice date or not. For companies that have same invoice date than accounting date it is like too much information having to columns with the same date:
> 
> 
> 
> Is that complex to implement?

I will explore, and back to you with the result

ChrisOForgeFlow:
> > Functional tests LGTM. However I would like the user to have an option to decide whether to show the invoice date or not. For companies that have same invoice date than accounting date it is like too much information having to columns with the same date:
> > 
> > Is that complex to implement?
> 
> I will explore, and back to you with the result

Please check the change now @AaronHForgeFlow @MiquelRForgeFlow

account_financial_report/report/templates/open_items.xml
```diff
@@ -173,9 +188,15 @@
         <!-- # lines or centralized lines -->
         <div class="act_as_row lines">
             <!--## date-->
-            <div class="act_as_cell left">
+            <div class="act_as_cell left" t-if="date_by == 'date'">
                 <span t-raw="line['date'].strftime('%d/%m/%Y')" />
             </div>
+            <!--## invoice date-->
+            <div class="act_as_cell left" t-if="date_by == 'invioce_date'">
```
MiquelRForgeFlow:
```suggestion
            <div class="act_as_cell left" t-if="date_by == 'invoice_date'">
```

account_financial_report/report/aged_partner_balance.py
```diff
@@ -165,7 +166,7 @@ def _get_move_lines_data(
         move_lines = [
             move_line
             for move_line in move_lines
-            if move_line["date"] <= date_at_object
+            if (move_line[date_by] or move_line["date"]) <= date_at_object
```
MiquelRForgeFlow, line 169:
```suggestion
            if ((move_line[date_by] == "invoice_date" and move_line["invoice_date"]) or move_line["date"]) <= date_at_object
```
same in open_items.py

ChrisOForgeFlow: There are some cases where invoice_date field can be empty because that I use this comparison to avoid comparison error with boolean value

MiquelRForgeFlow: But this comparison is wrong, as `move_line[date_by]` it's always set. Also my suggestion is correct, because if invoice_date is empty, then `or move_line["date"]` will take its place.

ChrisOForgeFlow: But move_line[date_by] data depends of value of date_by, and date_by can be for this case "date" or "invoice_date" for data in first case I'm sure always retrieve move_line["date"] in second case move_line["invoice_date"] can be a False or None value, and or helps to get always move_line["date"], maybe I'm missing something?

PR #1033, opened 2023-04-27
[14.0][FIX] account_financial_report: Prevent error related to currency from General ledger

victoralmau:
Related to: https://github.com/OCA/account-financial-reporting/issues/1039

Prevent error related to currency from General ledger

Use case: Generate report (showing foreign currency) with accounts with defined currency.

Traceback:

```
File "/odoo/odoo-server/odoo/addons/base/models/ir_qweb_fields.py", line 448, in value_to_html
fmt = "%.{0}f".format(display_currency.decimal_places)
AttributeError: 'int' object has no attribute 'decimal_places'

The above exception was the direct cause of the following exception:

Error to render compiling AST
AttributeError: 'int' object has no attribute 'decimal_places'
Template: account_financial_report.report_general_ledger_lines
Path: /t/div/div[2]/t[6]/t[1]/div[1]/t[1]/span/t
Node: <t t-raw="account_or_group_item_object['init_bal']['bal_curr']" t-options="{'widget': 'monetary', 'display_currency': account['currency_id']}"/>

```
Please @pedrobaeza can you review it?

@Tecnativa TT42804

PR #1041, opened 2023-05-05
[16.0][FIX] account_financial_report: Prevent error related to currency fro…

ferran-S73:
…m General ledger

Use case: Generate report (showing foreign currency) with accounts with defined currency.

Traceback:

File "/odoo/odoo-server/odoo/addons/base/models/ir_qweb_fields.py", line 448, in value_to_html fmt = "%.{0}f".format(display_currency.decimal_places) AttributeError: 'int' object has no attribute 'decimal_places'

The above exception was the direct cause of the following exception:

Error to render compiling AST
AttributeError: 'int' object has no attribute 'decimal_places' Template: account_financial_report.report_general_ledger_lines Path: /t/div/div[2]/t[6]/t[1]/div[1]/t[1]/span/t
Node: <t t-raw="account_or_group_item_object['init_bal']['bal_curr']" t-options="{'widget': 'monetary', 'display_currency': account['currency_id']}"/>

TT42804

Cherry-pick and port from https://github.com/OCA/account-financial-reporting/pull/1033

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -531,7 +531,10 @@ class AbstractReportXslx(models.AbstractModel):
             format_amt = report_data["formats"]["format_amount"]
             field_prefix = "format_amount"
         if "currency_id" in line_object and line_object.get("currency_id", False):
-            currency = line_object["currency_id"]
+            if isinstance(line_object["currency_id"], int):
+                currency = self.env["res.currency"].browse(line_object["currency_id"])
+            else:
+                currency = line_object["currency_id"]
             field_name = "{}_{}".format(field_prefix, currency.name)
             if hasattr(self, field_name):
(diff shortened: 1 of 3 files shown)


ferran-S73:
Hi @victoralmau @GSLabIt @pedrobaeza I made a port to 16.0 of this fix, I would appreciate it if you could review. Thanks!

pedrobaeza:
There are still concerns in 14.0 PR. The problem is to avoid at maximum the code mess mixing variable names, or having non homogeneous structures you can't rely on.

ferran-S73:
Hi @victoralmau @pedrobaeza I updated this code from it's last version in 14.0. It should be ready for review now.

pedrobaeza:
Why 2 commits? Can you squash them?

ferran-S73:
> Why 2 commits? Can you squash them?

@pedrobaeza done, hope it's alright now

PR #1044, opened 2023-05-23
[15.0][FIX] account_financial_report: fixing calculation bug

oreju:
fixing bug when calculation of credit value is incorrect

### DIFF
--- a/account_financial_report/report/abstract_report.py
+++ b/account_financial_report/report/abstract_report.py
@@ -100,7 +100,6 @@ class AgedPartnerBalanceReport(models.AbstractModel):
                 else:
                     move_line["amount_residual_currency"] = debit_amount_currency[ml_id]
             if ml_id in credit_ids:
-                move_line["amount_residual"] -= credit_amount[ml_id]
                 if move_line.get("amount_residual", False):
                     move_line["amount_residual"] -= credit_amount[ml_id]
                 else:


pedrobaeza:
Without steps to reproduce the problem, this will be hardly accepted, as there's no support in your claim about what is fixing and why it's a bug.

oreju:
Hello @pedrobaeza ,

My solution is fixing problem when generating report for Aged Partner Balance. In my case I mentioned that residual amount was doubled. When I debugged code I mentioned that amount residual was subtracted twice. You can check it here -> https://github.com/OCA/account-financial-reporting/blob/15.0/account_financial_report/report/abstract_report.py#L103 this line should not be here because we have already code there in conditional statement that do correct calculation of credit -> https://github.com/OCA/account-financial-reporting/blob/15.0/account_financial_report/report/abstract_report.py#L104-L107 . 

If you look at previous version (v14) you will see that this line doesn't exist. Looks that it's migration bug. V14 -> https://github.com/OCA/account-financial-reporting/blob/14.0/account_financial_report/report/abstract_report.py#L103-L106

Let me know if you need more explanation. 

Thank you!

pedrobaeza (approved):
OK, I see that it comes from a bad cherry-pick here:

https://github.com/OCA/account-financial-reporting/commit/4df403a83385c983ea08eca13b3b109f224f4f66#diff-260f9ea3aa1ce82c648c78dc37fadbc5a2994d87814123220854711391474b34R103

So fast-tracking this:

/ocabot merge patch

Thanks for the PR

PR #1052, opened 2023-07-13
14.0 - FIX - Hide general ledger entries, when hide_at_0 flag is set and entries is grouped by partner

patrickt-oforce:
Before this fix, when hide_at_0 flag is set and general ledger is grouped by partner, the report show entries even if final balance of grouped `move_lines` is zero

Checking if balance amount of 'fin_bal' is 0, allow to hide that group of move lines
Fixed general ledger entries

When hide_at_0 flag is set and general ledger entries are grouped by partner, the item outlined in red should not be show on report, because the ending balance is zero, as indicated in help text of hide_at_0 flag
GEN_LED_WO_FIX

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -626,13 +626,9 @@ class GeneralLedgerReport(models.AbstractModel):
                     rec_after_date_to_ids,
                 )
                 group_item.update({"move_lines": move_lines})
-                if (
-                    hide_account_at_0
-                    and float_is_zero(
-                        data[data_id]["init_bal"]["balance"],
-                        precision_rounding=rounding,
-                    )
-                    and group_item["move_lines"] == []
+                if hide_account_at_0 and float_is_zero(
+                    group_item["fin_bal"]["balance"],
+                    precision_rounding=rounding,
                 ):
                     continue
                 list_grouped += [group_item]


matteoopenf (approved):

patrickt-oforce:
@OCA/accounting-maintainers can be reviewed functionally?

patrickt-oforce:
@OCA/accounting-maintainers i think ther is some problem with pre-commit, because after rebasing my branch on oca/14.0 HEAD and running it locally, i've all check pass; but in CI seems not found some module to import. Can you check and help me to find a solution?

matteoopenf:
t think the problem is that pr https://github.com/OCA/account-financial-reporting/pull/1082

matteoopenf (approved):

pedrobaeza:
I think the text should change to "Without activity", but if there are items in the period, the partner should be shown.

sergiocorato:
> I think the text should change to "Without activity", but if there are items in the period, the partner should be shown.

I agree, this option is not clearly explained, while this is: https://github.com/OCA/account-financial-reporting/blob/65af714370740439ced7828714c069cb42b0b28f/account_financial_report/wizard/trial_balance_wizard.py#L46

As this change would affect current behavior, the help should be fixed for this field and this one: https://github.com/OCA/account-financial-reporting/blob/7baad45572d0b43bf524417d52b90cb92b15c187/account_financial_report/wizard/open_items_wizard.py#L30 and another option like `hide_account_at_final_balance_0` should be created.

PR #1056, opened 2023-07-31
[12.0][FIX] account_financial_report: Support Postgres v14+

carmenbianca:
Postgres 14+ expects the argument anycompatiblearray instead of anyarray to array_cat. We detect the Postgres version and proceed with the correct type.

The refactoring in #660 would also fix this problem, and certainly fixes it for future Odoo versions, but that PR is marked do-not-merge.

Internal task: https://gestion.coopiteasy.be/web#view_type=form&model=project.task&id=10965&active_id=10965&menu_id=

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -482,11 +482,19 @@ WHERE report_trial_balance_account.account_group_id = computed.account_group_id
 
     def _add_account_group_account_values(self):
         """Compute values for report_trial_balance_account group in child."""
+        # Since Postgres 14, the argument to array_cat must be
+        # anycompatiblearray instead of anyarray. See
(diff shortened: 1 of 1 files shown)


carmenbianca:
For clarity, I have tested this against a Postgres 12 and a Postgres 14 database.

For easy testing, use:

```
podman run --name postgres[version_here] -e POSTGRES_HOST_AUTH_METHOD=trust -p [YOUR_PORT]:5432 postgres:[version_here]
createuser -p [YOUR_PORT] -U postgres -s $USER
```

and then set `db_port` in odoo.conf.

Alternatively, use Docker instead of podman. They have the same command-line arguments, I think.

robinkeunen (approved):
👍

remytms (approved):
LGTM

sergiocorato (approved):
LGTM

PR #953, opened 2022-11-22
[13.0][FIX] account_sale_stock_report_non_billed: Date can't be comparated with bool

CarlosRoca13:
cc @Tecnativa TT40573

ping @sergio-teruel @pedrobaeza

PR #1058, opened 2023-08-16
[15.0][FIX] account_sale_stock_report_non_billed: Date can't be comparated with bool

CarlosRoca13:
cc @Tecnativa TT44745

fw-port of: https://github.com/OCA/account-financial-reporting/pull/953

ping @pedrobaeza @stefan-tecnativa

### DIFF
--- a/account_sale_stock_report_non_billed/models/stock_move.py
+++ b/account_sale_stock_report_non_billed/models/stock_move.py
@@ -108,29 +108,30 @@ class StockMove(models.Model):
         ) * self.sale_line_id.price_reduce
 
     @api.depends("sale_line_id")
-    @api.depends_context("date_check_invoiced_moves")
+    @api.depends_context("non_billed_date")
     def _compute_not_invoiced_values(self):
         for move in self:
-            if not self.env.context.get("date_check_invoiced_moves"):
+            if not self.env.context.get("non_billed_date") or not self.env.context.get(
+                "non_billed_date_start"
+            ):
                 move.quantity_not_invoiced = 0
                 move.price_not_invoiced = 0
                 continue
-            date_start = self.env.context.get("date_check_invoiced_moves_start", False)
-            date_end = self.env.context.get("date_check_invoiced_moves", False)
-            if date_start:
-                date_start = fields.Date.from_string(date_start)
-            if date_end:
-                date_end = fields.Date.from_string(date_start)
+            date_start = self.env.context["non_billed_date_start"]
+            date_end = self.env.context["non_billed_date"]
             invoices_not_cancel = move.invoice_line_ids.filtered(
                 lambda l: l.move_id.state != "cancel"
             )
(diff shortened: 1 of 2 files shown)


account_sale_stock_report_non_billed/models/stock_move.py
```diff
@@ -127,6 +127,8 @@ def _compute_not_invoiced_values(self):
             moves_in_date = invoices_not_cancel.mapped("move_line_ids").filtered(
                 lambda m: m.date_done >= date_start and m.date_done <= date_end
             )
+            if not self.env.context.get("date_check_iterval_restrict", False):
+                date_start = False
```
pedrobaeza: I don't get why do you need this. Pass the minimum variables in context, and the absence of one of them means that no check should be done.

CarlosRoca13: Because we need the date_start and the date_end to filter the move_line_ids in any situation, but when the user doesn't want to use this dates as a range for invoices too the date_start has to be reset to False.

pedrobaeza: Yes, and that's why I say you only need to pass one variable: the start date. If not passed, then you use `.get(..., False)`, and being `False` means that no check has to be done.

CarlosRoca13: But I need the date_start to take the `moves_in_date`

pedrobaeza: OK, I'm getting it. They should be independent variables for avoiding that clunky logic. Pass these 3 possible variables to context:

- `non_billed_date`: Old `date_check_invoiced_moves`
- `non_billed_date_start`: Old `date_check_invoiced_moves_start`
- `non_billed_invoice_date_start`: This one is set to `self.stock_move_non_billed_threshold` if `self.interval_restrict_invoices`.

and you get each of them on the computing code, with no inter-dependencies between them.

account_sale_stock_report_non_billed/models/stock_move.py
```diff
@@ -108,29 +108,34 @@ def _set_not_invoiced_values(self, qty_to_invoice, invoiced_qty):
...
                 lambda l: l.move_id.state != "cancel"
             )
             moves_in_date = invoices_not_cancel.mapped("move_line_ids").filtered(
                 lambda m: m.date_done >= date_start and m.date_done <= date_end
             )
+            invoice_date_start = False
+            if self.env.context.get("non_billed_invoice_date_start", False):
+                invoice_date_start = fields.Date.from_string(
+                    self.env.context["non_billed_invoice_date_start"]
+                )
```
pedrobaeza: Since v12, the fields are already date type, so no need of `fields.Date.from_string`

```suggestion
            invoice_date_start = self.env.context.get("non_billed_invoice_date_start", False)
```

pedrobaeza (approved):
/ocabot merge patch

PR #1061, opened 2023-09-05
[16.0][ADD] account_journal_ledger_by_entry_number: Replaces Move Name with Entry Number in Journal Ledger

ghost:
After testing the Journal Ledger report, and considering:

* Spanish Journal Ledger must be deposited at an official entity which requires that entries must be chronological and correlative
* Auto-secuence is not generating itself in a strict correlative order (see image from runboat):
  * Acts as you have selected *Group by Journal* when it's not the case.
  * Shows a "00000" element which we are not sure should be appearing 
image
* Auto-secuence, even if we fix it to become correlative, needs the current entry number \[name\] to stablish a correspondence with the number it shows, and it's not static, which does not contribute anything in Spain
* We have decided to be using account-financial-tools /account_journal_general_sequence/ in our clients, which creates a new field entry_number in journal entries and even allows it to be renumbered,

We have created a module that overwrites the current column "Entry" in the report with ordered and correlative entry numbers allocated in the field \[entry_number\] instead of the ones in the field \[name\] that we think is more effective and is even kept in DB.

The result (sorry in advance for being in spanish, but columns are the same): 
image

We are open to any change or improvement so we encourage you to try it.

@pedrobaeza @HaraldPanten


pedrobaeza:
This feature is already present in current `account_financial_report` module, printing journal ledger, and selecting the auto-sequence.

HaraldPanten:
I think that Pedro is right. You're not exporting the Journal Ledger in the correct way.

When exporting the Journal Ledger:

- You don't have to group by journal.
- You have to order by date.

Then you'll export the Journal Ledger in a correct format. with the auto sequence in a correlative order.

I think that auto sequence doesn't need to be kept in the DB, as it doesn't offer any valuable information. Why should we keep it? Could you describe us a situation where we could need it?

THX

ghost:
After succesfully testing your suggestions, we are going to close this Pull Request.

Thank you @pedrobaeza @HaraldPanten

PR #1065, opened 2023-09-07
[15.0][FIX] account_sale_stock_report_non_billed: Date can't be comparated with bool

CarlosRoca13:
cc @Tecnativa TT44980

ping @pedrobaeza @chienandalu


pedrobaeza (approved):
Fast-tracking this for solving totally the problem, as tested in v13:

/ocabot merge patch

PR #1068, opened 2023-09-12
[16.0][FIX] account_financial_report: Fix currency bug on trial balance report

aritzolea:
Fixes trial balance report so it uses currency record for display_currency option instead record's ID. It was previously throwing the following error:

`Error while render the template`
`AttributeError: 'int' object has no attribute 'decimal_places'`
`Template: account_financial_report.report_trial_balance_line`
`Path: /t/div/t[3]/t[1]/t/t/div/span/t`
`Node: <t t-esc="balance[\'initial_currency_balance\']" t-options="{\'widget\': \'monetary\', \'display_currency\': balance[\'currency_id\']}"/>`

For the error to show up there needs to be a move in a different currency from the company's currency.

Edit: Added another commit which fixed initial balance value when using foreign currency

### DIFF
--- a/account_financial_report/report/templates/trial_balance.xml
+++ b/account_financial_report/report/templates/trial_balance.xml
@@ -559,6 +559,10 @@
                 <t t-if="not show_partner_details">
                     <t t-if="balance['type'] == 'account_type'">
                         <t t-if="balance['currency_id']">
+                            <t
+                                t-set="balance_currency"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please check fixes about this in previous versions and cherry-pick them.

ljsalvatierra-factorlibre (approved):

gerard-vacas (approved):

PR #1070, opened 2023-09-24
[16.0][MIG] mis_template_financial_report : Migration to 16.0

jguenat:
Resuming work from @hbrunn 
- Moved equity in a new kpi on liability side.
feedback welcome :)


hbrunn (changes requested):
it's not that simple unfortunately, because as it is, the horizontal display only works for the PDF export, but it should also be horizontal in the preview and the excel export

jguenat:
Yes indeed I overlooked that feature.
Not sure I have the time to work on it though... Maybe later

PR #1073, opened 2023-09-28
[16.0][FIX] account_financial_report: Empty format

SirAionTech:
Forward port of https://github.com/OCA/account-financial-reporting/pull/1048 with https://github.com/OCA/oca-port.


pedrobaeza (approved):
What about v15?

Fast-tracking as already merged in 14

/ocabot merge patch

SirAionTech:
> What about v15?
> 
> Fast-tracking as already merged in 14
> 
> /ocabot merge patch

I skipped it because I don't use it, but here you are https://github.com/OCA/account-financial-reporting/pull/1074

pedrobaeza:
Thanks, it's important to have all the patches aligned in all versions, specially if they are not hard to be fw-ported. We do the same although our customers won't be affected.

PR #1074, opened 2023-09-29
[15.0][FIX] account_financial_report: Empty format

SirAionTech:
Forward port of https://github.com/OCA/account-financial-reporting/pull/1048 with https://github.com/OCA/oca-port.

As requested in https://github.com/OCA/account-financial-reporting/pull/1073#pullrequestreview-1649264293.
I did not try it because I don't have a `15.0` environment.


pedrobaeza (approved):
As the code in the same in the 3 versions, it's safe enough to not do a full test, but just code review.

/ocabot merge patch

PR #1076, opened 2023-10-04
[MIG] mis_template_financial_report

SirAionTech:
Standard migration from `15.0` (https://github.com/OCA/account-financial-reporting/tree/08f6349402515c5d10759cf2bdf262378dd9fda9/mis_template_financial_report).

There is already an existing migration in https://github.com/OCA/account-financial-reporting/pull/1070 but based on 
> Yes indeed I overlooked that feature.
Not sure I have the time to work on it though... Maybe later

(from https://github.com/OCA/account-financial-reporting/pull/1070#issuecomment-1735502823)

I assume it's not going to work on it, so I am superseding it.

@jguenat @hbrunn would you mind having a look?


SirAionTech:
> it's not that simple unfortunately, because as it is, the horizontal display only works for the PDF export, but it should also be horizontal in the preview and the excel export

_Originally posted by @hbrunn in https://github.com/OCA/account-financial-reporting/pull/1070#pullrequestreview-1641264600_

Both issues are present in `15.0` too, and I don't think the migration is the right place to fix them.
About XLSX, it's also listed in the known issues of the module in `15.0` (see https://github.com/OCA/account-financial-reporting/tree/08f6349402515c5d10759cf2bdf262378dd9fda9/mis_template_financial_report#known-issues-roadmap).

hbrunn:
I'll probably never understand why people consider 'it was wrong in the past' a valid reason to do it wrong now. You'd have to rewrite all of this anyways for v16 with the conversion to owl, no matter what happened in v15.

Please consider merging https://github.com/hbrunn/account-financial-reporting/commit/9ff3b4fdaf5a442c6463754fbfec43b5ee433e9a to have feature parity with v12, xls is indeed out of scope for the migration, as that never existed. I made this more general so that for v17 it's simple to split off all the code to `mis_builder_horizontal`, and only keep the kpi definitions here.

SirAionTech:
> I'll probably never understand why people consider 'it was wrong in the past' a valid reason to do it wrong now.

Me neither, and I don't think that is what I am doing here: I am simply migrating the module as is.
I think that in modules migration the goal is to have the same module compatible with the new Odoo major version, with same functionality.
If the module had issues in the old version, the migration is not the moment to fix them: this should help speed up the migration to have the same functionality in the new version.

> Please consider merging https://github.com/hbrunn/account-financial-reporting/commit/9ff3b4fdaf5a442c6463754fbfec43b5ee433e9a to have feature parity with v12, xls is indeed out of scope for the migration, as that never existed. I made this more general so that for v17 it's simple to split off all the code to mis_builder_horizontal, and only keep the kpi definitions here.

Thanks! I'll look into it

pedrobaeza:
Superseded by #1144

PR #1089, opened 2023-12-20
[14.0][FIX] account_financial_report - show all accounts when group by none

ghost:
The general ledger does not show all accounts when grouping by none. 
The current behavior applies a filter on only receivables and payables.

image
image

Here is the result without the fix :

image

Here is the resultat with the fix : 

image

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -54,10 +54,11 @@ class GeneralLedgerReport(models.AbstractModel):
         return ["receivable", "payable"] if grouped_by != "taxes" else ["other"]
 
     def _get_acc_prt_accounts_ids(self, company_id, grouped_by):
-        accounts_domain = [
-            ("company_id", "=", company_id),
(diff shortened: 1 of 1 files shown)


account_financial_report/report/general_ledger.py
```diff
@@ -51,7 +51,12 @@ def _get_taxes_data(self, taxes_ids):
         return taxes_data
 
     def _get_account_internal_types(self, grouped_by):
-        return ["receivable", "payable"] if grouped_by != "taxes" else ["other"]
+        if grouped_by == "partners":
+            return ["receivable", "payable"]
+        elif grouped_by == "taxes":
+            return ["other"]
+        else:
+            return ["receivable", "payable", "other", "liquidity"]
```
pedrobaeza: Can you better return `[]` and change the code where this is used for avoiding that if any new type, this won't work?

ghost: I changed the fix a bit. The issue around new types should be good now

pedrobaeza (approved):

dreispt (approved):

PR #980, opened 2022-12-26
[15.0][FIX] account_financial_report: generate ledger document

Reyes4711-S73:
In general ledger report shows an error due to the fact that the ball_curr varible does not generated

In a runbot I test this:
**view general ledger**
image
image

**export xlsx**
image
image

PR #1090, opened 2024-01-03
[16.0][FIX] account_financial_report: generate ledger document

gurneyalex:

port of #980 to 16.0


pedrobaeza (approved):
/ocabot merge minor

PR #1091, opened 2024-01-03
[FIX] [16.0] account_financial_report: fixing bug when calculation of credit is duplicated

gurneyalex:
forward port of #1044 to 16.0


pedrobaeza (approved):
/ocabot merge patch

PR #1093, opened 2024-01-08
[14.0][FIX]  account_financial_report: reduce report url

adrip-s73:
When you try to get the report through the "general_ledger_wizard" with too many account created the report url is too long and raise the next error:

image

My fix is avoinding to pass all account parameter to the report and instead getting them from the wizard id data.

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -776,21 +776,22 @@ class GeneralLedgerReport(models.AbstractModel):
 
     def _get_report_values(self, docids, data):
         wizard_id = data["wizard_id"]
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why not doing the same for all the parameters as you are in there?

adrip-s73:
Okey i will make the change for all the parameters

mtbochaca (approved):
Also fix if we filter by partner.

pedrobaeza (approved):
Thanks for the improvement. This can be apply on the rest of reports, but for now it's ok:

/ocabot merge patch

Would you fw-port it to upper versions?

PR #1098, opened 2024-01-22
[15.0][FIX]account_financial_report: missing accounts

ChristophAbenthungCibex:
Accounts with leading 0s break the search that loads the accounts for trial balance reports.

Steps to reproduce:
1. Login
2. Go to 'Invoicing' -> 'Configuration' -> 'Chart of Accounts' ([Chart of Accounts](http://oca-account-financial-reporting-15-0-e8fb133409fc.runboat.odoo-community.org/web#cids=1&menu_id=119&action=241&model=account.account&view_type=list))
3. Create Button -> Field 'Code' has to start with at least a 0. E.g: 000123
Values for fields 'Account Name' and 'Type' don't matter
4. Go to 'Reporting' -> 'Trial Balance' (A wizard should open)
5. Select for the field 'From Code' an account with leading 0s in the code. E.g. 000123
6. Select for the 'to' field any account with code that starts with 2 or higher. E.g. 999999

In the list of accounts the first account code is 131000. In the view 'Chart of Accounts' can be seen that there are some accounts not loaded. This is because of the broken search. Those codes are stored as chars in the database. In the wizard they will be converted to integers. That casting removes the leading 0s. E.g. 000123 will be 123. In the search that integer will be casted to a char again. As a result, any account with code that is lexicographically smaller will not be loaded. Eg. codes that are 012345 or 100000 will not be loaded if the start code is 000123

chart_of_accounts
trial_balance_loaded_accounts

### DIFF
--- a/account_financial_report/tests/test_trial_balance.py
+++ b/account_financial_report/tests/test_trial_balance.py
@@ -31,6 +31,16 @@ class TestTrialBalanceReport(AccountTestInvoicingCommon):
         )
         cls.group2 = group_obj.create({"code_prefix_start": "2", "name": "Group 2"})
         # Set accounts
+        cls.account001 = cls._create_account_account(
+            cls,
+            {
+                "code": "001",
+                "name": "Account 001",
+                "user_type_id": cls.env.ref(
+                    "account.data_account_type_other_income"
+                ).id,
+            },
+        )
         cls.account100 = cls.company_data["default_account_receivable"]
         cls.account100.group_id = cls.group1.id
         cls.account110 = cls.env["account.account"].search(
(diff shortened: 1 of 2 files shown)


OKT96 (approved):
LGTM!

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions

ChristophAbenthungCibex:
> /ocabot merge patch
> 
> Please fw-port it to upper versions

@pedrobaeza I created a new Pull Request for version 16 (#1112) . I hope I have done the porting correctly. If not please let me know how to do it, I haven't really found guidelines on how to.

PR #1099, opened 2024-01-22
Add currency column to general_ledger excel export file

RaoufJabriCamptocamp:
- For some users working with different currencies, being able to filter report by the used currency is helpful. We add exporting the column as an option in the export wizard.


RaoufJabriCamptocamp:
ping @gurneyalex

PR #1107, opened 2024-02-07
[15.0][FIX] account_financial_report: error currency object

Saran440:
This PR fixed error run GL in view it error because object currency browse with `account['currency_id']` (object currency)
Selection_008


AaronHForgeFlow:
Can the error be reproduced in runboat?

PR #1110, opened 2024-02-15
[14.0][IMP] account_financial_report: Add Missing Partner support in trial balance

victoralmau:
Related to https://github.com/OCA/account-financial-reporting/issues/1097

Add Missing partner support in trial balance

If there are records without partner_id set and we show the detail of partners we need to take into account those records grouped as "Missing Partner".

Please @pedrobaeza can you review it?

@Tecnativa TT47771

PR #1111, opened 2024-02-15
[16][IMP] account_financial_report : Display "Partner missing" line in trial balance with partner details option

### LINKED ISSUE #1097: [16][account_financial_report] - Show Partner Details option does not display line with empty partners making the totals of the report inconsistent
With module account_financial_report, when printing the trial balance with the option "show partner details", the lines with no partner are not displayed, but are still counted for into the totals.

balance_wizard

balance

We can see that we have 0 lines with credit but the total is 100, it is very confusing. This is because I have created an accounting item for this account with no partner.

Before (I am not sure until when/which version) we has a line with the value "No partner allocated".
This way the total of the report  was consistent with the sum of all lines.
I am not sure if this behavior was removed on purpose or not and for which reason, but, if we don't want to display this "No partner allocated" line, I guess we'd want to remove the balance/credit/debit of these line without partner from the totals.

I personally liked the "No partner allocated" behavior and could not find why it was removed.

Anyway, I could try to fix it one way or another (adding the No partner allocated line or removing the no partner amounts from totals) but first I need to be sure on the right way to go

What do you think @alexis-via @pedrobaeza @JordiBForgeFlow

florian-dacosta:
Aims to fix https://github.com/OCA/account-financial-reporting/issues/1097

@pedrobaeza @alexis-via

### DIFF
--- a/account_financial_report/i18n/account_financial_report.pot
+++ b/account_financial_report/i18n/account_financial_report.pot
@@ -1108,6 +1108,14 @@ msgid "Limit hierarchy levels"
 msgstr ""
 
 #. module: account_financial_report
+#. odoo-python
+#: code:addons/account_financial_report/report/trial_balance.py:0
+#: code:addons/account_financial_report/report/trial_balance.py:0
+#, python-format
+msgid "Missing Partner"
+msgstr ""
+
+#. module: account_financial_report
 #: model:ir.model.fields,field_description:account_financial_report.field_journal_ledger_report_wizard__move_target
 msgid "Move Target"
 msgstr ""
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Isn't this the same as #1110 ?

florian-dacosta:
As a matter of fact it is !
I did not see the PR, I had plan to do it since I have created the related issue but just had the time now.

Not sure why the other is in draft state, but we can go ahead with the other one, should I close this one and make a new PR, cherry-picking the commit of the other PR?

pedrobaeza:
Please test it, and if it's OK, we will make the fw-ports to all versions.

florian-dacosta:
@pedrobaeza Ok, I've tested the other PR and approved it, I'll close this one then, waiting for the fw-ports

PR #1112, opened 2024-02-19
[16.0][FIX]account_financial_report: missing accounts

ChristophAbenthungCibex:
Porting fix from Version 15 to Version 16. See original Pull Request #1098 for further information.


pedrobaeza:
Can you please squash both commits into one?

ChristophAbenthungCibex:
> Can you please squash both commits into one?

Done. Sorry was not sure if it would break something if i squash a merge commit with another.

PR #1117, opened 2024-03-04
[16.0] [FIX] account_financial_report: show currency name instead of dictionary

eLBati:
Steps:

 - Accounting > Reporting > OCA accounting reports > Journal Ledger
 - Select Date range and activate Foreign Currency

Get "{'name': 'EUR'} " in currency field

Desired: get EUR

### DIFF
--- a/account_financial_report/report/templates/journal_ledger.xml
+++ b/account_financial_report/report/templates/journal_ledger.xml
@@ -302,7 +302,9 @@
         <t t-if="display_currency">
             <div class="act_as_cell" name="currency_name">
                 <t t-if="move_line['currency_id']">
-                    <span t-esc="currency_ids_data.get(move_line['currency_id'], '')" />
+                    <span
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Maybe the solution would be to not put in first place the whole content in the dictionary? cc @victoralmau

victoralmau:
I understand error and fix.

Currently there are many dictionaries that could be removed in Journal Ledger: 
- account_ids_data
- partner_ids_data
- currency_ids_data
- move_ids_data
- tax_line_data
- move_line_ids_taxes_data 

and thus reduce and simplify the code. Do you think this change is necessary?

victoralmau (approved):

pedrobaeza:
pre-commit is broken. Not related to this module, but it would be great if you fix it (here or in other PR).

victoralmau:
Pre-commit is now fixed if we do a rebase.

pedrobaeza:
OK, then we can launch the command that does the rebase on merge:

/ocabot merge patch

Issue #1119, opened 2024-03-08
[14.0] account_financial_report : The invoicing manager can not cretae date range and  date range type


abenzbiria:
##Context

The OCA module adds the date.range model which ONLY allows users in the Administration / Configuration group to:

+ Create different types of date ranges

+ Defining date ranges on reports (general balance example)

The module account_financial_report adds a set of financial reports  where we can use the "Date Range" field which allows you to select or create a date range.

![gl_wiazrd](https://github.com/OCA/account-financial-reporting/assets/3594109/85fd8f5e-1197-4dad-a6a4-72972241284e)

## Module
account_financial_report

## Describe the bug

Users in the Invoicing Manager group can not create a date range and date type

## To Reproduce

**Versions : 12 and 14** 

1. Log in with an invoicing manager
2- Edit the Invoicing/reporting/General Ledger

![Capture d’écran du 2024-03-08 17-52-42](https://github.com/OCA/account-financial-reporting/assets/3594109/0d031904-b4ef-4d27-994d-05fa979bbce9)

3. I can not create a range date for printing the GL report

![gl_wiz](https://github.com/OCA/account-financial-reporting/assets/3594109/485f66f2-1073-441a-9c8d-5c64a632e4b1)

**Expected behavior**

As an Invoicing Manager I can create date range and types from the GL wizard.

PR #1133, opened 2024-03-15
[14.0][FIX] account_financial_report: Date Range Access

majouda:
- This PR is for the issue number #1119 
- Grant access to account manager to configure date range and date range type from different reports wizards


majouda:
Hello @pedrobaeza
Can you please review this PR?
Thank you.

pedrobaeza:
You should use https://odoo-community.org/shop/date-range-account-10136?#attr=17382 instead.

Issue #1135, opened 2024-03-18
[16.0][Account_financial_reporting] General ledger doesn´t group by taxes


Jesarregui:
Summary:
When its generated, the general ledger report, in Odoo version 16 it is not grouped by taxes.

Modules:
[Account financial report](https://github.com/OCA/account-financial-reporting/tree/16.0/account_financial_report)

Describe the bug:
Generate a general ledger report configuring a tax account and grouping it by taxes. It looks like the following screenshot:
![V16_general_ledger](https://github.com/OCA/account-financial-reporting/assets/153611478/4cc015d8-58e3-4d69-b7c6-8a95f0de37f2)

Expected behaviour:

The correct operation would be, however, like the following screenshot taken on a v15

![v15_general_ledger](https://github.com/OCA/account-financial-reporting/assets/153611478/f051d0de-8157-452c-9e64-5143787ad729)

To reproduce:

Generate a general ledger type report with the following parameters:

- Includes an accounting account with tax.
- It is grouped by taxes.

HaraldPanten:
@pedrobaeza @victoralmau  Did you know that? THX

HaraldPanten:
Closing, solved in https://github.com/OCA/account-financial-reporting/pull/1136

PR #1136, opened 2024-03-19
[16.0][FIX] account_financial_report: Filter correctly by account_type in General ledger

victoralmau:
Related to https://github.com/OCA/account-financial-reporting/issues/1135 and https://github.com/OCA/account-financial-reporting/commit/36c5b3580c94bcced885530ae8e0a9edfbb17a5d#diff-dbf187923ecd83ab6d1624836fe8ae266f0edc25c24f207d28f8bd9b879b8017

Filter correctly by account_type in General ledger

In v15 it was filtered by the internal_type field (type of account.account.type) to use "other" (most of the accounts had it set). Now in v16, there is the account_type field exists and has to be compatible. The in or not in operator is set to "simplify" the domain and avoid having to set almost all types.

@Tecnativa

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -50,18 +50,19 @@ class GeneralLedgerReport(models.AbstractModel):
             )
         return taxes_data
 
-    def _get_account_internal_types(self, grouped_by):
-        return (
-            ["asset_receivable", "liability_payable"]
-            if grouped_by != "taxes"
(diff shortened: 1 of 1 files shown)


Jesarregui (approved):
Functional test: LGTM

Selección_4401

HaraldPanten (approved):
Functional review by our team.

pedrobaeza (approved):
/ocabot merge patch

PR #1137, opened 2024-03-19
[REF] account_tax_balance: Allow to override query

SirAionTech:
After this PR, other modules can edit the query or its parameters before execution.

In our use case this is useful because we want to gather the moves by another field instead of `date`.

### DIFF
--- a/account_tax_balance/models/account_tax.py
+++ b/account_tax_balance/models/account_tax.py
@@ -32,17 +32,10 @@ class AccountTax(models.Model):
             context.get("target_move", "posted"),
         )
 
-    def _account_tax_ids_with_moves(self):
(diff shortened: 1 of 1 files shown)


rvalyi (changes requested):
the docstring doesn't match anymore with what the extracted method does.

SirAionTech:
> the docstring doesn't match anymore with what the extracted method does.

Thanks for having a look!
I moved the docstring in the extracted method because I thought it was describing the query structure, but indeed the extracted method does not return any ID.
I restored the docstring where it was before, let me know if it looks better now

rvalyi (approved):
yes I think it's better because it properly describes was is returned by the method.

sergiocorato (approved):

PR #1141, opened 2024-04-05
[15.0][IMP] account_financial_report: Fix VAT tax tag report

dhx:
Currently the VAT report does not work when a tax has multiple account tags in the distribution as the account move line balance will end up in ALL tags (e.g. for the Reverse Charge Tax in Austria this leads to a zero balance in the tax column of the tags "+KZ 057" and "+KZ 066" / as the entries will cancel each other).

This fixes this issue by only using the balance of account move lines for linked tax tags also taking the tax_tag_invert flag into account.

### DIFF
--- a/account_financial_report/README.rst
+++ b/account_financial_report/README.rst
@@ -53,10 +53,6 @@ currency balances are not available.
 Known issues / Roadmap
 ======================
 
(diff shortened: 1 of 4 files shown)


dhx:
removed handling of the tax_tag_invert field -> https://github.com/OCA/account-financial-reporting/pull/1143

PR #1146, opened 2024-04-15
[15.0][FIX] account_financial_report: Fix intervals calculations

carolinafernandez-tecnativa:
- Fix error in intervals calculations.

-Case use error: filtering the aged report by the partner "Test customer", and putting the reference date at 04/18/2024 (English US date format). It should be in the column "30", but it appears in the column "2".

@Tecnativa
TT45574

@pedrobaeza @pilarvargas-tecnativa

### DIFF
--- a/account_financial_report/i18n/account_financial_report.pot
+++ b/account_financial_report/i18n/account_financial_report.pot
@@ -1841,7 +1841,6 @@ msgid "future"
 msgstr ""
 
 #. module: account_financial_report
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Pending on the modifications in 16.0

pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1148, opened 2024-04-29
[15.0][IMP] account_financial_report: Change font size

pilarvargas-tecnativa:
Set a larger font style for the html report with a relative font size to make it responsive. For this, a new html_container is defined so that this does not affect the other reports and the font style is customised only for the report in html view, without affecting the PDF print.

Instead of injecting the style in the html_container call, a new template is defined and the css is included to make it more maintainable.

cc @Tecnativa TT48995

@pedrobaeza @victoralmau please review

### DIFF
--- a/account_financial_report/static/src/css/report_html.css
+++ b/account_financial_report/static/src/css/report_html.css
@@ -1,0 +1,7 @@
+/* Styles of html report */
+.account_title span {
+    font-size: 1rem !important;
+}
+.act_as_cell {
+    font-size: 0.9rem !important;
+}
--- a/account_financial_report/report/templates/aged_partner_balance.xml
+++ b/account_financial_report/report/templates/aged_partner_balance.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <template id="aged_partner_balance">
-        <t t-call="web.html_container">
+        <t t-call="account_financial_report.html_container">
             <t t-foreach="docs" t-as="o">
                 <t t-call="account_financial_report.internal_layout">
                     <t
--- a/account_financial_report/report/templates/general_ledger.xml
+++ b/account_financial_report/report/templates/general_ledger.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <template id="general_ledger">
-        <t t-call="web.html_container">
+        <t t-call="account_financial_report.html_container">
             <t t-foreach="docs" t-as="o">
                 <t t-call="account_financial_report.internal_layout">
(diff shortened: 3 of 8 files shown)


pilarvargas-tecnativa:
Before the change there was no difference in font size between the html report and the printouts unless zoomed in. Now the change is visible as soon as the report is viewed.

image

pilarvargas-tecnativa:
@pedrobaeza can you check please?

pedrobaeza:
Trial balance is still using the old font size it seems.

pilarvargas-tecnativa:
> Trial balance is still using the old font size it seems.

I review it

carlosdauden:
A template should be added for more flexibility in selecting which fields to display.
You can draw inspiration from:
https://github.com/Tecnativa/tms/blob/15.0/tms_reporting/models/tms_print_template.py
https://github.com/Tecnativa/tms/blob/15.0/tms_reporting/views/report_account_invoice.xml#L12-L46

pedrobaeza:
IMO that's out of scope of this PR. It can be an enhancement, but take into account that this report requires mostly all the fields that are included. Without one of them, the information will be partial.

carlosdauden:
The analytical account is displayed even if the option is not selected in the wizard.
Displaying only the fields that are of interest to the user is the best way to have a clean view of the information.

pedrobaeza:
I think that can be considered a bug, as it shouldn't be shown if the check is not marked. That can be a regression on one of the multiple fixes that this module has suffered. But as said, this is out of scope of this PR, that intends another thing. Please put an issue about the analytic thing.

account_financial_report/static/src/css/report_html.css
```diff
@@ -0,0 +1,7 @@
+/* Styles of html report */
+.account_title span {
+    font-size: 1rem;
+}
+.act_as_cell {
+    font-size: 0.9rem;
```
pilarvargas-tecnativa:
```suggestion
    .account_title span {
        font-size: 1rem !important;
    }
    .act_as_cell {
        font-size: 0.9rem !important;
```


> Trial balance is still using the old font size it seems.

It has a direct style so it takes precedence over the style applied to the web template, you could include a ‘!important’ in the css of the web template.

https://github.com/OCA/account-financial-reporting/blob/15.0/account_financial_report/report/templates/trial_balance.xml#L48

pilarvargas-tecnativa:
@pedrobaeza Change made to the font size to apply to all including the trial balance in the web report respecting the sizes in the printed reports.

image

pedrobaeza (approved):

victoralmau (approved):

etobella (approved):

pedrobaeza:
Please fw-port it to 16 and 17

PR #1149, opened 2024-05-02
[16.0][FIX] account_financial_report: fix error false due_date

Saran440:
This PR fixed error can't run report aged partner balance when create JE directly.

Step to test
1. Go to menu Invoicing > Accounting > Journal Entries
2. Create new JE > Save
3. Run report `Aged Partner Balance` by filter account payable. it can't run report

Error Code: 
image

`TypeError: unsupported operand type(s) for -: 'datetime.date' and 'bool'`

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -56,6 +56,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         if not due_date or today <= due_date:
             ag_pb_data[acc_id]["current"] += residual
             ag_pb_data[acc_id][prt_id]["current"] += residual
+            due_date = today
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
maybe checking there is no due_date before setting a value?

pedrobaeza (approved):

pedrobaeza:
I suppose this one should be also be fw-ported to v17

Saran440:
@AaronHForgeFlow yes, it error because `account.move.line` don't have a value `due_date` (date_maturity) in the view Journal Entries. It invisible field.
Selection_001


@pedrobaeza Thank you for advise, I fwd to v17 https://github.com/OCA/account-financial-reporting/pull/1152

PR #1151, opened 2024-05-02
[15.0][IMP] account_financial_report: Add grouping by analytical account in Trial balance

victoralmau:
Changes done: 
- [x] Change t-esc and t-raw to t-out fromTrial Balance report
- [x] Refactor from Trial Balance
- [x] Add grouping by analytical account in Trial balance

ejemplo

@Tecnativa TT48969

### DIFF
--- a/account_financial_report/report/templates/trial_balance.xml
+++ b/account_financial_report/report/templates/trial_balance.xml
@@ -18,9 +18,9 @@
         <!-- Defines global variables used by internal layout -->
         <t t-set="title">
             Trial Balance -
-            <t t-raw="company_name" />
+            <t t-out="company_name" />
             -
-            <t t-raw="currency_name" />
+            <t t-out="currency_name" />
         </t>
         <t t-set="company_name" t-value="Company_Name" />
         <!--       <t t-set="res_company" t-value="company_id"/>-->
@@ -28,7 +28,7 @@
             <div class="row">
                 <h4
                     class="mt0"
-                    t-esc="title or 'Odoo Report'"
+                    t-out="title or 'Odoo Report'"
                     style="text-align: center;"
                 />
             </div>
@@ -92,8 +92,8 @@
                                 view-type="form"
                             >
                                         <t
-                                    t-raw="accounts_data[account_id]['code']"
-                                /> - <t t-raw="accounts_data[account_id]['name']" />
+                                    t-out="accounts_data[account_id]['code']"
+                                /> - <t t-out="accounts_data[account_id]['name']" />
                             </span>
                         </div>
                         <div class="act_as_table data_table" style="width: 100%;">
@@ -139,9 +139,9 @@
             <div class="act_as_row">
                 <div class="act_as_cell">
                     From:
-                    <span t-esc="date_from" t-options="{'widget': 'date'}" />
+                    <span t-out="date_from" t-options="{'widget': 'date'}" />
                     To
-                    <span t-esc="date_to" t-options="{'widget': 'date'}" />
+                    <span t-out="date_to" t-options="{'widget': 'date'}" />
                 </div>
                 <div class="act_as_cell">
                     <t t-if="only_posted_moves">All posted entries</t>
@@ -154,7 +154,7 @@
                 <div class="act_as_cell">
                     <t t-if="limit_hierarchy_level">
                         Level
-                        <span t-esc="show_hierarchy_level" />
+                        <span t-out="show_hierarchy_level" />
                     </t>
                     <t t-if="not limit_hierarchy_level">No limit</t>
                 </div>
@@ -220,7 +220,7 @@
                             res-model="account.account"
                             view-type="form"
                         >
-                            <t t-esc="balance['code']" />
+                            <t t-out="balance['code']" />
                         </span>
                     </div>
                     <!--            ## Account/Partner-->
@@ -230,7 +230,7 @@
                             res-model="account.account"
                             view-type="form"
                         >
-                            <t t-esc="balance['name']" />
+                            <t t-out="balance['name']" />
                         </span>
                     </div>
                 </t>
@@ -242,7 +242,7 @@
                             res-model="account.group"
                             view-type="form"
                         >
-                            <t t-raw="balance['code']" />
+                            <t t-out="balance['code']" />
                         </span>
                     </div>
                     <div class="act_as_cell left" t-att-style="style">
@@ -252,7 +252,7 @@
                             res-model="account.group"
                             view-type="form"
                         >
-                            <t t-esc="balance['name']" />
+                            <t t-out="balance['name']" />
                         </span>
                     </div>
                 </t>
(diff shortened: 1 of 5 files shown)


victoralmau:
Ping @pedrobaeza

account_financial_report/report/trial_balance.py
```diff
@@ -22,6 +22,7 @@ def _get_initial_balances_bs_ml_domain(
         date_from,
         only_posted_moves,
         show_partner_details,
+        analytic_account_id,
```
pedrobaeza: I think this method was not well thought, as passing more and more arguments and build the domain inside is a pain for adding things like this one. Why not taking the occasion to refactor this into `_get_initial_balances_bs_ml_domain(self, extra_domain=None):` and build the extra_domain in the calling method. In fact, that domain can be used across several calls that you are also overriding.

You can do the refactor in an extra commit before the analytic account grouping option.

account_financial_report/wizard/trial_balance_wizard.py
```diff
@@ -67,6 +67,15 @@ class TrialBalanceReportWizard(models.TransientModel):
         comodel_name="account.account",
         help="Ending account in a range",
     )
+    grouped_by = fields.Selection(
+        selection=[("", "None"), ("analytic_account", "Analytic Account")],
```
pedrobaeza: This is problematic on next versions, so please put `"none"` or leave it non required (as it is) without default and don't add the "None" option.

account_financial_report/wizard/trial_balance_wizard_view.xml
```diff
@@ -25,6 +25,7 @@
                         </group>
                         <group name="other_filters">
                             <field name="target_move" widget="radio" />
+                            <field name="grouped_by" />
```
pedrobaeza: Only visible if analytic accounting enabled.

pedrobaeza (changes requested):
On a test on runboat, all the groups by analytic accounting are putting the "Undistributed loss/gain results" with value 0 (account 129 in Spanish CoA, 9999999 in the sample CoA), and even more as I checked the option "Hide accounts at 0" (but it should be independent - it mustn't never appear).

account_financial_report/report/trial_balance.py
victoralmau: Complete refactoring done. 
Now I will continue with the analytic account grouping option (in an extra commit).

victoralmau:
Changes done. Complete refactoring and added the option of grouping by analytical account in an extra commit.

victoralmau:
Finally all changes are done and tested in a customer use case.

account_financial_report/tests/test_trial_balance.py
```diff
@@ -501,16 +527,18 @@ def test_03_partner_balance(self):
         )
 
         # Re Generate the trial balance line
-        res_data = self._get_report_lines(with_partners=True)
-        total_amount = res_data["total_amount"]
+        tb_wizard, res_data = self._get_report_lines(with_partners=True)
+        self._test_reports_trial_balance(tb_wizard)
+
```
pedrobaeza: Please don't add empty lines inside a method (this applies to all the places).

account_financial_report/tests/test_trial_balance.py
```diff
@@ -444,16 +469,17 @@ def test_03_partner_balance(self):
...
 
         # Re Generate the trial balance line
-        res_data = self._get_report_lines(with_partners=True)
-        total_amount = res_data["total_amount"]
+        _, res_data = self._get_report_lines(with_partners=True)
+
+        trial_balance = res_data["trial_balance"]
         check_partner_receivable = self.check_partner_in_report(
-            self.account100.id, self.partner.id, total_amount
+            self.account100.id, self.partner.id, trial_balance
```
pedrobaeza: Can you use fixed amounts? These tests are wicked this way.

pedrobaeza (commented):
The refactoring is not successful, as the idea is to reduce the number or methods and arguments, and I still see a lot of both. When I told you https://github.com/OCA/account-financial-reporting/pull/1151#discussion_r1622302974, it was to remove all the parameters except one. And I think most of the domains can be shared and reduced, as it's the same over and over...

account_financial_report/tests/test_trial_balance.py
victoralmau: Fixed amounts? This was (and is) trying to pull the corresponding report line filtering by some data (account and partner).

victoralmau:
> The refactoring is not successful, as the idea is to reduce the number or methods and arguments, and I still see a lot of both. When I told you [#1151 (comment)](https://github.com/OCA/account-financial-reporting/pull/1151#discussion_r1622302974), it was to remove all the parameters except one. And I think most of the domains can be shared and reduced, as it's the same over and over...

This refactoring is not enough? If you prefer i can delete the report completely and do it again, or change it as much as you want with many more hours, but the problem will always be the same, there is **NOT** enough data (neither in the tests nor creating them manually) to test that all the use cases work correctly and nothing is broken, so you always have to test these changes in "_real environments_", the same happens with other reports, it is absolutely frustrating.

victoralmau:
New approach done (3rd) with a clear objective: not to change the existing behavior, only to add a new one.

Changes done (and the reason):
- Added `grouped_by` field in wizard to group by analytical account.
- Fields are added to the groupby when you want to group and queries are done to get the data by analytical accounts.
- Set the grouped data in `total_amount` (to use these data later).
- The data currently used in the report: trial_balance and total_amount have a fixed structure on account that we do not want to change (we want to keep the existing operation so far), set `trial_balance_grouped` to obtain the data grouped by analytical account and account + `total_amount_grouped` (to represent the sum of all totals). They are also used in the XLS export.
- A different way has been set in template when you want to group to show the data.
- It has been set in the template to add `analytic_account_id` to the domain when clicking on a data if we are grouping.

account_financial_report/report/templates/trial_balance.xml
```diff
@@ -36,43 +36,94 @@
             <t t-call="account_financial_report.report_trial_balance_filters" />
             <div class="act_as_table list_table" style="margin-top: 10px;" />
             <!-- Display account lines -->
+            <t t-set="aml_domain_extra" t-value="[]" />
```
pedrobaeza, line 39: Instead of having to add this extra domain everywhere, can't we add the extra bits in the already present `domain` variable?

acysos:
Hello,

Funcionalmente es correcto. I have tested it.

Greetings

acysos:
Hello,

Functionally it is correct. I have tested it.

Greetings

account_financial_report/report/templates/trial_balance.xml
victoralmau: I have changed to add `aml_domain_extra` only where it is really needed.

Just for clarification, you want to then add all this code:
```
<t
    t-set="aml_domain_extra"
    t-if="grouped_item['id'] > 0"
    t-value="[('analytic_account_id', '=', grouped_item['id'])]"
/>
<t
    t-set="aml_domain_extra"
    t-else=""
    t-value="[('analytic_account_id', '=', False)]"
/>
```

in the **7 places** where `aml_domain_extra` is currently added (in this case I have set global variable precisely to reduce code and do the same thing that `aml_domain_common` does for example).

victoralmau:
Ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge major

PR #1154, opened 2024-05-03
[14.0][IMP] account_financial_report: VAT report fields translation

toita86:
Currently in the VAT report, the date does not correspond to the localization (it should appear as 'DD-MM-AAA' if the localization is set in Italy for ex.). Additionally,  the column `based_on` is not appearing with the translated label.
With this improvement things work based on the localization.


account_financial_report/report/templates/vat_report.xml
```diff
@@ -76,7 +76,7 @@
                                 t-raw="tag_or_group['net']"
                                 t-options="{'widget': 'monetary', 'display_currency': res_company.currency_id}"
                             />
-                        </div>
+                            </div>
```
HekkiMelody: **chore**: I think previous formatting was correct. In general, to keep the PR clean, we tend to exclude modifications unrelated to the PR at hand. Formatting fixes can be done in their own commit, if necessary, but in this case they should be excluded.

HekkiMelody (changes requested):
Code review: LGTM, just clean up the changes please.

account_financial_report/report/templates/vat_report.xml
toita86: Fixed right now, thank you for pointing this out!

HekkiMelody (approved):
Code review, LGTM

francesco-ooops (approved):
Functional ok!

@simahawk what do you think?

francesco-ooops:
@OCA/accounting-maintainers anyone available for review/merge? Thanks!

francesco-ooops:
@pedrobaeza could you check? thanks!

pedrobaeza (commented):
All the dates thing is unrelated to the PR title. It should be explained and isolated in a commit.

francesco-ooops:
@pedrobaeza should be fixed now

pedrobaeza:
The commit message about the dates is still no explaining why these changes.

toita86:
@pedrobaeza I have updated the commit message, let me know if its ok.

pedrobaeza:
I still don't see any explanation in the commit message:

imagen

Why do you do `fields.Date.from_string`? Steps to reproduce any problem or issue?

toita86:
My bad, I have added the commit description explaining what was the problem and the fix for it.

pedrobaeza (approved):
/ocabot merge minor

PR #1158, opened 2024-05-07
[FIX]-Can not print xlsx report for Activity Statement, Outstanding S…

hieulucky111:
We will face the problem with exporting Xlsx for Activity Statement, Outstanding Statement and Detailed Activity Statement if Wizard ID is not match with ID of any partner.
image


MiquelRForgeFlow (approved):
Thanks

AaronHForgeFlow (approved):

PR #1161, opened 2024-05-07
[16.0][FIX]  partner_statement: can not print xlsx reports (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/account-financial-reporting/pull/1158.

### DIFF
--- a/partner_statement/wizard/activity_statement_wizard.py
+++ b/partner_statement/wizard/activity_statement_wizard.py
@@ -47,13 +47,14 @@ class ActivityStatementWizard(models.TransientModel):
             report_name = "p_s.report_activity_statement_xlsx"
         else:
             report_name = "partner_statement.activity_statement"
+        partners = self.env["res.partner"].browse(self._context["active_ids"])
         return (
             self.env["ir.actions.report"]
             .search(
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (approved):

AaronHForgeFlow:
/ocabot merge patch

Thank you

partner_statement/wizard/activity_statement_wizard.py
```diff
@@ -47,13 +47,14 @@ def _print_report(self, report_type):
             report_name = "p_s.report_activity_statement_xlsx"
         else:
             report_name = "partner_statement.activity_statement"
+        partners = self.env["res.partner"].browse(self._context["active_ids"])
```
rousseldenis, line 50: @MiquelRForgeFlow Don't forget that _context is deprecated :smiley: 

https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L5030

MiquelRForgeFlow: Well, when these backward-compatibility things get removed, then the code will be adapted haha

PR #1163, opened 2024-05-07
[16.0][ADD] partner_statement, add report current statement

ChrisOForgeFlow:
@ForgeFlow

cc @JordiBForgeFlow


partner_statement/report/current_statement_xlsx.py
```diff
@@ -0,0 +1,284 @@
+# Author: Christopher Ormaza
+# Copyright 2021 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, fields, models
+
+from odoo.addons.report_xlsx_helper.report.report_xlsx_format import FORMATS
+
+
+class OutstandingStatementXslx(models.AbstractModel):
```
MiquelRForgeFlow, line 10:
```suggestion
class OutstandingStatementXslx(models.AbstractModel):
```
...

partner_statement/wizard/current_statement_wizard.py
```diff
@@ -0,0 +1,29 @@
+from odoo import models
+
+
+class CurrentStatementWizard(models.TransientModel):
+    """Outstanding Statement wizard."""
```
MiquelRForgeFlow, line 5: ...

partner_statement/wizard/current_statement_wizard.py
```diff
@@ -0,0 +1,29 @@
...
+            report_name = "p_s.report_current_statement_xlsx"
+        else:
+            report_name = "partner_statement.current_statement"
+        return (
+            self.env["ir.actions.report"]
+            .search(
+                [("report_name", "=", report_name), ("report_type", "=", report_type)],
+                limit=1,
+            )
+            .report_action(self, data=data)
```
MiquelRForgeFlow, line 24: apply https://github.com/OCA/account-financial-reporting/commit/942a5f911f25b5b68f862cc9ce5a2c13f716ea4a here too

partner_statement/report/current_statement.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class CurrentStatement(models.AbstractModel):
+    """Model of Current Statement"""
+
+    _inherit = "statement.common"
+    _name = "report.partner_statement.current_statement"
+    _description = "Partner Current Statement"
+
+    def _display_lines_sql_q1(self, partners, date_end, account_type):
```
MiquelRForgeFlow, line 14:
```suggestion
    def _display_current_lines_sql_q1(self, partners, date_end, account_type):
```
same for the other methods

partner_statement/report/current_statement.py
```diff
@@ -0,0 +1,160 @@
...
+            CASE WHEN (l.currency_id is not null AND l.amount_currency > 0.0)
+                THEN avg(l.amount_currency)
+                ELSE avg(l.debit)
+            END as debit,
+            CASE WHEN (l.currency_id is not null AND l.amount_currency < 0.0)
+                THEN avg(l.amount_currency * (-1))
+                ELSE avg(l.credit)
+            END as credit,
+            CASE WHEN l.balance > 0.0
+                THEN round(l.balance - sum(coalesce(pd.amount, 0.0)),2)
```
MiquelRForgeFlow, line 34: why apply round? if you have a good reason, then apply it to other reports too, but my opinion is not to apply it in the queries, as maybe you have a different rounding instead of 2. I think it should be applied somewhere in the python part of the code.

partner_statement/report/__init__.py
```diff
@@ -5,3 +5,5 @@
 from . import activity_statement_xlsx
 from . import detailed_activity_statement_xlsx
 from . import outstanting_statement_xlsx
+from . import current_statement
+from . import current_statement_xlsx
```
MiquelRForgeFlow, line 9: Is there too much difference between current_statement and outstanding_statement? If not, perhaps it's better to make current_statement be an inherit of outstanding_statement, the same way detailed_activity_statement inherits activity_statement. Surely it would reduce code.


Or maybe it should just get merged into the outstanding_statement by adding an extra boolean parameter.

MiquelRForgeFlow (commented):
Needs update of partner_statement/security/ir.model.access.csv and also add views

Issue #1103, opened 2024-01-29
[15.0][partner_statement]Ending Balance mismatch in 'Detailed activity statement' and 'Activity statement'


imohamed-rizwan:
I tried to print both detailed and normal activity statement for a vendor , but both reports ending balance is different.
Hereby am providing the screenshot for both the reports.
**Detailed report**
![detailed_activity](https://github.com/OCA/account-financial-reporting/assets/78721564/567851f2-44c7-4314-8a2b-bb5ca7defe92)

**Normal one**
![normal_activity](https://github.com/OCA/account-financial-reporting/assets/78721564/c6d50066-15c5-4cce-9095-500c0e13bb12)

It is correct for Detailed Activity Statement and showing wrong for normal ones.

**aging shows correct amount**

![aging](https://github.com/OCA/account-financial-reporting/assets/78721564/a8a3ea80-fc01-44ae-95bb-1ebbfa57036a)

imohamed-rizwan:
Created a PR reference to this
https://github.com/OCA/account-financial-reporting/pull/1104

PR #1104, opened 2024-01-29
[15.0][FIX] partner_statement Activity Statement End Balance mismatch

imohamed-rizwan:
Activity statement shows different End Balance amount comparing with Detailed Activity Statement

PR #1164, opened 2024-05-07
[15.0][FIX] partner_statement: don't consider reconciled lines outside date range

MiquelRForgeFlow:
Supersedes https://github.com/OCA/account-financial-reporting/pull/1104 and fixes https://github.com/OCA/account-financial-reporting/issues/1103.

If the reconciled sublines are outside the date range of the report, then mark them as red and don't consider them in the applied_amount of the above line (as they are already included in the prior balance).

Before this fix:
300367597-567851f2-44c7-4314-8a2b-bb5ca7defe92

After this fix, in the above image, the sublines (which are outside date range) should be in red color, and the applied amount of the Payment should be 0.

### DIFF
--- a/partner_statement/report/activity_statement_xlsx.py
+++ b/partner_statement/report/activity_statement_xlsx.py
@@ -310,7 +310,7 @@ class ActivityStatementXslx(models.AbstractModel):
                 FORMATS["current_money_format"] = workbook.add_format(
                     {"align": "right", "num_format": money_string}
                 )
-                bg_grey = "#CCCCCC"
+                bg_grey = "#ADB5BD"
                 FORMATS["format_tcell_left_blocked"] = copy_format(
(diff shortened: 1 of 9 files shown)


AaronHForgeFlow (approved):
Code + Functional :+1:

MiquelRForgeFlow:
With the fix:

 - Detailed:
 Selection_3154

- Activity:
Selection_3155

- Detailed xlsx:
Selection_3156

MiquelRForgeFlow:
@gdgellatly would you like to take a look? any opinion?

MiquelRForgeFlow:
@imohamed-rizwan @dannyadair @victoralmau @max3903 any opinions?

imohamed-rizwan:
Neat and clean, thanks brother @MiquelRForgeFlow  :1st_place_medal: :+1:

PR #1165, opened 2024-05-07
[14.0][IMP] account_financial_report: general ledger account range

WesleyOliveira98:
This improvement allows filtering accounts by range in the General Ledger for chart of accounts that use a dotted format, such as `1.1.1.01`. Currently, this functionality is not available because accounts with dots fail the `isdigit()` check.

Brazilian Chart of Accounts Example (https://github.com/OCA/l10n-brazil/blob/14.0/l10n_br_coa_simple/data/account.account.template.csv)
image

General Ledger with Brazilian Accounts
image

When attempting to remove dots from accounts and compare them as integers using the `filtered` method, the filter does not apply correctly if the accounts have different numbers of digits (when we use the `search` method, this isn't a problem). So, we implemented a solution where we search for the position of each account in the chart of accounts, finding the position of the `start_range` account and the `end_range` account, and then retrieve the accounts between these positions.

image

cc: @marcelsavegnago @kaynnan @Matthwhy

### DIFF
--- a/account_financial_report/tests/test_general_ledger.py
+++ b/account_financial_report/tests/test_general_ledger.py
@@ -742,3 +742,21 @@ class TestGeneralLedgerReport(AccountTestInvoicingCommon):
         wizard.account_type_ids = False
         wizard._onchange_account_type_ids()
         self.assertEqual(wizard.account_ids, account_model)
+
+    def test_validate_account_range(self):
+        accounts = self.env["account.account"].search([])
+        wizard = self.env["general.ledger.report.wizard"].create(
+            {
+                "company_id": False,
+                "account_code_from": accounts[0].id,
+                "account_code_to": accounts[9].id,
+            }
+        )
+        wizard.on_change_account_range()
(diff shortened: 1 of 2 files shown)


WesleyOliveira98:
It was necessary to search for the last appearance of the end_range code in case you have the same account code in two or more different companies, the index method returns the first appearance of the code, we also added some unit tests

kaynnan (approved):
LGTM

WesleyOliveira98:
ping @OCA/accounting-maintainers

marcelsavegnago:
ping @OCA/accounting-maintainers

marcelsavegnago (approved):

marcelsavegnago:
ping @WesleyOliveira98

WesleyOliveira98:
ping @OCA/accounting-maintainers

WesleyOliveira98:
@OCA/accounting-maintainers could someone reopen and review this pull request?

PR #1169, opened 2024-05-17
[16.0][FIX] account_financial_report: Multi-company support in test_report_without_aged_report_configuration

victoralmau:
Multi-company support in test_report_without_aged_report_configuration

If we installed `l10n_en` first and then `account_financial_report` the tests would fail because all accounting accounts were taken into account instead of only those of the corresponding company.

@Tecnativa


pedrobaeza (approved):
Fast-tracking this as obvious:

/ocabot merge patch

PR #1170, opened 2024-05-20
[16.0][IMP] account_financial_report: Change font size

pilarvargas-tecnativa:
Fw-port of #1148

Set a larger font style for the html report with a relative font size to make it responsive. For this, a new html_container is defined so that this does
not affect the other reports and the font style is customised only for the report in html view.

cc @Tecnativa TT48995

@pedrobaeza @victoralmau please review


chienandalu (approved):

pedrobaeza (approved):
Tested on runboat and working

/ocabot merge minor

PR #1177, opened 2024-05-30
[16.0][FIX] partner_statement: don't consider reconciled lines outside date range (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/account-financial-reporting/pull/1164.


AaronHForgeFlow (approved):
:+1:

PR #1181, opened 2024-05-31
[15.0][FIX] partner_statement: tests date relativity

chienandalu:
Depending on the day running the test, it will success or not.

@MiquelRForgeFlow I didn't go deep into the causes (I don't know the module), but it looks like the tests aren't resilient enough.

cc @Tecnativa

### DIFF
--- a/partner_statement/tests/test_activity_statement.py
+++ b/partner_statement/tests/test_activity_statement.py
@@ -3,11 +3,14 @@
 
 from datetime import date
 
+from freezegun import freeze_time
+
 from odoo import fields
 from odoo.tests import new_test_user
 from odoo.tests.common import TransactionCase
 
 
+@freeze_time("2024-05-01", tick=True)
 class TestActivityStatement(TransactionCase):
     """Tests for Activity Statement."""
 


partner_statement/tests/test_activity_statement.py
```diff
@@ -3,11 +3,14 @@
...
 from datetime import date
 
+from freezegun import freeze_time
+
 from odoo import fields
 from odoo.tests import new_test_user
 from odoo.tests.common import TransactionCase
 
 
+@freeze_time("2024-05-01", tick=True)
```
MiquelRForgeFlow, line 13: Thanks, but I would put this in `test_onchange_aging_type` below instead. The problem comes with that 30 days comparison.

This  problem comes from
Selection_3350

pedrobaeza: I don't see what you mean. Can you please propose the patch in other PR?

MiquelRForgeFlow: Just freezing the specific test.

Selection_3351

pedrobaeza (approved):
Merging this one as one valid approach, although the test is partial for only months with 30 days.

/ocabot merge patch

PR #1182, opened 2024-05-31
[15.0][FIX] partner_statement: fix test due to last change

MiquelRForgeFlow:
Supersedes https://github.com/OCA/account-financial-reporting/pull/1181.

In https://github.com/OCA/account-financial-reporting/pull/1164 was done this:
Selection_3350


pedrobaeza:
Non valid, so let's go with the other one.

PR #1202, opened 2024-07-15
[15.0][ADD] account_financial_report_sale: Open items grouped by delivery address

carolinafernandez-tecnativa:
- This module extends financial reports when needed to have dependency on sales. 
- Add the posibility on Open Items to group by partner delivery address.

@Tecnativa
TT50027
@pedrobaeza @sergio-teruel


account_financial_report_sale/report/open_items.py
```diff
@@ -0,0 +1,69 @@
...
+            grouped_by,
+            analytic_account_ids,
+            no_analytic,
+        )
+        if grouped_by == "partner_shipping":
+            for move_line in move_lines:
+                move = self.env["account.move"].browse(move_line["move_id"][0])
+                partner = move.partner_shipping_id
+                group_id = partner.id or 0
+                group_name = partner.name or _("Missing Delivery Address")
```
pedrobaeza: We need full name for children contacts (this doesn't happen on the partner on the journal items, as they are always the parent one):

```suggestion
                group_name = partner.display_name or _("Missing Delivery Address")
```

pedrobaeza (approved):

sergio-teruel (approved):

PR #1166, opened 2024-05-08
[14.0][IMP] account_financial_report: general ledger filter by account types

WesleyOliveira98:
This improvement reintroduces the functionality of filtering accounts by the `user_type_id` field, which was present in version 12 and was essential for filtering accounts in the general ledger report.

image

cc: @marcelsavegnago @kaynnan @Matthwhy

PR #1233, opened 2024-10-15
[17.0][FW] account_financial_report: multiple ports from 14.0

henrybackman:
Port of the following PRs from 14.0 to 17.0:
- #932
- #1154
- #1166


mmequignon (approved):
Not much context, but code looks good

pedrobaeza:
Well, the CI is red, so first, that should be fixed.

mmequignon:
> Well, the CI is red, so first, that should be fixed.

You're right, I missed this, `user_type_id` has been replaced by `account_type` (Many2one to Selection)

henrybackman:
> > Well, the CI is red, so first, that should be fixed.
> 
> You're right, I missed this, `user_type_id` has been replaced by `account_type` (Many2one to Selection)

Yes, this was the existing issue and it has now been resolved by changing the field in General Ledger Wizard to `account_type` as well. This however means that the wizard supports only one account type instead of many, but it was considered that expanding to multiple was beyond the scope of the forward porting.

henrybackman:
@WesleyOliveira98 While forward porting your changes, I made a change to the General Ledge Wizard to accommodate changes in 17.0. This has a functional impact:
- Before it was possible to select multiple account types but I implemented a simplified fix which allows selecting only a single account type 

Could I ask you to review the port and my changes?
Fix commit:
- [https://github.com/OCA/account-financial-reporting/pull/1233/commits/f980e17eb386906c228cad553d3eaa6c8ccee649](https://github.com/OCA/account-financial-reporting/pull/1233/commits/f980e17eb386906c228cad553d3eaa6c8ccee649)

WesleyOliveira98:
> * Before it was possible to select multiple account types but I implemented a simplified fix which allows selecting only a single account type

Hello @henrybackman, I don't think this is a good change, users sometimes select more than one account type at the same time

PR #1232, opened 2024-10-10
[16.0][FIX] account_financial_report: Display the values of Amount cur. + Cumul cur. values in general ledger only if necessary

victoralmau:
Display the values of _Amount cur_. + _Cumul cur._ values in general ledger only if necessary

The currency_id field of the lines is always set, so follow the same logic as in report/templates/general_ledger and show the data only if the currency is set in the account.

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT51204

PR #1235, opened 2024-10-16
[15.0][FIX] account_financial_report: Display the values of Amount cur. + Cumul cur. values in general ledger only if necessary

victoralmau:
Backport from 16.0: https://github.com/OCA/account-financial-reporting/pull/1232

Display the values of _Amount cur_. + _Cumul cur._ values in general ledger only if necessary

The currency_id field of the lines is always set, so follow the same logic as in report/templates/general_ledger and show the data only if the currency is set in the account.

@Tecnativa TT51204

### DIFF
--- a/account_financial_report/report/general_ledger_xlsx.py
+++ b/account_financial_report/report/general_ledger_xlsx.py
@@ -150,6 +150,7 @@ class GeneralLedgerXslx(models.AbstractModel):
         tags_data = res_data["tags_data"]
         filter_partner_ids = res_data["filter_partner_ids"]
         foreign_currency = res_data["foreign_currency"]
+        company_currency = res_data["company_currency"]
         # For each account
         for account in general_ledger:
             # Write account title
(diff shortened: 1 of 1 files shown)


pedrobaeza:
No subtotals and totals in currency?

imagen

victoralmau:
> No subtotals and totals in currency?
> 
Similar to v16, subtotals are only shown ((in both types of reports)) if the account has a currency set. https://github.com/OCA/account-financial-reporting/blob/16.0/account_financial_report/report/templates/general_ledger.xml#L705

pedrobaeza:
But I think that's not correct either: they should show the subtotals in currency. Having a currency in the account is different from showing the amount in currency.

victoralmau:
I think that can be raised in another PR. Also, I don't know exactly how it is supposed to be displayed, because if there are several lines with different currencies (EUR + GBP), should it all be added up in the total? In which currency would it be displayed?

pedrobaeza:
Yeah, indeed that belongs to other PR/issue. Anyhow, we have a similar problem right now with the accumulated balance in currency, as if we have several currencies, the accumulated amount has no sense mixing currencies, but I think it's a very corner case. By default, let's put the first currency if unique, and if not, let's put a float without currency.

pedrobaeza (approved):

carlosdauden (approved):

PR #1236, opened 2024-10-17
[15.0][IMP] account_financial_report: Add Amount cur. to General Ledger totals if the account has not set currency

victoralmau:
Add Amount cur. to General Ledger totals if the account has not set currency

Related to https://github.com/OCA/account-financial-reporting/pull/1235#issuecomment-2419269015

@Tecnativa

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -774,6 +774,7 @@ class GeneralLedgerReport(models.AbstractModel):
             list_centralized_ml += list(centralized_ml[jnl_id].values())
         return list_centralized_ml
 
+    # flake8: noqa: C901
     def _get_report_values(self, docids, data):
         wizard_id = data["wizard_id"]
         wizard = self.env["general.ledger.report.wizard"].browse(wizard_id)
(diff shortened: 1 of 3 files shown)


pedrobaeza:
The initial balance in currency is incorrect: 
image

pedrobaeza:
Still some weird amounts in http://oca-account-financial-reporting-15-0-pr1236-61a0d339a2d6.runboat.odoo-community.org

imagen

victoralmau:
Ok, I see what you mean, it shows -1224.01 instead of -601.74, but is it really an error? IMO no, I explain what happens and why.

1224.01=622.27 + 601.74

622.27 is something that although it is not shown on screen in the _Amount cur._ column (although in the _Cumul. Bal._ column)., it has been defined in https://github.com/OCA/account-financial-reporting/blob/15.0/account_financial_report/report/general_ledger.py#L150. 
The real problem (IMO) is that defining that value that way is “wrong” because we should only add the values of the lines if the currency is different from the company currency (but by performance it is not) or if there is a currency defined in the account. I think changing that value now would confuse some people. What do you think?

pedrobaeza:
Yeah, I understand, but it's incorrect. You have to only sum in the total if the currency is different from the company one.

victoralmau:
Changes done.

pedrobaeza (approved):
Tested and working correctly:

imagen

carlosdauden (approved):

PR #1242, opened 2024-10-24
[14.0][FIX] account_financial_report: Open Items report now sorted by account code

alexis-via:
BEFORE:

open_items_bug

AFTER:

open_items_fixed

Explaination of the bug:
When building the Qweb or XLSX report, the module uses open_items_move_lines_data.keys().
In python3, the order of a dict is the order of insertion of the keys. In the code of report/open_items.py, the keys of open_items_move_lines_data were inserted by the order of the first move line of the account.
This PR make sure that the keys in open_items_move_lines_data are inserted in the order of the account codes.

### DIFF
--- a/account_financial_report/report/open_items.py
+++ b/account_financial_report/report/open_items.py
@@ -206,11 +206,18 @@ class OpenItemsReport(models.AbstractModel):
 
     @api.model
     def _order_open_items_by_date(
-        self, open_items_move_lines_data, show_partner_details, partners_data
+        self,
+        open_items_move_lines_data,
+        show_partner_details,
+        partners_data,
+        accounts_data,
     ):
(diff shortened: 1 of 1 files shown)


alexis-via:
The bug is also present in v17, so I will forward port

open_items_bug_v17

nibraun (approved):
Works better

AaronHForgeFlow (approved):

pedrobaeza:
@eLBati are you overwriting this method in your localization? The signature is changing here.

alexis-via:
I grepped in OCA/l10n-italy for the method _order_open_items_by_date in branches 14, 15, 16 and 17, and it didn't return any result. So I think they don't inherit this method.

PR #1243, opened 2024-10-29
[16.0][IMP] account_financial_report: Add grouping by analytical account in Trial balance

victoralmau:
FWP from 15.0: https://github.com/OCA/account-financial-reporting/pull/1151

Changes done: 
- [x] Change t-esc and t-raw to t-out fromTrial Balance report
- [x] Refactor from Trial Balance
- [x] Add grouping by analytical account in Trial balance

ejemplo

@Tecnativa TT48969


pedrobaeza (approved):
Tested on runboat.

As there's now the possibility of having more than one analytic account in the analytic field, we should put a warning banner in the wizard form to specify that there can be duplicated amounts due to that.

carlosdauden (approved):

PR #1244, opened 2024-10-29
[16][FIX] account_financial_report : fix wizard model names sent to qweb reports for vat and aged balance reports

florian-dacosta:
The issue seems obvious, we get a wizard_id related to respectively `aged.partner.balance.report.wizard` and `vat.report.wizard` models but we browse it with another model (`open.items.report.wizard`).

As a result, if you try to read a field from the wizard in the qweb template, it will likely fail with an error like "The record is inexisting or has been deleted"

The problem does only appear in case of override of the qweb template and trying to read of field of the variable `o` which represent the  wizard.

I will forward port-it to 17.

A fast review is very welcome if you could as it is quite an easy one
@pedrobaeza @alexis-via @victoralmau @carlosdauden

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -442,8 +442,8 @@ class AgedPartnerBalanceReport(models.AbstractModel):
         )._calculate_percent(aged_partner_data)
         return {
             "doc_ids": [wizard_id],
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1252, opened 2024-11-20
[REF] mis_builder_cash_flow: Allow to override query

SirAionTech:
In another module, we have added a column to `mis.cash_flow.forecast_line` and would like to use it in the MIS reports.

The problem is that the MIS reports do not use directly `mis.cash_flow.forecast_line`: they use the view `mis.cash_flow` that exposes only some selected fields of `mis.cash_flow.forecast_line`.

With this change, it is possible for other modules to change the view generation before it is executed.
This allows us to add the new column of `mis.cash_flow.forecast_line` in the view `mis.cash_flow`.

Any suggestion is appreciated: I admit I don't know MIS builder modules that well, so there may be other solutions for this use-case.
Please let me know what you think.

### DIFF
--- a/mis_builder_cash_flow/report/mis_cash_flow.py
+++ b/mis_builder_cash_flow/report/mis_cash_flow.py
@@ -55,8 +55,8 @@ class MisCashFlow(models.Model):
             "selection"
         ]
 
-    def init(self):
-        query = """
+    def _init_query(self):
+        return """
             SELECT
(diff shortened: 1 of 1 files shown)


rvalyi:
cc @marcelsavegnago

rvalyi (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1254, opened 2024-11-27
[16.0][IMP] account_financial_report: Add Amount cur. to General Ledger totals if the account has not set currency

victoralmau:
FWP from 15.0: https://github.com/OCA/account-financial-reporting/pull/1236 + https://github.com/OCA/account-financial-reporting/pull/1256

Add Amount cur. to General Ledger totals if the account has not set currency

Related to https://github.com/OCA/account-financial-reporting/pull/1235#issuecomment-2419269015

Please @pedrobaeza and  @carlosdauden can you review it?

@Tecnativa

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -764,6 +764,7 @@ class GeneralLedgerReport(models.AbstractModel):
             list_centralized_ml += list(centralized_ml[jnl_id].values())
         return list_centralized_ml
 
+    # flake8: noqa: C901
     def _get_report_values(self, docids, data):
         wizard_id = data["wizard_id"]
         company = self.env["res.company"].browse(data["company_id"])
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Tested on runboat.

account_financial_report/report/general_ledger.py
```diff
@@ -838,6 +839,59 @@ def _get_report_values(self, docids, data):
...
                         del account["list_grouped"]
         general_ledger = sorted(general_ledger, key=lambda k: k["code"])
+        # Set the bal_curr of the initial balance to 0 if it does not correspond
+        # (reducing the corresponding of the bal_curr of the initial balance).
+        for gl_item in general_ledger:
+            if (
+                not gl_item["currency_id"]
+                or gl_item["currency_id"] != company.currency_id
+            ):
+                gl_item["fin_bal"]["bal_curr"] -= gl_item["init_bal"]["bal_curr"]
```
Tardo, line 851: This cause an error if the check "Show foreign currency" is not enabled because 'gl_item["init_bal"]["bal_curr"]' will never be defined.

Tardo (changes requested):

account_financial_report/report/general_ledger.py
victoralmau: Thank you, fixed in 15.0 (https://github.com/OCA/account-financial-reporting/pull/1256) and applied here.

Tardo: No! thank you for the good work! :)

Tardo (approved):

PR #1256, opened 2024-11-28
[15.0][FIX] account_financial_report: Avoid error in General ledger if the Show foreign currency checkbox is not checked.

victoralmau:
Avoid error in General ledger if the Show foreign currency checkbox is not checked.

Related to https://github.com/OCA/account-financial-reporting/pull/1236 and https://github.com/OCA/account-financial-reporting/pull/1254#discussion_r1860844349

@Tecnativa TT51996


pedrobaeza (approved):
Fast-tracking the bugfix:

/ocabot merge patch

PR #1260, opened 2024-12-03
[16.0][FIX] account_financial_report: KeyError rendering report

yajo:
Some reports were failing to render without this fix.

It seems there was a regression introduced in 0a2b2736cdc1a73d6248b91fa79a0f4c6aedc99b.

image

<details>

```
Traceback (most recent call last):
  File "<1582>", line 352, in template_1582
  File "<1582>", line 197, in template_1582_content
KeyError: 'fin_bal_currency_id'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/[http.py](https://http.py/)", line 1653, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/custom/src/odoo/odoo/service/[model.py](https://model.py/)", line 133, in retrying
    result = func()
  File "/opt/odoo/custom/src/odoo/odoo/[http.py](https://http.py/)", line 1680, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/custom/src/odoo/odoo/[http.py](https://http.py/)", line 1794, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/[ir_http.py](https://ir_http.py/)", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/custom/src/odoo/odoo/[http.py](https://http.py/)", line 734, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/auto/addons/report_xlsx/controllers/[main.py](https://main.py/)", line 49, in report_routes
    return super().report_routes(reportname, docids, converter, **data)
  File "/opt/odoo/custom/src/odoo/odoo/[http.py](https://http.py/)", line 734, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/auto/addons/web/controllers/[report.py](https://report.py/)", line 39, in report_routes
    html = report.with_context(context)._render_qweb_html(reportname, docids, data=data)[0]
  File "/opt/odoo/auto/addons/account_financial_report/models/[ir_actions_report.py](https://ir_actions_report.py/)", line 19, in _render_qweb_html
    return super(IrActionsReport, obj)._render_qweb_html(
[...]

PR #1261, opened 2024-12-04
[17.0][FIX] account_financial_report: KeyError rendering report

victoralmau:
FWP from 16.0: https://github.com/OCA/account-financial-reporting/pull/1260

Some reports were failing to render without this fix.

Example of use case:
- Define Account Currency (USD) to an account (101401 for example).
- Display the General ledger report

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #1265, opened 2025-01-02
[16.0][IMP] account_financial_report: Set grouped_by as required with non grouped item description. Fixes display of report when no agroupation is wanted

sergiobstoj:
This PR fixes two items.

In first place, the Grouped By field has two options that returns same result:
image

And with any of the two options, the result is not correct, as is not displaying all the data available for the accounts.

https://github.com/user-attachments/assets/6d943d4c-03ab-41b8-a9ee-71239ba487bc

With this PR we have the option limited to correct options:
image

And the result is displayed with all the info:

https://github.com/user-attachments/assets/1cb92ad8-933f-4150-b7e7-b70f4cd7b469

### DIFF
--- a/account_financial_report/wizard/general_ledger_wizard.py
+++ b/account_financial_report/wizard/general_ledger_wizard.py
@@ -74,8 +74,9 @@ class GeneralLedgerReportWizard(models.TransientModel):
         help="Ending account in a range",
     )
     grouped_by = fields.Selection(
-        selection=[("", "None"), ("partners", "Partners"), ("taxes", "Taxes")],
+        selection=[("none", "None"), ("partners", "Partners"), ("taxes", "Taxes")],
         default="partners",
+        required=True,
     )
     show_cost_center = fields.Boolean(
         string="Show Analytic Account",
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -180,8 +180,7 @@ class GeneralLedgerReport(models.AbstractModel):
             acc_id = gl["account_id"][0]
             data[acc_id] = self._prepare_gen_ld_data_item(gl)
             data[acc_id]["id"] = acc_id
-            if grouped_by:
-                data[acc_id][grouped_by] = False
+            data[acc_id][grouped_by] = False
         method = "_prepare_gen_ld_data_group_%s" % grouped_by
         if not hasattr(self, method):
             return data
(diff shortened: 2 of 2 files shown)


aritzolea (approved):

aritzolea:
@pedrobaeza Could you please review this PR?

account_financial_report/wizard/general_ledger_wizard.py
```diff
@@ -74,8 +74,9 @@ class GeneralLedgerReportWizard(models.TransientModel):
         help="Ending account in a range",
     )
     grouped_by = fields.Selection(
-        selection=[("", "None"), ("partners", "Partners"), ("taxes", "Taxes")],
+        selection=[("none", "None"), ("partners", "Partners"), ("taxes", "Taxes")],
```
pedrobaeza, line 77: I think changing this value will provoke side effects when checking for falsy values across the code. Have you checked it?

LauraCForgeFlow (approved):
Code + functional LGTM! :+1:

account_financial_report/wizard/general_ledger_wizard.py
dreispt: AFAICT a `"none"` string will nt be a falsy value, right?

dreispt:
```
>>> bool("none")
True
```

dreispt: I see your point, ignore my comment:
the problem is that this field will be always set (as it is required) and then always "truthy" even if set to "none".

pedrobaeza: That's why I don't feel comfortable changing this on this old stable version. Other dependent modules may fail because they check `not record.grouped_by`, and since this, the logic will change.

pedrobaeza (approved):
After a customer report, I have been checking the problem, and having an empty value as selection key prevents to put such value in the user defaults, and indeed there's an incorrect behavior leaving the dropdown empty and filtering any account, having no result in that case.

I haven't seen any OCA module depending on this one affected by the possible side effect, so let's merge for fixing both behaviors, and forwarding to upper versions at the same time, as the reported version is 17, but affecting the same.

Thanks for the PR, @sergiobstoj, although next time it would be interesting for a better comprehension to express the steps to reproduce the problem clear and concisely.

/ocabot merge patch

PR #1271, opened 2025-01-27
[FIX] account_financial_report: general ledger: sort by list_grouped name

mourad-ehm:
Actually the general ledger is some times not sorted by partner name when we grouped by partner (it seems random).
With this PR we force the sort by list_grouped name.


mourad-ehm:
@bealdav

mourad-ehm:
ping @alexis-via

PR #1281, opened 2025-02-05
[15.0][IMP] account_financial_report: Allow user cut label string of account move line in reports

sergio-teruel:
@Tecnativa 
Better inheritance report methods

With this change the user can:

- shorten the column label shown in the reports and in the html view, zero value not cut
- Hide any defined column

This setting option are stored to be used in next time wizard launch

image

Report generated before this PR:

image

Report generated after this PR:
image


sergio-teruel:
ping @carlosdauden

sergio-teruel:
@victoralmau @pilarvargas-tecnativa Can you review??

pilarvargas-tecnativa (approved):
Just a minor suggestion, could you add to the readme how to configure the columns in the corresponding tab of the report?

account_financial_report/models/account_financial_report_column.py
```diff
@@ -0,0 +1,18 @@
...
+    _order = "sequence, id"
+
+    res_model = fields.Char()
+    sequence = fields.Integer()
+    name = fields.Char(required=True, translate=True)
+    expression_label = fields.Char(required=True)
+    is_visible = fields.Boolean(string="Show", default=True)
+    field_type = fields.Char()
+    limit = fields.Integer()
+    # user_id = fields.Many2one(comodel_name="res.users")
```
victoralmau: Delete this commented field if it is not used.

account_financial_report/wizard/journal_ledger_wizard_view.xml
```diff
@@ -7,38 +7,56 @@
...
+                                </group>
+                                <group />
+                            </group>
+                            <separator string="Journals" />
+                            <group>
+                                <field name="journal_ids" widget="many2many_tags" />
+                            </group>
+                        </page>
+                        <page string="Columns">
+                            <group name="columns">
```
victoralmau:
```suggestion
                            <group name="columns">
```
Remove the group for better visibility (full width). This applies to all wizards (except general ledger).

victoralmau (changes requested):
Code review OK.

account_financial_report/models/account_financial_report_column.py
carlosdauden: Deleted

account_financial_report/wizard/journal_ledger_wizard_view.xml
carlosdauden: Removed

victoralmau (approved):

victoralmau:
Sorry, can you add the comments https://github.com/OCA/account-financial-reporting/pull/1281#pullrequestreview-2617785847 in the readme?

carlosdauden:
@pilarvargas-tecnativa comments added to readme

pedrobaeza:
Excellent new feature, thanks. Isn't the "Show Foreign Currency", "Show Analytic Tags", etc in conflict with this new option and should be simplified?

carlosdauden:
Is comented in the roadmap:
https://github.com/OCA/account-financial-reporting/pull/1281/files#diff-b3a543c24a906f18f5772cd72ae22cafd9bc8f4fdb60fcde318e2f9164deab44R101-R102

They do not conflict too much, because by default all the options on the columns page are activated and for them to be displayed both options must be ticked, so for practical purposes a user who is not aware of the new functionality continues to operate as usual without any problems.

pedrobaeza:
OK, please fw-port the improvement and this can be done in 17.

/ocabot merge minor

Saran440:
@carlosdauden 

Hi, Does this feature support only the HTML view?
I tested with PDF and Excel, and the column still appears.
I think this feature should also support Excel and PDF.
What do you think?

pedrobaeza:
The PDF export is fed from the same place as HTML, so that should work for both. Excel seems a different beast and indeed should be synced. I'm doing the corresponding fw-ports and later will check that, but any help here is appreciated.

PR #1284, opened 2025-02-11
[MIG] partner_statement: Migration to 18.0

kobros-tech:
this version is more efficient than #1285


partner_statement/readme/CONFIGURE.md
dannyadair, line 1: > In *Application / Accounting & Finance*, select *Billing* or *Billing Manager*

I believe this may be inaccurate / different between Community and Enterprise. Check the boxes for "Use activity statements" and "Use outstanding statements" or pick an accounting role that includes them ("Invoicing").

partner_statement/README.rst
dannyadair, line 1: > In Application / Accounting & Finance, select Billing or Billing Manager

I believe this may be inaccurate / different between Community and Enterprise. Check the boxes for "Use activity statements" and "Use outstanding statements" or pick an accounting role that includes them ("Invoicing").

partner_statement/readme/DESCRIPTION.md
dannyadair, line 1: > There are three types of statements: Activity, Detailed Activity, and Outstanding.

This may be outdated - there's just Activity and Outstanding?

dannyadair (approved):
Source LGTM, some documentation maybe outdated.
Took it for a spin and appears to be working as expected.

dannyadair:
I'd like to note that v18 Enterprise has `account_reports` which includes a "Customer Statement" web view with PDF/XLSX export (see screenshots). Not as many features, maybe this should be split with an Enterprise-compatible version adding features? As it stands I think a user can get confused with multiple ways of obtaining a statement.
Screenshot_20250320_125524
Screenshot_20250320_125543
Screenshot_20250320_125610

partner_statement/report/detailed_activity_statement_xlsx.py
```diff
@@ -98,24 +98,23 @@ def _write_currency_lines(self, row_pos, sheet, partner, currency, data):
...
                 format_tcell_left = FORMATS["format_tcell_left_reconciled"]
                 format_tcell_date_left = FORMATS["format_tcell_date_left_reconciled"]
                 format_distributed = FORMATS["format_distributed_reconciled"]
                 current_money_format = FORMATS["current_money_format_reconciled"]
             elif (
-                line.get("blocked")
-                and line.get("reconciled_line")
-                and not line.get("outside-date-rank")
+                # line.get("blocked")
+                line.get("reconciled_line") and not line.get("outside-date-rank")
```
MiquelRForgeFlow: why not directly delete this `elif`?

partner_statement/report/report_statement_common.py
```diff
@@ -445,8 +445,8 @@ def _get_report_values(self, docids, data=None):
                         line["currency_id"], currencies, 0.0, 0.0
                     )
                 line_currency = currency_dict[line["currency_id"]]
-                if not line["blocked"]:
-                    line_currency["amount_due"] += line["open_amount"]
+                # if not line["blocked"]:
+                #     line_currency["amount_due"] += line["open_amount"]
```
MiquelRForgeFlow:
```suggestion
                line_currency["amount_due"] += line["open_amount"]
```

MiquelRForgeFlow (commented):
I am thinking know if in the `account` migration (from OpenUpgrade), we should delete the lines with "blocked". If not, after migration, this report will treat those blocked lines as not blocked.

Comments done for `[MIG] partner_statement: Migration to 18.0` commit:

MiquelRForgeFlow:
> Not as many features, maybe this should be split with an Enterprise-compatible version adding features? As it stands I think a user can get confused with multiple ways of obtaining a statement.

@dannyadair The partner statements of this module can be hidden by not enabling them in the config settings. People should have to choose between using the enterprise ones or the OCA ones. I don't see any reason why someone would want both.

MiquelRForgeFlow:
Selection_4791

Please, squash bot commits. If not, it's difficult to comment the affected lines.

dannyadair:
> > Not as many features, maybe this should be split with an Enterprise-compatible version adding features? As it stands I think a user can get confused with multiple ways of obtaining a statement.
> 
> @dannyadair The partner statements of this module can be hidden by not enabling them in the config settings. People should have to choose between using the enterprise ones or the OCA ones. I don't see any reason why someone would want both.

@MiquelRForgeFlow Sorry I wasn't clear. I wouldn't install and then disable (just not install), my concern is the other way around. I can't see any setting for disabling the Enterprise statements (and account_reports includes more than that). We can adjust groups to hide menus/buttons but if the OCA statements are meant to be a replacement, the module should do that imho.

MiquelRForgeFlow:
> We can adjust groups to hide menus/buttons but if the OCA statements are meant to be a replacement, the module should do that imho.

In fact, we should not care about Enterprise. OCA modules in general don't care, it happens often the functionality duplication. For example, the `hr_timesheet_sheet` module. I would say that If you happen to have both Enterprise and this module, then it's your custom problem how to handle it.

dannyadair:
> In fact, we should not care about Enterprise. OCA modules in general don't care, it happens often the functionality duplication. For example, the `hr_timesheet_sheet` module. I would say that If you happen to have both Enterprise and this module, then it's your custom problem how to handle it.

I see. Sorry I wasn't aware. That makes sense. Thank you.

AaronHForgeFlow:
Conflicts, and superseded here https://github.com/OCA/account-financial-reporting/pull/1364

PR #1285, opened 2025-02-11
[MIG] partner_statement: Migration to 18.0

kobros-tech:
this version is more careful than #1284


gdgellatly:
Closing this in favor of #1284 . Adding in the blocked feature into a report is poor practice and not with this modules rationale. If someone wanted to add it back, then it would be a separate module. It was a crappy feature anyway, glad its gone.

PR #1287, opened 2025-02-13
[16.0][PERF] Improve journal ledger performance with lots of lines

hugosantosred:
Improves the performance when obtaining the Journal Ledger with lots of account moves and lines.

For example in a demo database with 475K customer invoices before this changes the report in xlsx format is generated in 1900 seconds. With this changes is generated in 120 seconds

### DIFF
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -3,6 +3,7 @@
 
 import itertools
 import operator
+from collections import defaultdict
 
 from odoo import models
 
@@ -181,13 +182,23 @@ class JournalLedgerReport(models.AbstractModel):
         return {"move_line_ids": tuple(move_lines.ids)}
 
     def _get_move_lines(self, move_ids, wizard, journal_ids):
-        move_lines = self.env["account.move.line"].search(
-            self._get_move_lines_domain(move_ids, wizard, journal_ids),
-            order=self._get_move_lines_order(move_ids, wizard, journal_ids),
+        move_lines = (
+            self.env["account.move.line"]
+            .with_context(prefetch_fields=False)
+            .search(
+                self._get_move_lines_domain(move_ids, wizard, journal_ids),
+                order=self._get_move_lines_order(move_ids, wizard, journal_ids),
+            )
         )
-        move_lines_exigible = self.env["account.move.line"].search(
-            self._get_move_lines_domain(move_ids, wizard, journal_ids)
-            + self.env["account.move.line"]._get_tax_exigible_domain(),
+        # Get the exigible move lines ids instead of the recordset to increase
+        # performance with a large number of journal items
+        move_lines_exigible_ids = set(
+            self.env["account.move.line"]
+            .search(
+                self._get_move_lines_domain(move_ids, wizard, journal_ids)
+                + self.env["account.move.line"]._get_tax_exigible_domain(),
+            )
+            .ids
         )
         move_line_ids_taxes_data = {}
         if move_lines:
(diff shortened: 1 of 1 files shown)


aritzolea (approved):
Code review. Thanks a lot for this huge performance improvement!

liebana (approved):
Functional review, LGTM.

account_financial_report/report/journal_ledger.py
```diff
@@ -100,9 +101,9 @@ def _get_move_lines_data(self, ml, wizard, ml_taxes, auto_sequence, exigible):
             base_debit = ml_taxes and ml.debit or 0.0
             base_credit = ml_taxes and ml.credit or 0.0
             base_balance = ml_taxes and ml.balance or 0.0
-            tax_debit = ml.tax_line_id and ml.debit or 0.0
-            tax_credit = ml.tax_line_id and ml.credit or 0.0
-            tax_balance = ml.tax_line_id and ml.balance or 0.0
+            tax_debit = ml.tax_line_id.id and ml.debit or 0.0
```
pedrobaeza: Why is this better?

account_financial_report/report/journal_ledger.py
```diff
@@ -209,36 +218,34 @@ def _get_move_lines(self, move_ids, wizard, journal_ids):
...
+            exigible = ml.id in move_lines_exigible_ids
+            Move_Lines[move_id].append(
                 self._get_move_lines_data(ml, wizard, taxes, auto_sequence, exigible)
             )
-        account_ids_data = self._get_account_data(accounts)
-        partner_ids_data = self._get_partner_data(partners)
-        currency_ids_data = self._get_currency_data(currencies)
-        tax_line_ids_data = self._get_tax_line_data(tax_lines)
+        account_ids_data = self._get_account_data(
+            move_lines.mapped("account_id").with_context(
```
pedrobaeza: No need of the `mapped` (just syntactic sugar for better readability):

```suggestion
            move_lines.account_id.with_context(
```

account_financial_report/report/journal_ledger.py
```diff
@@ -209,36 +218,34 @@ def _get_move_lines(self, move_ids, wizard, journal_ids):
...
+                prefetch_fields=["name", "code", "account_type"]
```
pedrobaeza: Not sure the gain on this part.

hugosantosred: Tested it with no appreciable differences, so I removed it for clarity

account_financial_report/report/journal_ledger.py
hugosantosred: Changed

account_financial_report/report/journal_ledger.py
hugosantosred: Tested and its the same in terms of performance, so I rolled back these changes to keep them as they were before.

account_financial_report/report/journal_ledger.py
```diff
@@ -209,36 +218,22 @@ def _get_move_lines(self, move_ids, wizard, journal_ids):
                     "description": tax_description,
                 }
         Move_Lines = {}
-        accounts = self.env["account.account"]
-        partners = self.env["res.partner"]
-        currencies = self.env["res.currency"]
-        tax_lines = self.env["account.tax"]
         auto_sequence = len(move_ids)
+        Move_Lines = defaultdict(list)
+
```
pedrobaeza: Avoid empty lines inside the method.

pedrobaeza (commented):
Better this way, thanks. The only missing thing is to comment in code before all the `.id` stuff that this technique is used for avoiding continuous recordset recreation, which affects dramatically the performance on systems with huge amount of journal items. This way, you avoid that in the future someone is tempted to restore the recordset thing.

account_financial_report/report/journal_ledger.py
hugosantosred: Done, thanks!

hugosantosred:
> Better this way, thanks. The only missing thing is to comment in code before all the `.id` stuff that this technique is used for avoiding continuous recordset recreation, which affects dramatically the performance on systems with huge amount of journal items. This way, you avoid that in the future someone is tempted to restore the recordset thing.

Added the comment in the exigibility check

hugosantosred:
@pedrobaeza Could you have a look if it is already correct with the latest changes?

Thank you!

pedrobaeza (approved):
Please fw-port it to upper versions:

/ocabot merge patch

PR #1291, opened 2025-02-20
[17.0][IMP] account_financial_report: Allow user cut label string of account move line in reports

carlosdauden:
fw-port from https://github.com/OCA/account-financial-reporting/pull/1281

@Tecnativa


pedrobaeza (approved):
Updated extracting the important parts, and improving compatibility with other possible extension modules.

PR #1292, opened 2025-02-24
[16.0][FIX] account_tax_balance: Fallback to load CoA

pedrobaeza:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with error:

```
odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale
```

provoked by the lack of a CoA installed, so we put little code to select the first available CoA if `l10n_generic_coa` is not installed.

@Tecnativa


HaraldPanten (approved):

pedrobaeza:
Tested on integration tests, so merging

/ocabot merge patch

PR #1293, opened 2025-02-26
[18.0][FIX] account_financial_report: replace old field company_id in account.account by company_ids

bguillot:
In v18 account.account field company_id as been replaced by company_ids

### DIFF
--- a/account_financial_report/wizard/general_ledger_wizard.py
+++ b/account_financial_report/wizard/general_ledger_wizard.py
@@ -108,7 +108,7 @@ class GeneralLedgerReportWizard(models.TransientModel):
             )
             if self.company_id:
                 self.account_ids = self.account_ids.filtered(
-                    lambda a: a.company_id == self.company_id
+                    lambda a: self.company_id in a.company_ids
                 )
 
     def _init_date_from(self):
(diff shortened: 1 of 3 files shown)


account_financial_report/wizard/general_ledger_wizard.py
```diff
@@ -108,7 +108,7 @@ def on_change_account_range(self):
             )
             if self.company_id:
                 self.account_ids = self.account_ids.filtered(
-                    lambda a: a.company_id == self.company_id
+                    lambda a: self.company_id in a.company_ids
```
florian-dacosta: I think we could include company_ids in the search (with a if to increment the domain) and this way get rid of the filtered.
But it is more a refactore than a fix, and I believe we should do a large refactore of this module to be able to get a report for multiple companies (but necessarly all). It could be done at this time.

florian-dacosta (approved):
LGTM code review

TB-Ph35 (approved):

alexis-via:
fixes https://github.com/OCA/account-financial-reporting/issues/1335

alexis-via (approved):

PR #1297, opened 2025-03-08
[17.0][FIX] account_financial_report: fix analytic_distribution var name

ramiadavid:
The variable analytic_ids is duplicated and an attempt is being made to overwrite the values ​​inside the for


pedrobaeza (approved):
Clear bugfix

/ocabot merge patch

PR #1300, opened 2025-03-24
[16.0][IMP] partner_statement: Allow installation of partner_statement on 16.0 - Issue 1299

PhillipJVenter:
When attempting to install the partner_statement module, the installation fails with a KeyError: 'active_ids'. The error occurs when the module tries to access "active_ids" from the context, but it is missing.

### DIFF
--- a/partner_statement/wizard/statement_common.py
+++ b/partner_statement/wizard/statement_common.py
@@ -21,7 +21,7 @@ class StatementCommon(models.AbstractModel):
     date_end = fields.Date(required=True, default=fields.Date.context_today)
     show_aging_buckets = fields.Boolean(default=True)
     number_partner_ids = fields.Integer(
-        default=lambda self: len(self._context["active_ids"])
+        default=lambda self: self._context.get("active_ids") and len(self._context.get("active_ids"))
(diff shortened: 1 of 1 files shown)


partner_statement/wizard/statement_common.py
```diff
@@ -21,7 +21,7 @@ class StatementCommon(models.AbstractModel):
     date_end = fields.Date(required=True, default=fields.Date.context_today)
     show_aging_buckets = fields.Boolean(default=True)
     number_partner_ids = fields.Integer(
-        default=lambda self: len(self._context["active_ids"])
+        default=lambda self: self._context.get("active_ids") and len(self._context.get("active_ids"))
```
MiquelRForgeFlow:
```suggestion
        default=lambda self: len(self._context.get("active_ids", []))
```

MiquelRForgeFlow:
@PhillipHumint Please, squash both commits.

BTW, I think the issue happened due to an upstream odoo commit, and after that much time, maybe they have already fixed it with another commit. Could you check if you still need this fix using the last odoo commits?

PR #1303, opened 2025-04-01
[17.0][FIX] account_financial_report: Fix error with multiple analytic accounts in distribution

BernatObrador:
Fixed an issue where an error was raised when distributions contained more than one analytic account. The error occurred because the system attempted to convert a comma-separated string (e.g., '516, 986') into an integer.

Error:
image

cc https://github.com/APSL HT01808

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review

### DIFF
--- a/account_financial_report/report/general_ledger_xlsx.py
+++ b/account_financial_report/report/general_ledger_xlsx.py
@@ -195,16 +195,18 @@ class GeneralLedgerXslx(models.AbstractModel):
                             taxes_description += taxes_data[tax_id]["tax_name"] + " "
                         if line["tax_line_id"]:
                             taxes_description += line["tax_line_id"][1]
-                        for account_id, value in line["analytic_distribution"].items():
-                            if value < 100:
-                                analytic_distribution += "%s %d%% " % (
-                                    analytic_data[int(account_id)]["name"],
-                                    value,
-                                )
-                            else:
-                                analytic_distribution += (
-                                    "%s " % analytic_data[int(account_id)]["name"]
-                                )
+                        analytic_list = []
+                        for account_ids, percentage in line[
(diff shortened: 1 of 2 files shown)


mpascuall (approved):
LGTM!

pedrobaeza (changes requested):
There's no need of such complex patch for something that you only need to touch slightly the QWeb, and now you are showing always 100%, while it shouldn't.

BernatObrador:
Hi @pedrobaeza I think it's needed to show all the % for every account in the distribution, because you can have more than 1 account with different percentages Ex:

Distribution: {"1": 25, "4":25, "10": 50}.

Will be shown as:

First_account 25%
Second_account 25%
Last_account 50%

You cannot show only 100% because it can be divided by more than one account.

What do you think about it?

pedrobaeza:
I'm saying the contrary: don't show 100%, as it's the usual thing, as it was before, and it's wasting space.

BernatObrador:
@pedrobaeza I was not understanding your first comment.

Right now it doesn't show the 100% ones.

pedrobaeza (commented):
OK, right, I read bad the code. Anyway, as said, the patch is too large when you only need to touch the QWeb report for displaying the text field. No iteration nor change of structure.

BernatObrador:
The iteration is necessary because some keys in the distribution dictionary contain multiple analytic IDs in a single string, such as {"1,2": 50, "3,4": 50}.
To handle these cases, I need to split the key and iterate over each individual ID.
I’m not sure how to process them otherwise without introducing a second loop.

pedrobaeza (approved):
/ocabot merge patch

PR #1313, opened 2025-05-06
[IMP] partner_statement: Allow to exclude accounts

SirPyTech:
Some journal items need to be excluded from the outstanding report because they are payments that are late not because of the partner's fault.
All these kind of journal items can be identified by their account.
This module allows to exclude specific accounts (with selectors like 70%) from the outstanding statements.

### DIFF
--- a/partner_statement/report/activity_statement.py
+++ b/partner_statement/report/activity_statement.py
@@ -16,6 +16,9 @@ class ActivityStatement(models.AbstractModel):
     _description = "Partner Activity Statement"
 
     def _initial_balance_sql_q1(self, partners, date_start, account_type):
+        excluded_accounts_ids = tuple(
+            self.env.context.get("excluded_accounts_ids", [])
+        ) or (-1,)
         return str(
             self._cr.mogrify(
                 """
@@ -46,6 +49,7 @@ class ActivityStatement(models.AbstractModel):
             ) as pc ON pc.credit_move_id = l.id
             WHERE l.partner_id IN %(partners)s
                 AND at.type = %(account_type)s
+                AND aa.id not in %(excluded_accounts_ids)s
                 AND l.date < %(date_start)s AND not l.blocked
                 AND m.state IN ('posted')
                 AND (
@@ -118,6 +122,9 @@ class ActivityStatement(models.AbstractModel):
     def _display_activity_lines_sql_q1(
         self, partners, date_start, date_end, account_type
     ):
+        excluded_accounts_ids = tuple(
+            self.env.context.get("excluded_accounts_ids", [])
+        ) or (-1,)
         payment_ref = _("Payment")
         return str(
             self._cr.mogrify(
@@ -155,6 +162,7 @@ class ActivityStatement(models.AbstractModel):
             JOIN account_journal aj ON (l.journal_id = aj.id)
             WHERE l.partner_id IN %(partners)s
                 AND at.type = %(account_type)s
+                AND aa.id not in %(excluded_accounts_ids)s
                 AND %(date_start)s <= l.date
                 AND l.date <= %(date_end)s
                 AND m.state IN ('posted')
--- a/partner_statement/report/outstanding_statement.py
+++ b/partner_statement/report/outstanding_statement.py
@@ -14,6 +14,9 @@ class OutstandingStatement(models.AbstractModel):
 
     def _display_outstanding_lines_sql_q1(self, partners, date_end, account_type):
         partners = tuple(partners)
+        excluded_accounts_ids = tuple(
+            self.env.context.get("excluded_accounts_ids", [])
+        ) or (-1,)
         return str(
             self._cr.mogrify(
                 """
@@ -60,6 +63,7 @@ class OutstandingStatement(models.AbstractModel):
                 WHERE l2.date <= %(date_end)s
             ) as pc ON pc.credit_move_id = l.id
             WHERE l.partner_id IN %(partners)s AND at.type = %(account_type)s
+                AND aa.id not in %(excluded_accounts_ids)s
                 AND (
                     (pd.id IS NOT NULL AND
                         pd.max_date <= %(date_end)s) OR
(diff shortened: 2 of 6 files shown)


partner_statement/report/activity_statement.py
```diff
@@ -16,6 +16,9 @@ class ActivityStatement(models.AbstractModel):
     _description = "Partner Activity Statement"
 
     def _initial_balance_sql_q1(self, partners, date_start, account_type):
+        excluded_accounts_ids = tuple(
+            self.env.context.get("excluded_accounts_ids", [])
+        ) or (-1,)
```
MiquelRForgeFlow, line 21:
```suggestion
        )
```
-1 is treated like 1

partner_statement/report/activity_statement.py
```diff
@@ -46,6 +49,7 @@ def _initial_balance_sql_q1(self, partners, date_start, account_type):
             ) as pc ON pc.credit_move_id = l.id
             WHERE l.partner_id IN %(partners)s
                 AND at.type = %(account_type)s
+                AND aa.id not in %(excluded_accounts_ids)s
```
MiquelRForgeFlow, line 52:
```suggestion
                %(excluded_accounts)s
```

partner_statement/report/activity_statement.py
MiquelRForgeFlow, line 21:
```suggestion
        excluded_accounts_ids = tuple(
            self.env.context.get("excluded_accounts_ids", [])
        )
        excluded_accounts = ("AND aa.id not in %s" % excluded_accounts_ids) if excluded_accounts_ids else ""
```

SirPyTech: Thanks for all the suggestions, I should be able to work on this maybe next week.
In the meanwhile, could you please clarify why 
> -1 is treated like 1

?
I thought this _(-1, ) trick_ was working well

MiquelRForgeFlow (approved):
code LGTM

partner_statement/report/activity_statement.py
SirPyTech: Hi,
I'm sorry but I had to revert this change because it is actually not working as expected: it was injecting in the query a string
<details>
<summary>Query</summary>
<pre>
WITH Q1 as (
    SELECT l.id, m.name AS move_id, l.partner_id, l.date, l.name,
        l.blocked, l.currency_id, l.company_id,
    CASE WHEN l.ref IS NOT NULL
        THEN l.ref
        ELSE m.ref
    END as ref,
    CASE WHEN (l.currency_id is not null AND l.amount_currency > 0.0)
        THEN avg(l.amount_currency)
        ELSE avg(l.debit)
    END as debit,
    CASE WHEN (l.currency_id is not null AND l.amount_currency < 0.0)
        THEN avg(l.amount_currency * (-1))
        ELSE avg(l.credit)
    END as credit,
    CASE WHEN l.balance > 0.0
        THEN l.balance - sum(coalesce(pd.amount, 0.0))
        ELSE l.balance + sum(coalesce(pc.amount, 0.0))
    END AS open_amount,
    CASE WHEN l.balance > 0.0
        THEN l.amount_currency - sum(coalesce(pd.debit_amount_currency, 0.0))
        ELSE l.amount_currency + sum(coalesce(pc.credit_amount_currency, 0.0))
    END AS open_amount_currency,
    CASE WHEN l.date_maturity is null
        THEN l.date
        ELSE l.date_maturity
    END as date_maturity
    FROM account_move_line l
    JOIN account_account aa ON (aa.id = l.account_id)
    JOIN account_account_type at ON (at.id = aa.user_type_id)
    JOIN account_move m ON (l.move_id = m.id)
    LEFT JOIN (SELECT pr.*
        FROM account_partial_reconcile pr
        INNER JOIN account_move_line l2
        ON pr.credit_move_id = l2.id
        WHERE l2.date <= '2025-05-23'::date
    ) as pd ON pd.debit_move_id = l.id
    LEFT JOIN (SELECT pr.*
        FROM account_partial_reconcile pr
        INNER JOIN account_move_line l2
        ON pr.debit_move_id = l2.id
        WHERE l2.date <= '2025-05-23'::date
    ) as pc ON pc.credit_move_id = l.id
    WHERE l.partner_id IN (10) AND at.type = 'receivable'
        'AND aa.id not in (6,)'  // <------------------------------------------ This is the injected string
        AND (
            (pd.id IS NOT NULL AND
                pd.max_date <= '2025-05-23'::date) OR
            (pc.id IS NOT NULL AND
                pc.max_date <= '2025-05-23'::date) OR
            (pd.id IS NULL AND pc.id IS NULL)
        ) AND l.date <= '2025-05-23'::date AND m.state IN ('posted')
    GROUP BY l.id, l.partner_id, m.name, l.date, l.date_maturity, l.name,
        CASE WHEN l.ref IS NOT NULL
            THEN l.ref
            ELSE m.ref
        END,
        l.blocked, l.currency_id, l.balance, l.amount_currency, l.company_id
    ),
     Q2 AS (
        SELECT Q1.partner_id, Q1.currency_id, Q1.move_id,
            Q1.date, Q1.date_maturity, Q1.debit, Q1.credit,
            Q1.name, Q1.ref, Q1.blocked, Q1.company_id,
            CASE WHEN Q1.currency_id is not null
                THEN Q1.open_amount_currency
                ELSE Q1.open_amount
            END as open_amount, Q1.id
        FROM Q1
        ),
     Q3 AS (
    SELECT Q2.partner_id, Q2.move_id, Q2.date,
        Q2.date_maturity, Q2.name, Q2.ref, Q2.debit,
        Q2.credit, Q2.debit-Q2.credit AS amount,
        COALESCE(Q2.currency_id, c.currency_id) AS currency_id,
        Q2.open_amount, Q2.blocked, Q2.id
    FROM Q2
    JOIN res_company c ON (c.id = Q2.company_id)
    WHERE c.id = 1 AND Q2.open_amount != 0.0
    )
SELECT partner_id, currency_id, move_id, date, date_maturity, debit,
    credit, amount, open_amount, COALESCE(name, '') as name,
    COALESCE(ref, '') as ref, blocked, id
FROM Q3
ORDER BY date, date_maturity, move_id 
</pre>
</details>

And this caused the query to return nothing when any account had to be excluded.

Too bad that both me and the unit tests excluded everything and checked that the report returned nothing!
I also improved the unit test to check that something is actually in the report when only some accounts are excluded.

I've done some research and apparently `mogrify` is supposed to work with parameters substitution, so injecting SQL code causes this behavior.
Let me know if you have any other suggestion.

PR #1315, opened 2025-05-06
[15.0][FIX] account_financial_report: Amount accumulated several times using hierarchy

carlosdauden:
When the group structure does not have all possible siblings defined, the amounts are accumulated several times to the ancestors.

With this order we make sure that the lower level groups are processed before their children to avoid that when the parent is processed the balance has been updated by the child.

@Tecnativa TT56255

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -5,6 +5,7 @@
 
 
 from odoo import _, api, models
+from odoo.exceptions import UserError
 from odoo.tools.float_utils import float_is_zero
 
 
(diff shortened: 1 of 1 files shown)


account_financial_report/report/trial_balance.py
```diff
@@ -653,8 +654,26 @@ def _get_data_grouped(self, total_amount, accounts_data, foreign_currency):
...
-        for group_id in group_ids:
+        processed_groups = []
+        # Sort groups so that parent groups are processed before child groups
+        groups = (
+            self.env["account.group"]
+            .browse(group_ids)
+            .sorted(key=lambda x: x.complete_code)
+        )
+        for group_id in groups:
+            group_id = group_id.id
```
pedrobaeza:
```suggestion
        for group in groups:
            group_id = group.id
```

account_financial_report/report/trial_balance.py
```diff
@@ -653,8 +654,26 @@ def _get_data_grouped(self, total_amount, accounts_data, foreign_currency):
...
             parent_id = groups_data[group_id]["parent_id"]
+            if group_id in processed_groups:
+                raise UserError(
+                    _(
+                        "There is a problem in the structure of account groups. "
```
pedrobaeza:
```suggestion
                        "There is a problem in the structure of the account groups. "
```

pedrobaeza (approved):

pedrobaeza (approved):

carlosdauden:

sergio-teruel (approved):

victoralmau (approved):

sergio-teruel:
@carlosdauden, do the fw to new versions

PR #1317, opened 2025-05-12
[17.0][FIX] account_financial_report: Amount accumulated several times using hierarchy

carlosdauden:
FW-Port https://github.com/OCA/account-financial-reporting/pull/1315

When the group structure does not have all possible siblings defined, the amounts are accumulated several times to the ancestors.

With this order we make sure that the lower level groups are processed before their children to avoid that when the parent is processed the balance has been updated by the child.

@Tecnativa TT56255

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -5,6 +5,7 @@
 
 
 from odoo import _, api, models
+from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


carlosdauden:

sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1327, opened 2025-06-03
[18.0][FIX] account_financial_report: fix for "group_id" is not searchable accessing account_ids

### LINKED ISSUE #1306: [18.0] account_financial_report: error non-stored field cannot be searched breaks odoo sh project update
## Module
account_financial_report

## Describe the bug
The field group_id is not stored field  in v18. 
https://github.com/OCA/account-financial-reporting/blob/c701d61b79d651fea3376a9fed1fdc4e7d646af6/account_financial_report/models/account_group.py#L15
This causes the error message in the logs

> 2025-04-08 23:55:20,702 21 ERROR ############## odoo.osv.expression: Non-stored field account.account.group_id cannot be searched

Image

In an Odoo sh project the error log breaks the code update

## To Reproduce
**Affected versions**: 18.0
**Setps:**
1. Install account_financial_report module in odoo sh project
2. Update odoo sh project or only see the error log in the logs

**Expected behavior**
Error free log console

micheledic:
### Before this PR
The group_id is not stored so it appear "group_id" is not searchable accessing account_ids

Fixes https://github.com/OCA/account-financial-reporting/issues/1306

### DIFF
--- a/account_financial_report/models/account_group.py
+++ b/account_financial_report/models/account_group.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
+from odoo.tools import SQL
 
 
 class AccountGroup(models.Model):
@@ -12,14 +13,17 @@ class AccountGroup(models.Model):
     )
     level = fields.Integer(compute="_compute_level", recursive=True)
     account_ids = fields.One2many(
-        comodel_name="account.account", inverse_name="group_id", string="Accounts"
+        comodel_name="account.account",
+        compute="_compute_account_ids",
+        string="Accounts",
+        store=False,
     )
     compute_account_ids = fields.Many2many(
         "account.account",
         recursive=True,
         compute="_compute_group_accounts",
         string="Compute accounts",
-        store=True,
+        store=False,
     )
     complete_name = fields.Char(
         "Full Name", compute="_compute_complete_name", recursive=True
@@ -36,6 +40,44 @@ class AccountGroup(models.Model):
                 group.complete_name = f"{group.parent_id.complete_name}/{group.name}"
             else:
                 group.complete_name = group.name
+
+    @api.depends("code_prefix_start", "code_prefix_end")
+    def _compute_account_ids(self):
+        """Retrieves every account from `self`.
+        In Odoo 18 the group_id on account is not stored so it raises
+        an error the one2many account_ids with inverse name group_id."""
+        group_ids = self.ids
+        self.account_ids = self.env["account.account"]
+        if not group_ids:
+            return
+        group_ids = SQL(",".join(map(str, group_ids)))
+        results = self.env.execute_query(
+            SQL(
+                """
+        SELECT
+        agroup.id AS group_id,
+        STRING_AGG(DISTINCT account.id::text, ', ') as account_ids
+        FROM  account_group agroup
+        inner join account_account account
+        ON agroup.code_prefix_start <= LEFT(%(code_store)s->>agroup.company_id::text,
+        char_length(agroup.code_prefix_start))
+        AND agroup.code_prefix_end >= LEFT(%(code_store)s->>agroup.company_id::text,
+        char_length(agroup.code_prefix_end))
+        WHERE agroup.id IN (%(group_ids)s)
+        GROUP BY agroup.id
+            """,
+                code_store=SQL.identifier("account", "code_store"),
+                group_ids=group_ids,
+            )
+        )
+        group_by_code = dict(results)
+        if not group_by_code:
+            return
+        for record in self:
+            if group_by_code.get(record.id, ""):
+                record.account_ids = list(
+                    map(int, group_by_code.get(record.id, "").split(", "))
+                )
 
     @api.depends("code_prefix_start", "parent_id.complete_code")
     def _compute_complete_code(self):
@@ -57,7 +99,6 @@ class AccountGroup(models.Model):
                 group.level = group.parent_id.level + 1
 
     @api.depends(
-        "account_ids",
         "group_child_ids.compute_account_ids",
     )
     def _compute_group_accounts(self):
--- a/account_financial_report/tests/test_trial_balance.py
+++ b/account_financial_report/tests/test_trial_balance.py
@@ -38,12 +38,10 @@ class TestTrialBalanceReport(AccountTestInvoicingCommon):
             {
                 "code": "001",
                 "name": "Account 001",
-                "group_id": cls.group2.id,
                 "account_type": "income_other",
             },
         )
         cls.account100 = cls.company_data["default_account_receivable"]
-        cls.account100.group_id = cls.group1.id
         cls.account110 = cls.env["account.account"].search(
             [
                 (
@@ -59,7 +57,6 @@ class TestTrialBalanceReport(AccountTestInvoicingCommon):
             {
                 "code": "200",
                 "name": "Account 200",
-                "group_id": cls.group2.id,
                 "account_type": "income_other",
             },
         )
@@ -76,7 +73,6 @@ class TestTrialBalanceReport(AccountTestInvoicingCommon):
             {
                 "code": "201",
                 "name": "Account 201",
-                "group_id": cls.group2.id,
                 "account_type": "income_other",
             },
         )
@@ -100,8 +96,6 @@ class TestTrialBalanceReport(AccountTestInvoicingCommon):
 
     def _create_account_account(self, vals):
         item = self.env["account.account"].create(vals)
-        if "group_id" in vals:
-            item.group_id = vals["group_id"]
         return item
 
     def _add_move(


micheledic:
@OCA/accounting-maintainers

micheledic:
@OCA/accounting-maintainers  ping

micheledic:
@OCA/accounting-maintainers ping

account_financial_report/models/account.py
```diff
@@ -11,3 +11,14 @@ class AccountAccount(models.Model):
         "the General Ledger report (the webkit one only), "
         "only centralized amounts per period.",
     )
+
+    def write(self, vals):
+        res = super().write(vals)
+        self.env["account.group"].invalidate_model(["account_ids"])
```
monen17: Could you please explain why this is needed?

account_financial_report/models/account.py
```diff
@@ -11,3 +11,14 @@ class AccountAccount(models.Model):
...
+        return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        self.env["account.group"].invalidate_model(["account_ids"])
```
monen17: Could you please explain why this is needed?

account_financial_report/models/account_group.py
```diff
@@ -28,14 +32,45 @@ class AccountGroup(models.Model):
...
-    def _compute_complete_name(self):
-        """Forms complete name of location from parent location to child location."""
-        for group in self:
-            if group.parent_id.complete_name:
-                group.complete_name = f"{group.parent_id.complete_name}/{group.name}"
-            else:
-                group.complete_name = group.name
+    @api.depends_context("company")
+    def _compute_account_ids(self):
+        """Retrieves every account from `self` and `self`'s subgroups.
```
monen17: The field `account_ids` used to only contain the accounts of group `self`; the accounts of the subgroups are added in field `compute_account_ids`.

Please only retrieve the accounts of group `self` so the existing behavior is preserved.

account_financial_report/models/account_group.py
```diff
@@ -28,14 +32,45 @@ class AccountGroup(models.Model):
         "Full Code", compute="_compute_complete_code", recursive=True
     )
 
-    @api.depends("name", "parent_id.complete_name")
-    def _compute_complete_name(self):
```
monen17: Why has this been removed? How are we computing the `complete_name` now?

account_financial_report/models/account_group.py
```diff
@@ -28,14 +32,45 @@ class AccountGroup(models.Model):
...
+                """
+SELECT
+ agroup.id AS group_id,
+STRING_AGG(DISTINCT account.id::text, ', ') as account_ids
+FROM  account_group agroup
+left join account_account account
+ON agroup.code_prefix_start <= LEFT(%(code_store)s->>%(root_company_id)s,
+char_length(agroup.code_prefix_start))
+AND agroup.code_prefix_end >= LEFT(%(code_store)s->>%(root_company_id)s,
+char_length(agroup.code_prefix_end))
```
monen17: Could you please clarify what is the `self.env.company.root_id.id` doing in a filter on account group prefixes?

account_financial_report/models/account_group.py
```diff
@@ -12,14 +13,17 @@ class AccountGroup(models.Model):
     )
     level = fields.Integer(compute="_compute_level", recursive=True)
     account_ids = fields.One2many(
-        comodel_name="account.account", inverse_name="group_id", string="Accounts"
+        comodel_name="account.account",
+        compute="_compute_account_ids",
+        string="Accounts",
+        store=False,
```
monen17, line 19: Please add a test verifying that the bug isn't happening anymore, I assume the `search` isn't working properly before this PR?

monen17 (changes requested):
Thanks for the PR!
Please shorten the commit title because it is cut with ellipsis: 
image
See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message:
> please check if the commit message is cut with ellipsis.

This should be enough:
> [FIX] account_financial_report: `group_id` is not searchable

any other detail can go in the commit description

As I said in https://github.com/OCA/account-financial-reporting/issues/1306#issuecomment-3036367307 the issue isn't very clear to me so I can't check if this PR is fixing it.
Only did the code review and there are some points that need clarification.

account_financial_report/models/account_group.py
micheledic: Hello,
on _compute_account_ids i build a SQL query that is the reverse of  the compute of the group
https://github.com/odoo/odoo/blob/07626050bd0104fecd8799b56d30245d00da3f27/addons/account/models/account_account.py#L471

To compute the child of a specific group, on the 18.0 it will search for group where the code of the account is contained between prefix_start and prefix_end . I build the reverse SQL query of the compute.

The query is build so if the compute is triggered on more than 1 record, the query is done once but it will compute all the group in self

account_financial_report/models/account_group.py
micheledic: my fault, i restore the compute

account_financial_report/models/account_group.py
micheledic: it is used on odoo 18.0 compute https://github.com/odoo/odoo/blob/18.0/addons/account/models/account_account.py#L471
I build the query so it will retrieve the information consistent with the account group compute . 
I guess it is used because if a company is a branch, it will be retrieved the information only of the data from the json field "code" for the root company

account_financial_report/models/account.py
micheledic: The account_ids is now computed so the depends on
https://github.com/OCA/account-financial-reporting/blob/1ffec3d4ee13709a4c2c30d0b21ac3324dc7b75f/account_financial_report/models/account_group.py#L60
will never be called.
I have not found another way to make the compute work , any idea?

account_financial_report/models/account_group.py
micheledic: It is an error when you access the account_ids field, i guess a test can't be done . It is an access error on a one2many with a compute not stored inverse field

account_financial_report/models/account.py
monen17: The field `account.group.account_ids` should be computed on the fly every time it is needed, so I think there is no need for this cache invalidation, could you please check?

Moreover, this is done every time any field of an account is updated: is it really necessary to invalidate the cache if an account changes its name? Or it becomes reconcilable?

account_financial_report/models/account_group.py
monen17: Access errors can be tested using `assertRaises`.
Do you mean that the field `account.group.account_ids` actually returns the correct value and the only problem is the ERROR in the logs?

If that's the case, any test that accesses `account.group.account_ids` should be enough, then we can check that there are no ERROR logs in the CI.

account_financial_report/models/account_group.py
monen17: Ok so you are not actually computing the accounts of
> `self`'s subgroups.

if that's the case, please remove this part of the docstring because it's misleading.

account_financial_report/models/account_group.py
monen17: The linked query is
```sql
                 SELECT DISTINCT ON (account_code.code)
                        account_code.code,
                        agroup.id AS group_id
                   FROM (VALUES %(account_code_values)s) AS account_code (code)
              LEFT JOIN account_group agroup
                     ON agroup.code_prefix_start <= LEFT(account_code.code, char_length(agroup.code_prefix_start))
                        AND agroup.code_prefix_end >= LEFT(account_code.code, char_length(agroup.code_prefix_end))
                        AND agroup.company_id = %(root_company_id)s
               ORDER BY account_code.code, char_length(agroup.code_prefix_start) DESC, agroup.id
```
(from https://github.com/odoo/odoo/blob/9f8c364f056c8937409f8ed91b8d1fa436c2d7c0/addons/account/models/account_account.py#L473-L481)

In the linked query, the `root_company_id` is only used in the last condition
> AND agroup.company_id = %(root_company_id)s

but in this PR's query `root_company_id` is used in combination with `code_prefix_start` and `code_store`:
> ON agroup.code_prefix_start <= LEFT(%(code_store)s->>%(root_company_id)s,
char_length(agroup.code_prefix_start))
AND agroup.code_prefix_end >= LEFT(%(code_store)s->>%(root_company_id)s,
char_length(agroup.code_prefix_end))

 could you please clarify why?

Moreover, I can't see any `code_store` in the linked query, how is it in this PR's query if this PR's query is just the opposite of the linked query?

account_financial_report/models/account_group.py
monen17: :+1: Thanks

monen17 (changes requested):

micheledic:
@monen17 The code_store is a JSON field within account.account. To access the code for a specific company, we have to use field_name->>index. For example, to get the code for company ID 1, you'd use code_store->>1. In the account_account compute query, you already have the code needed to determine the group, so there's no need to retrieve data from the code_store JSON.
In my case you are computing the account of a specific group , but you have to retrieve the data from the query. The fastest way is a single query with the comparison directly in SQL  getting the data from the json field

account_financial_report/models/account_group.py
micheledic: The code_store is a JSON field within account.account. To access the code for a specific company, we have to use field_name->>index. For example, to get the code for company ID 1, you'd use code_store->>1. In the account_account compute query, you already have the code needed to determine the group, so there's no need to retrieve data from the code_store JSON.
In my case you are computing the account of a specific group , but you have to retrieve the data from the query. The fastest way is a single query with the comparison directly in SQL getting the data from the json field

account_financial_report/models/account_group.py
micheledic: done

account_financial_report/models/account_group.py
micheledic: yes the problem is the error in logs

account_financial_report/models/account.py
micheledic: I have tried without invalidate but the test are failing because account_ids does not contain the account_id created. May. I guess that the reason is that if on the same transaction you create an account and than retrieve the data with a SQL query, the data of the account created is not flushed yet so it is not retrieven from the sql

micheledic (commented):
fatto

account_financial_report/models/account.py
monen17: Thanks for checking!
Then we have indeed to invalidate the cache when a new account is created.

account_financial_report/models/account.py
monen17: This is done every time any field of an account is updated: is it really necessary to invalidate the cache if an account changes its name? Or it becomes reconcilable?

Maybe we should invalidate the cache only when `code` is modified, could you please check if that's enough?

account_financial_report/models/account_group.py
monen17: If it is already working properly I don't see the need for all this, but since it keeps working it's fine 👍

account_financial_report/models/account_group.py
monen17: Thanks for the explanation!
I was missing that accounts are multi-company in `18.0`, then `code` is computed and `code_store` stores the account code for each account in the different companies.
I'm not 100% convinced of this query and would have preferred using the ORM but looks like it's working fine so it's ok :+1:

account_financial_report/models/account_group.py
```diff
@@ -37,6 +41,46 @@ def _compute_complete_name(self):
...
+                root_company_id=str(self.env.company.root_id.id),
+            )
+        )
+        group_by_code = dict(results)
+        self.account_ids = self.env["account.account"]
+        if not group_by_code:
+            return
+        for record in self:
+            record.account_ids = list(
+                map(int, group_by_code.get(record.id, "").split(", "))
```
monen17: Trying this in runboat results in an error for the following use-case:
1. Create a group
3. Create a new company
4. Open the group
5. Select both companies (the original company is still the active one)
6. Select the new company as active

**Actual**
This error is raised:
<details><summary>Trace</summary>
<pre>
  File "/opt/odoo/odoo/http.py", line 2123, in _transactioning
    return service_model.retrying(func, env=self.env)
  File "/opt/odoo/odoo/service/model.py", line 156, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 2090, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 2338, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 333, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 754, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 36, in call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 533, in call_kw
    result = getattr(recs, name)(*args, **kwargs)
  File "/opt/odoo/addons/web/models/models.py", line 86, in web_read
    values_list: list[dict] = self.read(fields_to_read, load=None)
  File "/opt/odoo/odoo/models.py", line 3843, in read
    return self._read_format(fnames=fields, load=load)
  File "/opt/odoo/odoo/models.py", line 4074, in _read_format
    vals[name] = convert(record[name], record, use_display_name)
  File "/opt/odoo/odoo/models.py", line 7061, in __getitem__
    return self._fields[key].__get__(self)
  File "/opt/odoo/odoo/fields.py", line 4620, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/odoo/fields.py", line 3059, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/odoo/fields.py", line 1303, in __get__
    self.compute_value(recs)
  File "/opt/odoo/odoo/fields.py", line 1485, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/odoo/models.py", line 5280, in _compute_field_value
    fields.determine(field.compute, self)
  File "/opt/odoo/odoo/fields.py", line 110, in determine
    return needle(*args)
  File "/mnt/data/odoo-addons-dir/account_financial_report/models/account_group.py", line 81, in _compute_account_ids
    map(int, group_by_code.get(record.id, "").split(", "))
AttributeError: 'NoneType' object has no attribute 'split'
</pre>
</details>

**Expected**
no error

Here is the video of the error being raised:

https://github.com/user-attachments/assets/195746f6-ec6f-4dd9-bce8-ec2cdbb808dd

monen17 (changes requested):

account_financial_report/models/account_group.py
micheledic: It is better using a query because if we have to do it by ORM, it will be slowly becuase it will not use all the index.
However now i rewritten the query, before it was bugged.
Does not have any sense to pass the company in the environment because every group is linked to a company so i will do a single query that get the correct account_ids for the company we are on .

account_financial_report/models/account.py
micheledic: I rewrited the compute now it is not needed to invalidate the model

account_financial_report/models/account_group.py
micheledic: fixed by rewrite of the query

account_financial_report/models/account_group.py
```diff
@@ -37,6 +41,44 @@ def _compute_complete_name(self):
...
+char_length(agroup.code_prefix_end))
+AND agroup.id IN (%(group_ids)s)
+GROUP BY agroup.id
+            """,
+                code_store=SQL.identifier("account", "code_store"),
+                group_ids=group_ids,
+            )
+        )
+        group_by_code = dict(results)
+        self.account_ids = self.env["account.account"]
```
monen17:
```suggestion
```
This is already assigned at the beginning of the method and hasn't changed since.

account_financial_report/models/account_group.py
monen17: Thanks!
I tried the same steps as above and no error is raised :clap: 
but as soon as I edit the `code_prefix_end` of the created group the following is raised:
<details><summary>Details</summary>
<pre>
  File "/opt/odoo/odoo/http.py", line 2123, in _transactioning
    return service_model.retrying(func, env=self.env)
  File "/opt/odoo/odoo/service/model.py", line 156, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 2090, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 2338, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 333, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 754, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 36, in call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 533, in call_kw
    result = getattr(recs, name)(*args, **kwargs)
  File "/opt/odoo/addons/web/models/models.py", line 1012, in onchange
    todo = [
  File "/opt/odoo/addons/web/models/models.py", line 1015, in <listcomp>
    if field_name not in done and snapshot0.has_changed(field_name)
  File "/opt/odoo/addons/web/models/models.py", line 1129, in has_changed
    return self[field_name].keys() != set(self.record[field_name]._ids) or any(
  File "/opt/odoo/odoo/models.py", line 7061, in __getitem__
    return self._fields[key].__get__(self)
  File "/opt/odoo/odoo/fields.py", line 4620, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/odoo/fields.py", line 3059, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/odoo/fields.py", line 1303, in __get__
    self.compute_value(recs)
  File "/opt/odoo/odoo/fields.py", line 1485, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/odoo/models.py", line 5280, in _compute_field_value
    fields.determine(field.compute, self)
  File "/opt/odoo/odoo/fields.py", line 110, in determine
    return needle(*args)
  File "/mnt/data/odoo-addons-dir/account_financial_report/models/account_group.py", line 78, in _compute_account_ids
    record.account_ids = list(
ValueError: invalid literal for int() with base 10: ''
</pre>
</details>
Video is 

https://github.com/user-attachments/assets/acd1a355-b57b-433a-8206-50438d289849

account_financial_report/models/account_group.py
```diff
@@ -37,6 +41,44 @@ def _compute_complete_name(self):
...
+SELECT
+ agroup.id AS group_id,
+STRING_AGG(DISTINCT account.id::text, ', ') as account_ids
+FROM  account_group agroup
+inner join account_account account
+ON agroup.code_prefix_start <= LEFT(%(code_store)s->>agroup.company_id::text,
+char_length(agroup.code_prefix_start))
+AND agroup.code_prefix_end >= LEFT(%(code_store)s->>agroup.company_id::text,
+char_length(agroup.code_prefix_end))
+AND agroup.id IN (%(group_ids)s)
```
monen17:
```suggestion
WHERE agroup.id IN (%(group_ids)s)
```
I don't know if it changes the result of the query, but it seems reasonable for this to be a WHERE condition, let me know what you think.

monen17 (changes requested):

account_financial_report/models/account_group.py
micheledic: fixed

account_financial_report/models/account_group.py
micheledic: removed

account_financial_report/models/account_group.py
micheledic: fixed!

micheledic:
@OCA/accounting-maintainers ping

micheledic:
@OCA/accounting-maintainers  ping

micheledic:
@OCA/accounting-maintainers ping

account_financial_report/models/account_group.py
```diff
@@ -37,6 +41,44 @@ def _compute_complete_name(self):
...
+        an error the one2many account_ids with inverse name group_id."""
+        group_ids = self.ids
+        self.account_ids = self.env["account.account"]
+        if not group_ids:
+            return
+        group_ids = SQL(",".join(map(str, group_ids)))
+        results = self.env.execute_query(
+            SQL(
+                """
+SELECT
```
pedrobaeza: Please don't use SQL and do it by ORM.

account_financial_report/models/account_group.py
```diff
@@ -12,14 +13,17 @@ class AccountGroup(models.Model):
...
+        string="Accounts",
+        store=False,
     )
     compute_account_ids = fields.Many2many(
         "account.account",
         recursive=True,
         compute="_compute_group_accounts",
         string="Compute accounts",
-        store=True,
+        store=False,
```
pedrobaeza, line 26: Why changing this?

pedrobaeza (changes requested):

account_financial_report/models/account_group.py
micheledic: The query compares the same character length of the prefix length, it is not possible in Odoo ORM. because it means that we have to do a search with the WHOLE database of account and filter them in python.
It is very slowly than a query that uses indexes.
The prefix start and end may have different length so the logic is to compare the same length of character of the prefix and it is not possible with the search but only with filtered on the entire databse recordset

account_financial_report/models/account_group.py
micheledic: Because the account_ids are computed with a query, it is not possible to add any depends on it ,a query is needed because there is not any direct relation between groups and account and the hierarchy can change at every time

account_financial_report/models/account_group.py
pedrobaeza: Uhm, I don't think the penalization is so high, as the number of accounts is limited (thousands at maximum). Anyway, at least put the query at the same indentation level that corresponds to Python code.

account_financial_report/models/account_group.py
pedrobaeza: OK

account_financial_report/models/account_group.py
micheledic: done!

pedrobaeza (approved):
OK, let's fix the problem

/ocabot merge patch

pedrobaeza:
Thanks for the patch

micheledic:
@pedrobaeza fixed the pre-commit error, now you can retrigger the merge!

PR #1333, opened 2025-06-10
[IMP] partner_statement: Show only overdue lines

SirPyTech:
Allow the user to only show overdue invoices in partner statements.

### DIFF
--- a/partner_statement/report/activity_statement.py
+++ b/partner_statement/report/activity_statement.py
@@ -125,6 +125,7 @@ class ActivityStatement(models.AbstractModel):
         excluded_accounts_ids = tuple(
             self.env.context.get("excluded_accounts_ids", [])
         ) or (-1,)
+        show_only_overdue = self.env.context.get("show_only_overdue", False)
         payment_ref = _("Payment")
         return str(
             self._cr.mogrify(
@@ -166,6 +167,11 @@ class ActivityStatement(models.AbstractModel):
                 AND %(date_start)s <= l.date
                 AND l.date <= %(date_end)s
                 AND m.state IN ('posted')
+                AND CASE
+                    WHEN %(show_only_overdue)s
+                    THEN COALESCE(l.date_maturity, l.date) <= %(date_end)s
+                    ELSE TRUE
+                END
             GROUP BY l.partner_id, m.name, l.date, l.date_maturity,
                 CASE WHEN (aj.type IN ('sale', 'purchase'))
                     THEN l.name
--- a/partner_statement/report/outstanding_statement.py
+++ b/partner_statement/report/outstanding_statement.py
@@ -17,6 +17,7 @@ class OutstandingStatement(models.AbstractModel):
         excluded_accounts_ids = tuple(
             self.env.context.get("excluded_accounts_ids", [])
         ) or (-1,)
+        show_only_overdue = self.env.context.get("show_only_overdue", False)
         return str(
             self._cr.mogrify(
                 """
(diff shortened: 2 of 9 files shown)


partner_statement/report/activity_statement.py
```diff
@@ -125,6 +125,7 @@ def _display_activity_lines_sql_q1(
         excluded_accounts_ids = tuple(
             self.env.context.get("excluded_accounts_ids", [])
         ) or (-1,)
+        show_only_overdue = self.env.context.get("show_only_overdue")
```
MiquelRForgeFlow:
```suggestion
        show_only_overdue = self.env.context.get("show_only_overdue", False)
```

partner_statement/report/outstanding_statement.py
```diff
@@ -17,6 +17,7 @@ def _display_outstanding_lines_sql_q1(self, partners, date_end, account_type):
         excluded_accounts_ids = tuple(
             self.env.context.get("excluded_accounts_ids", [])
         ) or (-1,)
+        show_only_overdue = self.env.context.get("show_only_overdue")
```
MiquelRForgeFlow:
```suggestion
        show_only_overdue = self.env.context.get("show_only_overdue", False)
```

partner_statement/report/report_statement_common.py
```diff
@@ -53,6 +53,7 @@ def _show_buckets_sql_q1(self, partners, date_end, account_type):
         excluded_accounts_ids = tuple(
             self.env.context.get("excluded_accounts_ids", [])
         ) or (-1,)
+        show_only_overdue = self.env.context.get("show_only_overdue")
```
MiquelRForgeFlow:
```suggestion
        show_only_overdue = self.env.context.get("show_only_overdue", False)
```

partner_statement/report/activity_statement.py
SirPyTech: :+1:

partner_statement/report/outstanding_statement.py
SirPyTech: :+1:

partner_statement/report/report_statement_common.py
SirPyTech: :+1:

Marco-ooops (approved):
test funzionale ok

partner_statement/report/outstanding_statement.py
```diff
@@ -71,6 +72,10 @@ def _display_outstanding_lines_sql_q1(self, partners, date_end, account_type):
                         pc.max_date <= %(date_end)s) OR
                     (pd.id IS NULL AND pc.id IS NULL)
                 ) AND l.date <= %(date_end)s AND m.state IN ('posted')
+                AND CASE
+                    WHEN %(show_only_overdue)s THEN l.date_maturity <= %(date_end)s
```
MiquelRForgeFlow:
```suggestion
                    WHEN %(show_only_overdue)s THEN COALESCE(l.date_maturity, l.date) <= %(date_end)s
```
could you apply this to all?

SirPyTech: Makes sense because the "Date due" shown is actually `date` when `date_maturity` is null:
https://github.com/OCA/account-financial-reporting/blob/a8b5360f696d9d1a2df90ea7f210604e329d6ae3/partner_statement/report/outstanding_statement.py#L45-L48
Thanks!
Please check if it's ok now

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
BTW, I am not sure if this is the good approach, because it may affect the detailed activity statement or other custom reports that inherit from these ones. Let me ask a question: if you only want show only overdue lines, why filter them when calculating data instead of filtering them later, when they are already calculated (maybe in `_get_report_values()` or in the xml templates and xlsx reports?

SirPyTech:
> if you only want show only overdue lines, why filter them when calculating data instead of filtering them later, when they are already calculated (maybe in `_get_report_values()` or in the xml templates and xlsx reports?

I preferred to filter the move lines at the source because otherwise I would have to filter them also in other calculations.
For instance: if the query returned all the lines and let's say that the total is 1000€, if I hide some lines only at the end, the total would still be a 1000€ but the shown lines would sum up to less than a 1000€.

francesco-ooops:
@MiquelRForgeFlow for us it's functionally ok, is it ok to merge as is or you still have doubts?

PR #1334, opened 2025-06-11
[18.0][FIX] account_financial_report:

BhaveshHeliconia:
Fix general ledger report print issue when there are multiple analytic accounts on single analytic distribution line.

- Steps to Reproduce:
  - Create journal entry having more than one analytic accounts on single analytic distribution line and post it.
     
      image (3)
      
      image (4)
      
  - Try to create general ledger report from "Invoicing > Reporting > OCA accounting reports > General Ledger". Following error occurs
   
    image (5)

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -484,7 +484,8 @@ class GeneralLedgerReport(models.AbstractModel):
             for tax_id in move_line["tax_ids"]:
                 taxes_ids.add(tax_id)
             for analytic_account in move_line["analytic_distribution"] or {}:
-                analytic_ids.add(int(analytic_account))
+                for analytic_account_id in analytic_account.split(","):
+                    analytic_ids.add(int(analytic_account_id))
             if move_line["full_reconcile_id"]:
                 rec_id = move_line["full_reconcile_id"][0]
                 if rec_id not in full_reconcile_ids:
--- a/account_financial_report/report/general_ledger_xlsx.py
+++ b/account_financial_report/report/general_ledger_xlsx.py
@@ -195,16 +195,17 @@ class GeneralLedgerXslx(models.AbstractModel):
                             taxes_description += taxes_data[tax_id]["tax_name"] + " "
                         if line["tax_line_id"]:
                             taxes_description += line["tax_line_id"][1]
-                        for account_id, value in line["analytic_distribution"].items():
-                            if value < 100:
-                                analytic_distribution += "%s %d%% " % (
-                                    analytic_data[int(account_id)]["name"],
-                                    value,
-                                )
-                            else:
-                                analytic_distribution += (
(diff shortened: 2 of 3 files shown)


BhaveshHeliconia:
Hey @OCA/accounting-maintainers when you get a chance, kindly review and consider merging it.

pedrobaeza:
Please explain the reasons for this change: steps to reproduce the problem, or the whys, etc.

BhaveshHeliconia:
@pedrobaeza,
Replaced old "company_id" field with "company_ids" at all required places in module due to changes in "account.account" model of odoo v18 core.

Steps to reproduce:
- Goto "Invoicing > Reporting > OCA accounting reports > General Ledger" and generate report.
- Issue in "Reconcile name" column value.
image

account_financial_report/report/general_ledger.py
```diff
@@ -342,7 +342,7 @@ def _get_move_line_data(self, move_line):
             "rec_id": move_line["full_reconcile_id"][0]
             if move_line["full_reconcile_id"]
             else False,
-            "rec_name": move_line["full_reconcile_id"][1]
+            "rec_name": move_line["matching_number"]
```
pedrobaeza: Instead of bringing another field, which is the same:

```suggestion
            "rec_name": str(move_line["full_reconcile_id"][0])
```

pedrobaeza (commented):
Please remove `[18.0]` from the commit message.

BhaveshHeliconia:
@pedrobaeza : 
Thank you for the review.
I've made the requested changes — removed [18.0] from the commit message and avoided introducing a duplicate field by reusing the existing one.
Please let me know if there's anything else to address.

pedrobaeza (approved):
Thanks for the changes.

The first commit message line is orphaned without summary. You can change it to `[FIX] account_financial_report: company_id > company_ids + matching number`

alexis-via:
You modified the fix of @bguillot displaying reconcile ID instead of matching_number. And matching_number is what we expect to see in the general ledger report, because it is what is displayed to the accountant in odoo.

alexis-via:
In fact, matching_number has the ID of the account.full.reconcile when the account.move.line is fully reconciled and has "P" when it is partially reconciled. It's nice to have the info that the line is partially reconciled.
It's why I prefer to stay with @bguillot  initial fix.

BhaveshHeliconia:
@alexis-via : Please review when you get chance

lk-eska:
This PR is nothing to do with matching number (I think it is removed in a later commit). This fixes the analytic distribution problem and I think it should be merged. 

@BhaveshHeliconia would you fix the commit message?
@pedrobaeza can we accept this as is?

pedrobaeza:
@lk-eska indeed the original intention of the PR has been changed meanwhile. Have you tried the current patch that is solving your problem?

lk-eska:
Yes it fixes. @pedrobaeza

PR #1339, opened 2025-06-20
[15.0][FIX] account_financial_report: Taxes option missed in journal ledger.

carlosdauden:
A Typo error causes the oden not to be as expected when searching for multiple reports with the same report_name (which may not be ideal either) and causes the configuration set in the paper format not to be applied.
Update translation terms to avoid the above issue.

@Tecnativa TT56812

### DIFF
--- a/account_financial_report/i18n/es.po
+++ b/account_financial_report/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 12.0+e\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-09-21 15:05+0000\n"
-"PO-Revision-Date: 2024-06-14 14:42+0000\n"
+"POT-Creation-Date: 2025-06-20 23:20+0000\n"
+"PO-Revision-Date: 2025-06-21 01:20+0200\n"
 "Last-Translator: Víctor Martínez <victor.martinez@tecnativa.com>\n"
 "Language-Team: \n"
 "Language: es\n"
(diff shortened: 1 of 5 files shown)


carlosdauden:

victoralmau (approved):
It will be pending to fix the error in CI

carlosdauden:
Rebased to try remove unrelated errors

account_financial_report/i18n/es.po
```diff
@@ -6,46 +6,46 @@ msgid ""
...
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 4.17\n"
+"X-Generator: Poedit 3.4.2\n"
 
 #. module: account_financial_report
 #: model_terms:ir.ui.view,arch_db:account_financial_report.report_aged_partner_balance_lines_header
 msgid "&gt; 120 d."
-msgstr "&gt; 120 d."
+msgstr ""
```
pedrobaeza: These changes shouldn't be done in this PR. I think you have your POEdit to remove the translations equal to the source, but it's creating a big diff and marking the language as not translated, where it was.

pedrobaeza (commented):
Weird option to require on the journal ledger (instead of general ledger) the taxes, but OK, it's an option more. Please check the translation question.

account_financial_report/i18n/es.po
carlosdauden: It's not a configuration option in my POEdit, I removed it manually because that's the trend I see lately, maybe it's weblate who is creating that trend?

pedrobaeza: It's an option in pre-commit that was later deactivated, as it's not convenient. It saves bytes, but on the other hand, it provokes a lot of diffs and translation platform thinks that the language is not translated. In 15.0, it has never been activated.

PR #1341, opened 2025-06-26
[15.0][IMP] account_tax_balance: optional hide for financial_type in tree view

JoanSForgeFlow:
This PR makes the `financial_type` field optional in tree view. While the field makes sense from a technical perspective, it can be confusing from an accounting point of view. We are introducing the option to hide it when it's not relevant.

Additionally, we rename the filter title from "Move Type" to "Financial Type". This is the label shown in both the tree and form views. The previous label was misleading, as "Move Type" could be confused with the standard Odoo field `move_type`.

### DIFF
--- a/account_tax_balance/views/account_move_view.xml
+++ b/account_tax_balance/views/account_move_view.xml
@@ -8,7 +8,7 @@
         <field name="inherit_id" ref="account.view_move_tree" />
         <field name="arch" type="xml">
             <field name="state" position="after">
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

LoisRForgeFlow:
@JoanSForgeFlow does this need fwport?

PR #1347, opened 2025-07-01
[ADD] portal_partner_statement

SirPyTech:
Allow any commercial partner to download their partner statements from the portal.


francesco-ooops (approved):
Functional ok!

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@OCA/accounting-maintainers wdyt?

marcelsavegnago (approved):
LGTM

francesco-ooops:
@OCA/accounting-maintainers good to go? we can forward port it.

francesco-ooops:
@OCA/accounting-maintainers ping :)

portal_partner_statement/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "name": "Partner Statement from portal",
+    "version": "14.0.1.0.0",
+    "category": "Accounting & Finance",
+    "summary": "Download OCA Financial Reports from portal",
+    "author": "PyTech, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/account-financial-reporting",
+    "license": "AGPL-3",
+    "depends": [
+        "partner_statement",
+        "portal",
```
ivs-cetmix: 'partner_statement' module deepens on the 'account' module which already [depends](https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/account/__manifest__.py#L18) on the portal.
So there is no need to add 'portal' in dependencies again.

ivs-cetmix (changes requested):

portal_partner_statement/__manifest__.py
SirPyTech: Thanks for having a look!
I preferred to be explicit about it, but you are right this is redundant.

I have removed it, but was this causing any issues?
Because otherwise I can't see why this can block a PR from merging.

ivs-cetmix: Having a straight chain of inheritance is a good practice to minimise potential issues related to the order in which the `super` methods are applied. So let's follow this rule to avoid any potential issues in the future even if it doesn't create any issues now.

ivs-cetmix (approved):
/ocabot merge nobump

portal_partner_statement/__manifest__.py
SirPyTech: I was more concerned about "what if `portal` features will be extracted from `account`?", but your concern is valid too; thanks for the explanation! :smile:

PR #1348, opened 2025-07-01
[FIX] partner_statement: Create own invoice in test

SirPyTech:
The invoice found by the tests has changed, and it causes a failure in the tests.
Create a new invoice so that tests are not depending on records created by others.


MiquelRForgeFlow:
Why not do the same fix in the `test_show_only_overdue`?

SirPyTech:
> Why not do the same fix in the `test_show_only_overdue`?

Good idea, please check if it's ok now

MiquelRForgeFlow (approved):

PR #1350, opened 2025-07-03
[IMP] partner_statement: Translate statement title

SirPyTech:
Translating the statement titles is difficult, if not impossible in some cases.

For example: in the string "Detailed Supplier Statement ..." right now you can translate "Detailed ", or "Supplier ", but the translation of the whole string might need to change the order of the translated terms.

Another example: the string "Statement" cannot be translated because it is declared in a variable in the template.

After this PR, the whole statement title can be translated properly, for every statement.
Let me know what you think!

### DIFF
--- a/partner_statement/report/detailed_activity_statement.py
+++ b/partner_statement/report/detailed_activity_statement.py
@@ -1,7 +1,7 @@
 # Copyright 2022 ForgeFlow, S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import _, models
 
 from .outstanding_statement import OutstandingStatement
 
@@ -12,6 +12,34 @@ class DetailedActivityStatement(models.AbstractModel):
     _inherit = "report.partner_statement.activity_statement"
     _name = "report.partner_statement.detailed_activity_statement"
     _description = "Partner Detailed Activity Statement"
+
+    def _get_title(self, partner, **kwargs):
+        kwargs["context"] = {
+            "lang": partner.lang,
+        }
+        if kwargs.get("line_type") == "prior_lines":
+            if kwargs.get("account_type") == "receivable":
+                title = _(
+                    "Prior Balance up to %(ending_date)s in %(currency)s", **kwargs
+                )
+            else:
+                title = _(
+                    "Supplier Prior Balance up to %(ending_date)s in %(currency)s",
+                    **kwargs
+                )
+        elif kwargs.get("line_type") == "ending_lines":
+            if kwargs.get("account_type") == "receivable":
+                title = _(
+                    "Ending Balance up to %(ending_date)s in %(currency)s", **kwargs
+                )
+            else:
+                title = _(
+                    "Supplier Ending Balance up to %(ending_date)s in %(currency)s",
+                    **kwargs
+                )
+        else:
+            title = super()._get_title(partner, **kwargs)
+        return title
 
     def _get_account_display_prior_lines(
         self, company_id, partner_ids, date_start, date_end, account_type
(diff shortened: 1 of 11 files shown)


partner_statement/report/detailed_activity_statement.py
```diff
@@ -13,6 +13,32 @@ class DetailedActivityStatement(models.AbstractModel):
     _name = "report.partner_statement.detailed_activity_statement"
     _description = "Partner Detailed Activity Statement"
 
+    def _get_title(self, partner, **kwargs):
+        kwargs["context"] = {
+            "lang": partner.lang,
+        }
+        if kwargs.get("line_type") == "prior_lines":
```
MiquelRForgeFlow, line 20: where is it passed the "line_type"?

partner_statement/report/detailed_activity_statement.py
```diff
@@ -13,6 +13,32 @@ class DetailedActivityStatement(models.AbstractModel):
...
+            if kwargs.get("account_type") == "receivable":
+                title = _(
+                    "Prior Balance up to %(ending_date)s in %(currency)s", **kwargs
+                )
+            else:
+                title = _(
+                    "Supplier Prior Balance up to %(ending_date)s in %(currency)s",
+                    **kwargs
+                )
+        else:
```
MiquelRForgeFlow:
```suggestion
elif kwargs.get("line_type") == "ending_lines":
```

partner_statement/report/detailed_activity_statement.py
```diff
@@ -13,6 +13,32 @@ class DetailedActivityStatement(models.AbstractModel):
     _name = "report.partner_statement.detailed_activity_statement"
     _description = "Partner Detailed Activity Statement"
 
+    def _get_title(self, partner, **kwargs):
```
MiquelRForgeFlow, line 16:
```suggestion
    def _get_title(self, partner, **kwargs):
        title = super()._get_title(partner, **kwargs)
```

partner_statement/report/detailed_activity_statement.py
SirPyTech: The detailed activity statement actually prints the `outstanding_balance` template: https://github.com/OCA/account-financial-reporting/blob/e46dc9a45f4aeed7f6c0bd0540b2ff083567dc25/partner_statement/views/detailed_activity_statement.xml#L58

So you can find it in the `outstanding_balance` template.

partner_statement/report/detailed_activity_statement.py
SirPyTech: This was my first implementation, but it was raising an error because `super` expects `starting_date` too.
I _could_ add a big `if` in `super` that does nothing if `starting_date` isn't there, but then we'd have a `super` call that does nothing and that didn't look great.

partner_statement/report/detailed_activity_statement.py
SirPyTech: 👍
I added an `else` just to be sure there always is a `title`.

SirPyTech (commented):
Thanks for the review!

Marco-ooops (approved):
test funzionale ok

francesco-ooops:
@MiquelRForgeFlow we good to go? thanks!

MiquelRForgeFlow:
@SirPyTech 

Without your PR:
Selection_4754

With your PR:
Selection_4755

partner_statement/report/detailed_activity_statement.py
SirPyTech: In order to fix the bug you found in https://github.com/OCA/account-financial-reporting/pull/1350#issuecomment-3045515158, I added a call to `super`; big thanks for catching it!
Please check if that's ok now.

MiquelRForgeFlow (approved):

PR #1356, opened 2025-07-17
[17.0][IMP] account_financial_reporting: Improve analytic distribution display and show amounts per percentage in general ledger

BernatObrador:
Enhanced the layout of analytic account distribution in the general ledger report.
Now, alongside the analytic account name and percentage, the corresponding monetary amount is also shown.
This improves readability and helps users better understand the distribution of amounts across cost centers.

cc https://github.com/APSL HT02234

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review

### DIFF
--- a/account_financial_report/report/templates/general_ledger.xml
+++ b/account_financial_report/report/templates/general_ledger.xml
@@ -494,6 +494,7 @@
                                                 t-att-res-id="analytic_id"
                                                 res-model="account.analytic.account"
                                                 view-type="form"
+                                                style="display: inline-flex; justify-content: space-between; width: 100%;"
                                             >
                                                 <t
                                                     t-set="analytic_id"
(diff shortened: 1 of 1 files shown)


mpascuall (approved):
LGTM!

peluko00 (approved):
LGTM!

ppyczko (approved):
LGTM

pedrobaeza:
Can you put a screenshot of the before/after?

BernatObrador:
New:
image

Old:
image

account_financial_report/report/templates/general_ledger.xml
```diff
@@ -503,14 +504,26 @@
...
-                                                        t-out="account_name + ' (' + str(percentage) + '%)'"
+                                                        t-set="amount"
+                                                        t-value="line['debit'] if line['debit'] else line['credit']"
                                                     />
-                                                </t>
-                                                <t t-else="">
-                                                    <t t-out="account_name" />
-                                                </t>
+                                                    <t
+                                                        t-out="amount * (percentage / 100.0)"
```
pedrobaeza, line 524: This should be shown only if the percentage is lower than 100.

BernatObrador: Done :heavy_check_mark:

account_financial_report/report/templates/general_ledger.xml
```diff
@@ -507,14 +504,27 @@
...
+                                                            />%)</span>
+                                                    </t>
+                                                </span>
+                                                <span
+                                                    style="font-size: 90%; color: #444;"
+                                                >
                                                     <t
-                                                        t-out="account_name + ' (' + str(percentage) + '%)'"
+                                                        t-set="amount"
+                                                        t-value="line['debit'] if line['debit'] else line['credit']"
```
pedrobaeza, line 520: Shouldn't this be conditional as well? What this represents and why not using `balance`?

BernatObrador: The thing is that here should be debit or credit, because balance it's the total accumulated, not the amount for that line,

So it represents the total of the current line, and it's used to split it for the % of every analytic account.

pedrobaeza: And the conditional thing?

BernatObrador: It's indeed a conditional itself, `line['debit'] if line['debit'] else line['credit']`, and it will always be debit or credit if it has analytic, so it's no need to add a specific t-if for that.

pedrobaeza: But it will be shown always, no matter if the percentage is 100%.

BernatObrador: This is only setting up the value amount, it will not be shown when it's 100%, because here is where it's printed:
`
 <t
          t-if="percentage &lt; 100"
          t-out="amount * (percentage / 100.0)"
`

pedrobaeza: OK, true.

pedrobaeza (approved):
/ocabot merge minor

You can fw-port it to upper versions

PR #1321, opened 2025-05-22
[REF] partner_statement: Allow to override XLSX structure

SirPyTech:
I'm proposing this refactoring because I have to add columns to the outstanding report.
In the views I can use xpaths to add the new column.
In the xlsx it would require some reverse-engineering that is possible, but discouraged (by the author itself, see https://stackoverflow.com/questions/62865195/python-xlsxwriter-extract-value-from-cell#comment111848986_63251349).

This refactoring separates preparing the data from writing them that I believe is good practice too.
Let me know what you think!

PR #1357, opened 2025-07-17
[14.0][REF] partner_statement: Allow to override XLSX structure (2nd part)

MiquelRForgeFlow:
Continuation of https://github.com/OCA/account-financial-reporting/pull/1321.


MiquelRForgeFlow:
@SirPyTech

PR #1270, opened 2025-01-23
[18.0][MIG] migration partner_statement to 18.0

psugne:
Note, this module has dependency from https://github.com/OCA/reporting-engine/tree/17.0/report_xlsx_helper which is not merged yet ([PR](https://github.com/OCA/reporting-engine/pull/958))

PR #1364, opened 2025-07-22
[18.0][MIG] partner_statement (final PR)

MiquelRForgeFlow:
This PR restores https://github.com/OCA/account-financial-reporting/pull/1285, but adding last changes in v17.

This PR supersedes https://github.com/OCA/account-financial-reporting/pull/1284 and https://github.com/OCA/account-financial-reporting/pull/1270.

BTW, https://github.com/OCA/account-financial-reporting/pull/1285 was closed by the argument that adding in the "blocked" feature into a report is poor practice and not with this modules rationale.

But I have to say, that the "blocked" feature had a purpose. This feature was moved to Enterprise in some way, and people that we don't use Enterprise has to be able to still use it. If we had some lines "blocked" before in our DBs, the report now would show the same results. Otherwise, different result would show and that would be a problem. ~~That's why instead of directly adding the feature in the module, a new module has been created to preserve that feature, and then added as dependency.~~
EDIT: After some thought, I re-added again the feature in the module instead of having a different module for that feature.


AaronHForgeFlow (approved):
I agree with the arguments given

gdgellatly:
I really really really really disagree. How do changes like this get decided and merged in a day, Pretty much forcing everyone to fork who doesn't want it.

@MiquelRForgeFlow Let's be crystal clear, the ability to block move lines has absolutely nothing to do with a partner statement and nothing to do with reporting. This absolutely should not be in here and the process by which this was slid in, with zero consultation, just an internal, you're my colleague, so I will just approve it and quickly merge is very concerning

\cc @OCA/board

pedrobaeza:
@gdgellatly the flag at `account.move.line` is just a tool for this module for not showing certain information. IMO, it's worse to have a separate module adding this field, and make this module dependent of that other module (because creating another extra glue module is too much burden for just this feature).

If you don't want that field, you just need to hide it in custom/extra module without forking.

And anyway, as you mention @OCA/board, this is only using the same rules as they have made in OCA/bank-payment, where the opinion of few have prevailed over the general good.

JordiBForgeFlow:
@MiquelRForgeFlow We should avoid doing this kind of quick merges, only with approvals from our team, and also knowing that there's an open discussion about a topic that one of the members have explicitly raised.

gdgellatly:
> @gdgellatly the flag at `account.move.line` is just a tool for this module for not showing certain information. IMO, it's worse to have a separate module adding this field, and make this module dependent of that other module (because creating another extra glue module is too much burden for just this feature).
> 
> If you don't want that field, you just need to hide it in custom/extra module without forking.
> 
> And anyway, as you mention @OCA/board, this is only using the same rules as they have made in OCA/bank-payment, where the opinion of few have prevailed over the general good.

@pedrobaeza No I disagree, the blocked field is a major source of pain outside of Europe. In any case, it is a feature that does not belongs in a reporting repository. The purpose of partner statement is to print a statement for a partner, not to add a function to block move lines. 

And it also should not be a dependency of this module. The correct way would have been to create the module bringing the blocked field in the correct repo, some accounting workflow repo or credit control, then refactor partner-statement to allow dynamic adding to statements and then create a third module to bring back in blocked for those using the other module.

But you are also right, I would expect that if they are bringing such a new feature, it would be in a fork, not in the main repo silently merged with no discussion.

EDIT: You must understand, that fundamentally, if this is allowed to stand, and you want statements in OCA, because of a historical field in Odoo, even in 10 years time, you must add an extra feature you don't want, and then it is your problem to get rid of it. This is lunacy.

pedrobaeza:
I don't agree with your way of reasoning. If a feature has been there for years and people expect to have it in the module, it's reasonable to re-add that field if it's missing in the core to continue having the same feature set. And IMO, you are doing a world just for a field. If it was a problem during 10 years, didn't you put a solution though and extra module hiding such field? You can apply the same solution now.

MiquelRForgeFlow:
Sorry people, maybe I should have waited a bit more to get more approvals (which sure I would have get them). I will take into account this for the next time.

JordiBForgeFlow:
@pedrobaeza I agree with @gdgellatly. Someone could want to use the 'blocked' field and not have to depend on a module like this one, that is to be used merely for reporting purposes. My recommendation would be to revert the field on this module and then, if someone needs the field (I doubt it), make an account_move_line_blocked module and then use glue modules with partner_statement. @MiquelRForgeFlow Let's check who required this field and challenge if it's necessary.

pedrobaeza:
I don't agree, but it's your employee. First of all, there's no such need for now that someones requires a field for "other purposes". Anyway, if needed, it can be added several times, or renamed to be more specific (Like "Don't appear in partner statement"), but the straight way for migration is to keep it as is. I use the KISS principle in this. But as said, as you prefer. I don't have any benefit on one way or another.

gdgellatly:
> I don't agree, but it's your employee. First of all, there's no such need for now that someones requires a field for "other purposes". Anyway, if needed, it can be added several times, or renamed to be more specific (Like "Don't appear in partner statement"), but the straight way for migration is to keep it as is. I use the KISS principle in this. But as said, as you prefer. I don't have any benefit on one way or another.

The problem here is we disagree what the straight way is. Odoo used to have a feature to block invoices. They removed it. That doesn't mean bring the feature back in a reporting module. Fields are removed all the time, hell entire models are, we don't bring them all back in a report, just because the report used to rely on it. 

This field should absolutely be in the account_invoice_blocking module. And it should be reasonably telling that after a year it is not migrated. It is just a feature that was always not great, caused far more issues than it solved, but what's worse now is adding it back here now means you will have 2 different results between other reports that don't use this now deleted feature and your statements. It is a lot more than just hiding a field (which is actually already hidden).

Honestly part of me is like, OK I deal with it for v18, have the column but don't show it anywhere, with a note to scrap in v19, by then people have had 2 full years to cycle them out. Meanwhile I run on a fork so I don't have to deal with, it's already done anyway, and going through it there is quite a lot of other things I want to hit as well.

JordiBForgeFlow:
@gdgellatly Miquel will send now a PR to remove the field

MiquelRForgeFlow:
Here it is: https://github.com/OCA/account-financial-reporting/pull/1369.

Issue #1363, opened 2025-07-22
[18.0] account_financial_report: incorrect file report.scss


jvargass:
## Module
[account_financial_report](https://github.com/OCA/account-financial-reporting/tree/18.0/account_financial_report)

## Describe the bug
The report.scss file is affecting all Odoo elements, so it converts the links to the color defined in this file.

## To Reproduce
**Affected versions**: 18.0

1. Open Odoo and see all links in color #00337b
<img width="442" height="136" alt="Image" src="https://github.com/user-attachments/assets/3b44ff41-5033-4a68-93d9-b7de6d7fc3a0" />

<img width="323" height="135" alt="Image" src="https://github.com/user-attachments/assets/03627697-a1d7-4683-8826-9b5d7b1ba1f0" />

**Expected behavior**
Everything defined in the SCSS file should be contained in a class used exclusively in reports. Something like:
```
.o_account_financial_reports_page {
      a {
          color: #00337b;
      }
      
      .act_as_table {
          display: table !important;
          background-color: white;
      }
      
      .act_as_row {
          display: table-row !important;
          page-break-inside: avoid;
      }
...
}
```

PR #1366, opened 2025-07-24
[18.0][FIX] account_financial_report: fix scss affect other places

ramiadavid:
As commented here: #1363 

Partial revert of 01f22b825679e2b2cf353e59ac1de9273569922d 
Adding css in web.assets_backend Affects the display of the Odoo interface
Reverts to how it was in v17


micheledic (changes requested):
Thie CSS file doesn't exists. It does not work the PR on runboat

account_financial_report/report/templates/layouts.xml
```diff
@@ -2,14 +2,21 @@
 <odoo>
     <template id="account_financial_report.html_container">
         <t t-call-assets="web.assets_backend" />
+        <link
+            href="/account_financial_report/static/src/css/report_html.css"
```
micheledic, line 6: The file does not exist in that path.
It is in /scss/ instead of /css/

account_financial_report/report/templates/layouts.xml
```diff
@@ -2,14 +2,21 @@
...
         <t t-set="body_classname" t-value="'container'" />
         <t t-call="web.report_layout">
             <t t-out="0" />
         </t>
     </template>
     <template id="account_financial_report.internal_layout">
         <div class="article o_account_financial_reports_page">
-            <t t-call-assets="web.assets_backend" />
+            <link
+                href="/account_financial_report/static/src/css/report.css"
```
micheledic, line 17: same here.

micheledic (approved):
LGTM

account_financial_report/static/src/css/report.css
pedrobaeza, line 1: Why are you changing this path?

micheledic: It is a css file , it 's correct to put in css folder imho

pedrobaeza: All of them goes to `scss` AFAIK. Anyway, that's out of the scope of the PR.

micheledic: > All of them goes to `scss` AFAIK. Anyway, that's out of the scope of the PR.

in `scss ` folder also the `.css` files ? in odoo core i find `css `files in `css` folder and `scss` in `scss `folder

pedrobaeza: @CarlosRoca13 what is the usual way in this? And as said, this is out of scope of the PR. This should only fix the problem itself, not to re-arrange files.

ramiadavid:
@pedrobaeza In this case, what I have done is undo the changes that were made in the migration and that were causing problems, to leave it the same as in v17, but if you prefer, I can modify it.

pedrobaeza:
@ramiadavid thanks. That makes more sense, but please then express that is a partial revert, link to the commit being reverted, etc. Don't be miser with the explanations in the commit message and the PR message.

account_financial_report/static/src/css/report.css
CarlosRoca13: The correct way would be to use a `css` folder for the `.css` files and a `scss` folder for the `.scss` files. However, when it comes to styles related to an Owl component, the styles should be placed inside the component’s folder.

ramiadavid:
@pedrobaeza It's okay now?

pedrobaeza (approved):
/ocabot merge patch

PR #1378, opened 2025-08-11
[17.0][FIX] account_financial_report: Remove buggy empty selection value

pedrobaeza:
Forward-port of #1265 squashing the commits and giving a more comprehensive commit message.

Selecting no value in the "Grouped by" general ledger wizard, but filtering any account, results in an empty display, due to the weird trick of having an empty selection value.

This, joined with the fact of not being able to predefine that empty value with standard tools, makes the switch necessary.

- Set grouped_by as required with non grouped item description.
- Switched selection value from "" (empty) to "none".
- Adapt general_ledger group_by conditionals.

@Tecnativa TT56818


pedrobaeza:
Fast-tracking for solving this bug that has been around a lot of time with the bugfix already provided:

/ocabot merge patch

PR #1381, opened 2025-09-01
[18.0][IMP] mis_template_financial_report: unallocated earnings

jguenat:
The account type equity_unaffected (Current Year Earnings)  is currently used on the P&L report which makes no sense IMO.

After reading [the official documentation](https://www.odoo.com/documentation/18.0/applications/finance/accounting/reporting/year_end.html#current-year-s-earnings) and some feedback from accounting colleagues I propose to put the equity_unaffected on the balance sheet.

We could go further and implement "Retained earnings" in the bottom of the balance sheet instead of showing the Profit / Loss there.

### DIFF
--- a/mis_template_financial_report/data/mis_report_kpi.xml
+++ b/mis_template_financial_report/data/mis_report_kpi.xml
@@ -1,17 +1,5 @@
 <odoo>
-    <record id="kpi_loss" model="mis.report.kpi">
-        <field name="name">loss</field>
-        <field name="description">Loss</field>
-        <field name="expression">balp[('account_type', 'like', 'expense%')][]</field>
-        <field name="auto_expand_accounts">true</field>
-        <field name="auto_expand_accounts_style_id" ref="style_details" />
-        <field name="style_id" ref="style_header" />
-        <field name="type">num</field>
-        <field name="compare_method">diff</field>
-        <field name="accumulation_method">sum</field>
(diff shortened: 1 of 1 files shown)


mis_template_financial_report/data/mis_report_kpi.xml
```diff
@@ -1,35 +1,33 @@
 <odoo>
-    <record id="kpi_loss" model="mis.report.kpi">
-        <field name="name">loss</field>
-        <field name="description">Loss</field>
-        <field name="expression">balp[('account_type', 'like', 'expense%')][]</field>
+    <!-- profit & loss -->
+    <record id="kpi_profit" model="mis.report.kpi">
```
pedrobaeza, line 3: Please reduce the diff to the minimum and don't reorder elements.

pedrobaeza: Or do it in a different commit with `[REF]` tag.

jguenat: ok, it was difficult to read the file as it was ordered strangly, is it okay for you if I do two commits?
One redordering and one with my kpi modifications?

hbrunn (approved):
thanks, I agree the reordering makes it more readable

pedrobaeza (approved):
/ocabot merge major

PR #1387, opened 2025-09-12
[18.0][FIX] account_financial_report: correctly sum amount_currency in read_group for trial balance

### LINKED ISSUE #1388: [18.0] account_financial_report: amount_currency not correctly returned for ending_balance
## Module
account_financial_report

## Describe the bug
If you try to export trial balance with some count in other currency, the value on ending balance is returned as 0.

## To Reproduce
**Affected versions**:

1. Set an account with foreign currency
2. Add a move with this count
3. Export trial balance
The ending balance in currency is 0

**Expected behavior**
The ending balance should be correct

- [ ]  18.0 #1387

micheledic:
### After this PR

On Odoo18 if you group:sum by amount_currency, if you don't put currency_id in the group field the amount_currency is removed by read_group function on account addon

Fixes: https://github.com/OCA/account-financial-reporting/issues/1388

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -194,7 +194,7 @@ class TrialBalanceReport(models.AbstractModel):
         initial_balances = self.env["account.move.line"].read_group(
             domain=domain,
             fields=["account_id", "balance", "amount_currency:sum"],
-            groupby=["account_id"],
+            groupby=["account_id", "currency_id"],
         )
         pl_initial_balance = 0.0
         pl_initial_currency_balance = 0.0
@@ -432,7 +432,7 @@ class TrialBalanceReport(models.AbstractModel):
             tb_initial_acc.append(
                 {"account_id": account.id, "balance": 0.0, "amount_currency": 0.0}
             )
-        groupby_fields = ["account_id"]
+        groupby_fields = ["account_id", "currency_id"]
         if grouped_by:
             groupby_fields.append("analytic_account_ids")
         initial_domain_bs = self._get_initial_balances_bs_ml_domain(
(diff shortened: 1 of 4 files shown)


micheledic:
@pedrobaeza another day, another pr!

pedrobaeza:
I don't quite get the problem. One account should be only in one currency, not in several...

micheledic:
> I don't quite get the problem. One account should be only in one currency, not in several...

@pedrobaeza 
https://github.com/odoo/odoo/blob/6fb3a58498841c171b4ca337f2c2e173c12c9f28/addons/account/models/account_move_line.py#L3485
Check this function, without the currency_id, the amount_currency will be empty. The only way is to pass the currency_id.
If you don't pass the currency_id, the amount_currency will be 0

pedrobaeza (approved):
OK, thanks for the explanation.

account_financial_report/report/trial_balance.py
monen17, line 1: Is the same change needed also in https://github.com/OCA/account-financial-reporting/blob/51b207773da98090b9a3d40d9b53318da9dd8680/account_financial_report/report/trial_balance.py#L223 and https://github.com/OCA/account-financial-reporting/blob/51b207773da98090b9a3d40d9b53318da9dd8680/account_financial_report/report/trial_balance.py#L480?

monen17 (approved):
Thanks for the PR!
Only code review.

Since it is a fix, a test would be needed:
> As a general rule, a bug fix should come with a unittest which would fail without the fix itself. This is to assure that regression will not happen in the future. It also is a good way to show that the fix works in all cases.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#6tests)
but I understand testing a report is not easy to implement and changes are minimal, so I think it's not mandatory for merge in this case.

Please consider to do the same changes in General Ledger that is probably affected too.

account_financial_report/report/trial_balance.py
micheledic: It Is not neeeded because  this groupby comes from the master read_group.
So the read_group group the data and for each line returns the groupby and other data that Will be used on this function

monen17: So we are extracting the `groupby` from the results of previous `read_group` that have already been edited, and those previous `groupby`s are injected in the `__context` thanks to https://github.com/odoo/odoo/blob/86fee5323d8572a54255c15b7e606be9c51d6119/odoo/models.py#L2920.
Thanks for the clarification :smile:
:+1:

micheledic:
@pedrobaeza can be merged ?

PR #1389, opened 2025-09-19
[16.0] [FIX] account_financial_report: use company currency from company report wizard not from currently active company

giarve:
The currency should be the main company currency for this report instead of the currently enabled company. Otherwise, values are incorrect because currency conversion is never performed.

must be fw ported to 17.0 and 18.0

### DIFF
--- a/account_financial_report/report/templates/aged_partner_balance.xml
+++ b/account_financial_report/report/templates/aged_partner_balance.xml
@@ -146,14 +146,14 @@
             <div class="act_as_cell amount">
                 <span
                     t-esc="partner['residual']"
-                    t-options="{'widget': 'monetary', 'display_currency': res_company.currency_id}"
+                    t-options="{'widget': 'monetary', 'display_currency': o.company_id.currency_id}"
                 />
             </div>
                 <!--## current-->
             <div class="act_as_cell amount">
                     <span
                     t-esc="partner['current']"
-                    t-options="{'widget': 'monetary', 'display_currency': res_company.currency_id}"
+                    t-options="{'widget': 'monetary', 'display_currency': o.company_id.currency_id}"
                 />
             </div>
             <t t-if="not age_partner_config">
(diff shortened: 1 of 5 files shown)


leemannd (approved):

pedrobaeza:
If the problem is that `res_company` is not pointing to the proper company (the active one vs the wizard one), shouldn't you change the value of `res_company` instead?

giarve:
> If the problem is that `res_company` is not pointing to the proper company (the active one vs the wizard one), shouldn't you change the value of `res_company` instead?

Should res_company be pointing to the wizard one this time or always the active company?

You mention it because the report could be generated without the wizard at some point right?

To change it is it only required to to a t-set from the wizard or to modify the data getter for the report?

pedrobaeza:
There's no sense `res_company` to be the active company. For what? You want to have the "report company".

marcos-mendez (changes requested):
**Automated Review -- Tests Failed**
### 1. **Root Cause**

The test failure occurs because the `o.company_id.currency_id` reference in the XML templates is not properly resolved during test execution, likely due to missing or incorrect initialization of the `o` variable (which should refer to the report object) in the test environment. This causes a failure in loading the module registry.

### 2. **Suggested Fix**

The change from `res_company.currency_id` to `o.company_id.currency_id` is correct for multi-company support, but it must be ensured that `o` is correctly defined in the template context. Specifically, in the affected XML templates (e.g., `aged_partner_balance.xml`, `general_ledger.xml`, etc.), verify that `o` is properly bound to the report object. If `o` is not defined, it should be explicitly set in the report's `render` method or template context using a `t-set` directive.

**Example fix in `aged_partner_balance.xml` (around line 149):**
```xml
<t t-set="o" t-value="self"/>
```

### 3. **Additional Code Issues**

- **Missing `t-set="o"` in templates**: If `o` is used in templates but not defined, it will cause runtime errors. All templates using `o.company_id` should ensure `o` is set to the report object.
- **Inconsistent use of `res_company` vs `o.company_id`**: The change is good for multi-company support, but it must be consistent across all templates and should not break backward compatibility unless intended.

### 4. **Test Improvements**

Add tests using `SavepointCase` or `TransactionCase` to cover multi-company scenarios:
- Test that `o.company_id.currency_id` correctly resolves to the report's company currency.
- Add test cases with multiple companies and ensure the currency is correctly displayed.
- Ensure that reports render correctly when `company_id` is set on the report object.

**Example test pattern (OCA style):**
```python
def test_aged_partner_balance_multicompany(self):
    # Create a second company
    company2 = self.env['res.company'].create({
        'name': 'Test Company 2',
        'currency_id': self.env.ref('base.EUR').id,
    })
    # Set up report with company2
    report = self.env['report.account_financial_report.aged_partner_balance'].create({
        'company_id': company2.id,
    })
    # Render and check currency
    result = report.render_html()
    self.assertIn('EUR', result)
```

This ensures that the `o.company_id` logic is tested in a realistic multi-company environment.
---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

PR #1320, opened 2025-05-21
[18.0][MIG] mis_builder_cash_flow: Migration to 18.0

geraldo29:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/account-financial-reporting/issues/1230

Followed technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0.

Changes:
1)Removed migration folder
2)Changed 17 to 18 version in a few places
3)Changed tree to list in a few places.
4)Added test-requirements.txt

PR #1399, opened 2025-10-08
[18.0][MIG] mis_builder_cash_flow

etobella:
Replaces https://github.com/OCA/account-financial-reporting/pull/1320/


luisDIXMIT (approved):
LGTM!

mis_builder_cash_flow/report/mis_cash_flow.py
```diff
@@ -59,7 +59,7 @@ def init(self):
             SELECT
                 -- we use negative id to avoid duplicates and we don't use
                 -- ROW_NUMBER() because the performance was very poor
-                -aml.id as id,
+                2*aml.id as id,
```
pedrobaeza: This is not correct, as there may be duplicates. Which is the problem with the negative? If not possible to be put, use a hash function like in https://github.com/OCA/hr-attendance/blob/27d69ba2ef5ab141fc5243d5baa9f4eedd199e63/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py#L54

etobella: From a mathematics perspective the negative and 2*n+1 / 2*n have the same result (and faster than the hash). I don't know why it was removed the negative, but we assumed that it gave some issues, for this reason we applied this logic.

Actually, by reference, it is the second example of the infinite hotel paradox of Hilbert:

https://en.wikipedia.org/wiki/Hilbert%27s_paradox_of_the_Grand_Hotel

pedrobaeza: But this is only multiplying by 2 the id, not doing other things, so yes, this is going to collapse for sure.

etobella: https://github.com/OCA/account-financial-reporting/pull/1399/commits/14f30002246e70b052071f757d7940060d2b297a#diff-5b5b3d9a47ac64faf8345abac475416b9a82535dd7f74de843483b96ffd9a72cL87

pedrobaeza: OK, understood the technique, sorry for not seeing that part. But the risk now is of overflowing the limits of the ID (int4), reducing the possible IDs to the half, and specially in this table, which can be conflicting, can we check why negative is not valid?

pedrobaeza: ping @etobella

luisDIXMIT: It seems to me that negative ids are valid because no errors were raised.
image

rvalyi: FYI Odoo is still using the negative id trick https://github.com/odoo/odoo/blob/master/addons/purchase/models/purchase_bill_line_match.py#L107 (though this bill matching thing has other unrelated issues)
I think the negative ids thing works unless proven otherwise.

pedrobaeza: Thanks both for the confirmation. Please @etobella @luisDIXMIT restore this, and it's good to go.

luisDIXMIT: Done!

mis_builder_cash_flow/report/mis_cash_flow.py
```diff
@@ -84,7 +84,7 @@ def init(self):
             WHERE aml.parent_state != 'cancel'
             UNION ALL
             SELECT
-                fl.id + 1000000 as id,
+                2*fl.id + 1 as id,
```
pedrobaeza: I still see the `2*fl.id` part...

luisDIXMIT: I fixed the negative issue, but isn’t the 2*id+1 approach better than delaying the conflict 100k?

etobella: One should be positive and the other negative, no other arithmetic functions are needed.

luisDIXMIT: I already see what you’re saying. I hope it’s correct now.

etobella: Thanks @luisDIXMIT Now it looks ok

pedrobaeza (approved):
/ocabot merge nobump

PR #1403, opened 2025-10-22
[18.0][FIX] mis_builder_cash_flow: Edit KPI expression and fix posted filter

ThiagoMForgeFlow:
Change the state field to parent_state of the mis.cash_flow model to compute correctly posted filter added by the mis_builder module. 

Method that allows adding the posted filter in MIS reports: 
```
    @api.model
    def _supports_target_move_filter(self, aml_model_name):
        return "parent_state" in self.env[aml_model_name]._fields
```

In migration to 16, KPI expressions were modified to adapt the account_internal_type field to account_type. However, in this process, in liquidity and balance KPIs, the liability_credit_card type was forgotten. This account_type in v15 had account_internal_type equal to liquidity.

### DIFF
--- a/mis_builder_cash_flow/data/mis_report.xml
+++ b/mis_builder_cash_flow/data/mis_report.xml
@@ -16,7 +16,7 @@
         <field name="sequence">20</field>
         <field
             name="expression"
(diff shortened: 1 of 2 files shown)


ThiagoMForgeFlow:
Hi @etobella what do you think about these changes?

etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1404, opened 2025-10-23
[18.0][FIX] partner_statement: Filter by company

etobella:
if you have 2 localizations installed, the test fails. This happens because the code_store is not filled and then the code is empty.

### DIFF
--- a/partner_statement/tests/test_outstanding_statement.py
+++ b/partner_statement/tests/test_outstanding_statement.py
@@ -136,6 +136,7 @@ class TestOutstandingStatement(TransactionCase):
             [
                 ("id", "!=", copy_account.id),
                 ("account_type", "=", wizard.account_type),
+                ("company_ids", "in", self.env.company.ids),
             ],
         )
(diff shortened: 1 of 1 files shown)


partner_statement/tests/test_outstanding_statement.py
```diff
@@ -136,6 +136,7 @@ def test_exclude_accounts(self):
             [
                 ("id", "!=", copy_account.id),
                 ("account_type", "=", wizard.account_type),
+                ("company_ids", "=", self.env.company.id),
```
MiquelRForgeFlow:
```suggestion
                ("company_ids", "in", self.env.company.ids),
```

luisDIXMIT: I suggest the same.

luisDIXMIT (approved):
LGTM. Minor suggestion.

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

PR #1407, opened 2025-11-03
[16.0][FIX] account_financial_report: Taxes option missed in journal ledger

pedrobaeza:
Forward-port of #1339 

A Typo error causes the oden not to be as expected when searching for multiple reports with the same report_name (which may not be ideal either) and causes the configuration set in the paper format not to be applied. Update translation terms to avoid the above issue.

@Tecnativa TT56812


carlosdauden (approved):

pedrobaeza:
Let's include this fix:

/ocabot merge patch

PR #1408, opened 2025-11-04
[18.0][IMP] account_financial_report: Allow user cut label string of account move line in reports

pedrobaeza:
Partial forward-port of #1281 and #1290, making it more compatible, and direct fw-port of #1291.

@Tecnativa

PR #1409, opened 2025-11-07
[18.0][FIX] account_financial_report: default set with sudo

pedrobaeza:
Follow-up of #1408 

If not, most users will have an access error.

@Tecnativa


pedrobaeza:
Fast-tracking for solving the issue:

/ocabot merge patch

PR #1410, opened 2025-11-07
[17.0][FIX] account_financial_report: default set with sudo

pedrobaeza:
If not, most users will have an access error.

@Tecnativa


pedrobaeza:
Another one fast-tracking for solving the issue:

/ocabot merge patch

PR #1415, opened 2025-11-22
[18.0][FIX] account_financial_report: display results considering archived accounts and journals

JordiBForgeFlow:
Archiving a journal or an account should not result in not being able to displaying the financial results related with them. For example, consider a sales journal that you don't want to use going forward. You will still want to see financial activity related with that journal for past periods.

### DIFF
--- a/account_financial_report/report/journal_ledger_xlsx.py
+++ b/account_financial_report/report/journal_ledger_xlsx.py
@@ -178,7 +178,11 @@ class JournalLedgerXslx(models.AbstractModel):
     def _generate_journal_content(
         self, workbook, report, res_data, ledger, report_data
     ):
-        journal = self.env["account.journal"].browse(ledger["id"])
+        journal = (
+            self.env["account.journal"]
+            .with_context(active_test=False)
+            .browse(ledger["id"])
+        )
         currency_name = (
             journal.currency_id
             and journal.currency_id.name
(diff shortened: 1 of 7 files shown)


account_financial_report/report/journal_ledger_xlsx.py
```diff
@@ -178,7 +178,11 @@ def _generate_no_group_content(self, workbook, report, res_data, report_data):
     def _generate_journal_content(
         self, workbook, report, res_data, ledger, report_data
     ):
-        journal = self.env["account.journal"].browse(ledger["id"])
+        journal = (
```
pedrobaeza, line 181: This one is loading a journal, so are you sure you require the context here?

account_financial_report/report/journal_ledger_xlsx.py
```diff
@@ -196,7 +200,11 @@ def _generate_no_group_taxes_summary(self, workbook, report, res_data, report_da
         )
 
     def _generate_journal_taxes_summary(self, workbook, ledger, report_data):
-        journal = self.env["account.journal"].browse(ledger["id"])
+        journal = (
```
pedrobaeza, line 203: The same for this one.

pedrobaeza (approved):
Thanks for bringing this to table. It's true that we are not thinking of that.

FernandoRomera:
@pedrobaeza , @JordiBForgeFlow 
There is a partial PR for this and it is merged.
https://github.com/OCA/account-financial-reporting/pull/1417/files
We can move forward with this one or better yet, duplicate it, removing what you did in 1417.

pedrobaeza:
Oh, sorry, I didn't remember this one and go for the specific bug I found!

Let's merge this one as well, as it seems there's no conflict.

/ocabot merge patch

PR #1416, opened 2025-11-25
[18.0][FIX] account_financial_report: Ability to filter by code

david-s73:
- The way accounts are collected has been modified, as they were returning ‘NewId’, which meant that filtering was not possible at any point. I check whether the first position is already a ‘NewId’ and, if so, I browse the IDs so that ‘self.account_ids’ is filled in correctly and filtered correctly.

- The example that it is not filtering is taken from a runboat. As you can see in the screenshots, it does not filter and shows all accounts.

image
image

- Here is an example of how it should look once the filter is working correctly. As you can see in the screenshot, after applying the filter, you can see which accounts will be displayed.

image


Reyes4711-S73 (approved):
LGTM

Raul-S73 (approved):

Reyes4711-S73:
@OCA/accounting-maintainers Please, can you review/merge this PR?

gurneyalex (approved):

PR #1418, opened 2025-12-10
[18.0][FIX] account_financial_report: journal ledger move_line_ids_taxes_data

qgroulard:
In the journal ledger report, only move lines with `tax_line_id` had their taxes displayed.
The case where taxes come from `move_line_ids_taxes_data` seems to have been forgotten.

We solve two problems.
First, `move_line_ids_taxes_data` was never propagated, thus always empty.
Second, tax name and description get from sql queries are json values. We now handle translations.

### DIFF
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -178,7 +178,7 @@ class JournalLedgerReport(models.AbstractModel):
             SELECT aml_at_rel.account_move_line_id, aml_at_rel.account_tax_id,
             at.description, at.name
             FROM account_move_line_account_tax_rel AS aml_at_rel
-            LEFT JOIN
+            INNER JOIN
                 account_tax AS at on (at.id = aml_at_rel.account_tax_id)
             WHERE account_move_line_id IN %(move_line_ids)s
         """
@@ -207,6 +207,7 @@ class JournalLedgerReport(models.AbstractModel):
         )
         move_line_ids_taxes_data = {}
         if move_lines:
+            lang = self.env.context.get("lang", "en_US")
             # Get the taxes ids for the move lines
             query_taxes_params = self._get_query_taxes_params(move_lines)
             query_taxes = self._get_query_taxes()
@@ -221,8 +222,12 @@ class JournalLedgerReport(models.AbstractModel):
                 if move_line_id not in move_line_ids_taxes_data.keys():
                     move_line_ids_taxes_data[move_line_id] = {}
                 move_line_ids_taxes_data[move_line_id][account_tax_id] = {
-                    "name": tax_name,
-                    "description": tax_description,
+                    "name": tax_name.get(lang) or tax_name.get("en_US"),
+                    "description": (
+                        tax_description.get(lang) or tax_description.get("en_US")
+                        if tax_description
+                        else False
+                    ),
                 }
         Move_Lines = {}
         auto_sequence = len(move_ids)
@@ -331,6 +336,7 @@ class JournalLedgerReport(models.AbstractModel):
             partner_ids_data = move_lines[3]
             currency_ids_data = move_lines[4]
             tax_line_ids_data = move_lines[5]
+            move_line_ids_taxes_data = move_lines[6]
         for move_data in moves_data:
             move_id = move_data["move_id"]
             move_data["report_move_lines"] = []
--- a/account_financial_report/report/journal_ledger.py
+++ b/account_financial_report/report/journal_ledger.py
@@ -203,6 +203,7 @@ class JournalLedgerReport(models.AbstractModel):
         )
         move_line_ids_taxes_data = {}
         if move_lines:
+            lang = self.env.context.get("lang", "en_US")
             # Get the taxes ids for the move lines
             query_taxes_params = self._get_query_taxes_params(move_lines)
             query_taxes = self._get_query_taxes()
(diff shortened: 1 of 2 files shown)


flotho:
ping @DorianMAG

flotho (changes requested):
Hi,

Thanks for this usefull PR.
Code could be optimized.
You defined L206 the lang variable. Its value is either en_US or another one, never None AFAICS.
Though, you don't need anymore to switch case on lang in L221 and L223
Could you change this ?

Regards

qgroulard:
> Hi,
> 
> Thanks for this usefull PR. Code could be optimized. You defined L206 the lang variable. Its value is either en_US or another one, never None AFAICS. Though, you don't need anymore to switch case on lang in L221 and L223 Could you change this ?
> 
> Regards

Hi, thanks for your review.

The thing is you don't necessarily have a translation for your current language.
If you look in DB, even though a language is installed, translatable fields will have no other key than "en_US" until they are actually translated once.

account_financial_report/report/journal_ledger.py
```diff
@@ -217,8 +218,12 @@ def _get_move_lines(self, move_ids, wizard, journal_ids):
                 if move_line_id not in move_line_ids_taxes_data.keys():
                     move_line_ids_taxes_data[move_line_id] = {}
                 move_line_ids_taxes_data[move_line_id][account_tax_id] = {
-                    "name": tax_name,
-                    "description": tax_description,
+                    "name": tax_name.get(lang) or tax_name.get("en_US"),
```
flotho, line 221: > > Hi,
> > Thanks for this usefull PR. Code could be optimized. You defined L206 the lang variable. Its value is either en_US or another one, never None AFAICS. Though, you don't need anymore to switch case on lang in L221 and L223 Could you change this ?
> > Regards
> 
> Hi, thanks for your review.
> 
> The thing is you don't necessarily have a translation for your current language. If you look in DB, even though a language is installed, translatable fields will have no other key than "en_US" until they are actually translated once.

Thanks for the answer.
I thought I've understood what you want to do.
What I'm proposing is to change this  :
` "name": tax_name.get(lang) or tax_name.get("en_US"), `
By 
` "name": tax_name.get(lang),`
You can do this because you defined a lang variable with a default value that will be en_US. So the test you made is redundant regarding the possible values in the lang variable.
Am I missing something?
regards

flotho (changes requested):

account_financial_report/report/journal_ledger.py
qgroulard: Let's say I have French and English installed on my DB and my current language is French.

- lang = "fr_FR"
- tax_name might be: {"en_US": "Tax 21%"}

Hence, `tax_name.get(lang)` will return `False`, which will eventually end up as an empty column in the report.

What I have implemented is a fallback on English translation. This is similar to what the ORM does, if you open the form view of the tax the name won't be empty if you have no French translation, the English name will be displayed instead.

flotho (approved):
Thanks for your clear answer.

I missed this one that data in the json could be "de"synchronized from installed languages.

So evrything is OK
LGTM

flotho:
Hi, 
Any chance to rebase to trigger CI/CD ?

qgroulard:
> Hi, Any chance to rebase to trigger CI/CD ?

Done :)

flotho (approved):
thanks for the update

flotho:
ping @DorianMAG

DorianMAG (approved):
Tested on ruboat.
LGTM
Thx for this fix

flotho:
ping @OCA/accounting-maintainers any chance to merge ?

qgroulard:
Hi @OCA/accounting-maintainers, could we avoid this PR to die ? :fearful:

ryanduguid (commented):
Thanks for this, and for keeping at it through the stale bot.

Code review only, I have not run it.

The two halves look separable and are not, which is worth recording for whoever
merges. `move_line_ids_taxes_data` is consumed in `journal_ledger_xlsx.py` at line
252 and in `templates/journal_ledger.xml` at line 279, and both take it from
`_get_report_values`. Until this PR that local was still the `{}` from the chained
initialisation, because nothing read `move_lines[6]`, even though `_get_move_lines`
has returned it all along. So the `elif move_line_taxes_data:` branch of
`_get_ml_tax_description` never ran, which is the symptom you describe.

Propagating it activates that branch, and only then do the values matter. In 18.0
`account.tax.name` is `Char(translate=True)` and `description` is
`Html(translate=html_translate)`, so both columns are jsonb and `_get_query_taxes`
hands back dicts. Without the unpacking I would expect `",".join(tax_names)` to raise
a TypeError on the first line that carries taxes. That is from reading rather than a
run, but if it holds, the propagation and the translation handling have to land
together and splitting the PR would ship a crash.

It also answers the earlier question about the `or ... get("en_US")` fallback. Your
reply was right. Odoo writes only the `en_US` key into the jsonb until a value is
actually translated, so `.get(lang)` on its own returns `None` for any installed but
untranslated language and the taxes column would quietly go blank. Worth keeping.

Two small things, neither a blocker:

1. `tax_description` is guarded with `if tax_description else False` but `tax_name`
   is not. `name` is required so it cannot be null on the tax itself, but
   `_get_query_taxes` uses a LEFT JOIN, so an orphan row in
   `account_move_line_account_tax_rel` would give `None` and `tax_name.get(lang)`
   would raise. If the LEFT JOIN is deliberate then the same guard on name would be
   consistent; if it is not, an INNER JOIN would say what you mean.

2. Would you add a regression assertion? `test_03_test_taxes_in_invoice` already
   builds the case, two base lines carrying `tax_15_p` and `tax_20_p` with no
   `tax_line_id`, so it needs only a check that the taxes description for those lines
   is a non-empty string. That would let a maintainer merge without a functional run.

Disclosure: I have #1567 open against the same two branches of
`_get_ml_tax_description`, so whichever lands second needs a trivial rebase. Yours is
older and already approved, so I would rather it went first.

@OCA/accounting-maintainers this has two approvals and the ready to merge label since
30 April. Could it get the no stale label so it does not close in three weeks?

qgroulard:
Thank you @ryanduguid for the review, good catch!
I believe the INNER JOIN makes more sense.
I have pushed a new version.

PR #1422, opened 2025-12-17
[18.0][ADD] account_move_zip_download

remi-filament:
This module is a simple backport of Odoo https://github.com/odoo/odoo/pull/219245 and https://github.com/odoo/odoo/pull/238660
(deployed on saas-18.4)

This functionality was already available in Odoo version 18.0 with Print entry named "PDF", however the behaviour was changed on 2025/12/08 when merging https://github.com/odoo/odoo/pull/236833 (PDF report replaces the "PDF" entry which does not allow for zip file download anymore).

### DIFF
--- a/account_move_zip_download/LICENSE
+++ b/account_move_zip_download/LICENSE
@@ -1,0 +1,845 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+
+  This version of the GNU Lesser General Public License incorporates
+the terms and conditions of version 3 of the GNU General Public
+License, supplemented by the additional permissions listed below.
+
+  0. Additional Definitions.
+
+  As used herein, "this License" refers to version 3 of the GNU Lesser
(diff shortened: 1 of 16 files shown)


yostashiro:
I don't see the Export ZIP option in runboat. Am I missing something?

image

In the meantime, https://github.com/OCA/reporting-engine/tree/18.0/report_pdf_zip_download should cover the functionality, although the usability is slightly different (there is no additional Export ZIP button).

remi-filament:
Hi @yostashiro,
Thanks for looking into it, the invoices need to be sent to see the button.
I agree that it is quite similar to what you can get from report_pdf_zip_download, although I see 2 advantages for this backport : 
1. It is a backport of how it works on saas-18.4+ so using it in v18 you will get the same thing after migration to upper version
2. The main difference with the other module is that it should reuse stored PDF files (which is why you need to send the invoices so that PDF is stored in Odoo) when report_pdf_zip_download would regenerate the PDF (and therefore use lots of resources for wkhtmltopdf)

coco2299:
Just for your information, the PR: https://github.com/odoo/odoo/pull/236833
As been revert because it was breaking the stable: https://github.com/odoo/odoo/pull/241846

I see tickets about that PR, maybe the revert doesn't work as expected : https://www.odoo.com/odoo/project/49/tasks/5422235

The PO confirmed that the behavior in 18.0 was expected in the meanwhile

PR #1425, opened 2025-12-28
[17.0][FIX] mis_template_financial_report: make compatible with mis_builder updates

hbrunn:
Fixes
TypeError: can't access property "2##1#", ctx.notes is undefined

since https://github.com/OCA/mis-builder/commit/a08dfa72af1986ac4fc7eec2c8ac92f0ed78627c we can't get away with breaking inheritance


pedrobaeza (approved):
Let's fast-track it for fixing the incompatibility:

/ocabot merge patch

PR #1426, opened 2026-01-08
[IMP] partner_statement: add Lithuanian translation

psugne:
[IMP] partner_statement: add Lithuanian translation


MiquelRForgeFlow (approved):
I don't know Lithuanian, but I believe everything is fine.

PR #1434, opened 2026-01-15
[18.0][FIX] account_financial_report: account selection by interval when handling NewId records

Ricardoalso:
Fixing some unfixed `on_change_account_range` methods following the same approach as the ones done in https://github.com/OCA/account-financial-reporting/pull/1416 and https://github.com/OCA/account-financial-reporting/pull/1293 

Useful explanation from the comments 

> Q: it is not clear to me why we no longer have to handle the NewId case here. Could you enlighten me?
> 
> A: Because this is an onchange method on an unsaved Transient Model record. Odoo wraps the assigned records with NewId objects. By calling the filtered on those records of NewId type, it fails.
And now we are only calling the search method once with the adapted domain, improving overall performances also.

### DIFF
--- a/account_financial_report/wizard/aged_partner_balance_wizard.py
+++ b/account_financial_report/wizard/aged_partner_balance_wizard.py
@@ -54,17 +54,14 @@ class AgedPartnerBalanceWizard(models.TransientModel):
         ):
             start_range = int(self.account_code_from.code)
             end_range = int(self.account_code_to.code)
-            self.account_ids = self.env["account.account"].search(
-                [
-                    ("code", ">=", start_range),
-                    ("code", "<=", end_range),
-                    ("reconcile", "=", True),
-                ]
-            )
+            domain = [
+                ("code", ">=", start_range),
+                ("code", "<=", end_range),
+                ("reconcile", "=", True),
+            ]
             if self.company_id:
-                self.account_ids = self.account_ids.filtered(
-                    lambda a: self.company_id in a.company_ids
-                )
+                domain.append(("company_ids", "=", self.company_id.id))
+            self.account_ids = self.env["account.account"].search(domain)
         return {
             "domain": {
                 "account_code_from": [("reconcile", "=", True)],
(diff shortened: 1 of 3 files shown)


account_financial_report/wizard/trial_balance_wizard.py
```diff
@@ -88,21 +88,12 @@ def on_change_account_range(self):
...
-                if self.company_id:
-                    self.account_ids = self.account_ids.filtered(
-                        lambda a: self.company_id in a.company_ids
-                    )
+            domain = [("code", ">=", start_range), ("code", "<=", end_range)]
+
+            if self.company_id:
+                domain.append(("company_ids", "in", [self.company_id.id]))
+
+            self.account_ids = self.env["account.account"].search(domain)
```
gurneyalex, line 94: it is not clear to me why we no longer have to handle the NewId case here. Could you enlighten me?

Ricardoalso: Because this is an `onchange` method on an unsaved Transient Model record. Odoo wraps the assigned records with NewId objects. By calling the filtered on those records of NewId type, it fails.

And now we are only calling the search method once with the adapted domain, improving overall performances also.

gurneyalex (approved):

pedrobaeza (commented):
Can you please remove the empty lines inside the methods?

Ricardoalso:
> Can you please remove the empty lines inside the methods?

@pedrobaeza :heavy_check_mark:

account_financial_report/wizard/aged_partner_balance_wizard.py
```diff
@@ -54,17 +54,14 @@ def on_change_account_range(self):
...
+            domain = [
+                ("code", ">=", start_range),
+                ("code", "<=", end_range),
+                ("reconcile", "=", True),
+            ]
             if self.company_id:
-                self.account_ids = self.account_ids.filtered(
-                    lambda a: self.company_id in a.company_ids
-                )
+                domain.append(("company_ids", "in", [self.company_id.id]))
```
pedrobaeza:
```suggestion
                domain.append(("company_ids", "=", self.company_id.id))
```

pedrobaeza (commented):
And the same for the rest

pedrobaeza (approved):

Ricardoalso:
@gurneyalex Can you trigger a merge on this one ? :pray:

pedrobaeza:
/ocabot merge patch

Please fw-port it to 19.

PR #1439, opened 2026-01-30
[18.0][FIX]account_financial_report - fix General Legder stack trace

luc-demeyer:
fix General Legder stack trace on archived analytic accounts


pedrobaeza (approved):
In the same mood than c8ac1b651d1b208d41579f962c2ba966f0e07b00, so let's merge it.

/ocabot merge patch

I think there may be other records with the same problem (partners maybe?)

PR #1417, opened 2025-12-10
[18.0][FIX] account_financial_report: Don't fail on inactive journal

pedrobaeza:
Steps to reproduce:

- Archive a journal that contains journal entries.
- Print a general ledger including at least one of those journal entries.
- You get a KeyError:

  ```
  File ".../addons/account_financial_report/models/ir_actions_report.py", line 27, in _render_xlsx
        return super(IrActionsReport, obj)._render_xlsx(report_ref, docids, data=data)
  File ".../addons/report_xlsx_helper/models/ir_actions_report.py", line 19, in _render_xlsx
    return super()._render_xlsx(report_ref, docids, data)
  File ".../addons/report_xlsx/models/ir_report.py", line 27, in _render_xlsx
    .create_xlsx_report(docids, data)  # noqa
  File ".../addons/report_xlsx/report/report_abstract_xlsx.py", line 105, in create_xlsx_report
    self.generate_xlsx_report(workbook, data, objs)
  File ".../addons/account_financial_report/report/abstract_report_xlsx.py", line 40, in generate_xlsx_report
    self._generate_report_content(workbook, objects, data, report_data)
  File ".../addons/account_financial_report/report/general_ledger_xlsx.py", line 287, in _generate_report_content
    "journal": journals_data[line["journal_id"]]["code"],
  KeyError: 93
  ```

That's because the code of the journals are fetched with a search without active_test=False context, so archived journals are not taken into account, and when later trying to get the information, failing.

The solution is to pass the context for getting all the journals, including archived ones.

TT59739

PR #1443, opened 2026-02-11
[15.0][FIX] account_financial_report: Archived journals in journal ledger

pedrobaeza:
If there are archived journals containing entries, and the journal ledger is issued without selecting any journal (which means, all the journals), it will be generated without taking into account that entries.

This commit makes the journal search where found with the context to take into account archived journals.

It's a pity that there's no global context in the abstract report that can be used for this for preventing similar errors, but the current architecture doesn't allow it, so let's go for the local patch.

Similar to #1417 and #1439.

@Tecnativa TT60917


pedrobaeza:
Important patch for having consistent information, so fast-tracking it:

/ocabot merge patch

PR #1455, opened 2026-03-03
[16.0]Update copier template to 1.39

pedrobaeza:
@Tecnativa


pedrobaeza:
The rest of the fix in #1456

PR #1476, opened 2026-03-13
[16.0][FIX] account_financial_report: If partner_id is 0, set the domain to ['partner_id','=', False]

Andrii9090-tecnativa:
@pedrobaeza can you review please?

@Tecnativa


pedrobaeza (commented):
The same problem as #1466, so fast-tracking it:

/ocabot merge patch

pedrobaeza (approved):

PR #1483, opened 2026-03-16
[16.0][IMP] account_finacial_report: Add the option to group by partner for accounts (account.account) where reconcile=True

Andrii9090-tecnativa:
Allow grouping by partner in General Ledger for accounts with `reconcile=True`.

@pedrobaeza can you review?

@Tecnativa

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -50,13 +50,16 @@ class GeneralLedgerReport(models.AbstractModel):
             )
         return taxes_data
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
As `asset_receivable` and `liability_payable` has both `reconcile=True`, I think you can simply switch the domain to this last one.

pedrobaeza (approved):

eduezerouali-tecnativa (approved):
Tested on runboat

PR #1487, opened 2026-03-20
[18.0][IMP] account_financial_report: use precomputed account value in aged partner balance

BhaveshHeliconia:
### Aged Partner Balance
- **Use `display_name` for account:** Replaced `line['line_rec'].account_id.display_name` with the precomputed `line['account']` value in the report template for consistency and better performance.


BhaveshHeliconia:
@victoralmau @pedrobaeza could you please review this when you get a chance? Thank you!

pedrobaeza:
I think each improvement should be in a different PR for discussing it independently

PR #1488, opened 2026-03-20
[18.0][IMP] account_financial_report: add total row in trial balance PDF report

BhaveshHeliconia:
- Added a totals footer row to the trial balance PDF/HTML/XLSX  report template, showing the sum of initial balance, debit, credit, period balance, and ending balance columns,


pedrobaeza (commented):
Thanks for the improvement. It's OK, but the text is not at all, as previously, the totals were not exported in XLSX as well.

BhaveshHeliconia:
> Thanks for the improvement. It's OK, but the text is not at all, as previously, the totals were not exported in XLSX as well.

Thanks for the review! You're right — I've corrected the description.

PR #1490, opened 2026-03-25
[18.0][FIX] account_financial_report: set company context for multi-company account code resolution

dnplkndll:
## Summary

Set the correct company context when resolving `account.code` in financial report generation, fixing `TypeError` in multi-company setups.

## Problem

In Odoo 18, `account.account.code` is a computed field backed by `code_store` (jsonb keyed by company ID). The `_get_accounts_data()` method in `abstract_report.py` browses accounts without setting the report's target company in the ORM environment. When the user's default company differs from the report's company, `account.code` returns `False` instead of a string, causing:

```
TypeError: unsupported operand type(s) for +: 'bool' and 'str'
```

### Steps to reproduce

1. Set up multi-company with 2+ companies, each with their own chart of accounts
2. Log in as a user whose default company is Company A
3. Open General Ledger wizard, select Company B
4. Click Export XLSX or Export PDF

### Traceback

```
File "account_financial_report/report/general_ledger_xlsx.py", line 151, in _generate_report_content
    account["code"] + " - " + accounts_data[account["id"]]["name"],
TypeError: unsupported operand type(s) for +: 'bool' and 'str'
```

### Root cause

`_get_accounts_data()` calls `self.env["account.account"].browse(accounts_ids)` without `with_company(company_id)`. The `code` computed property uses `self.env.company.id` to look up the code from `code_store`, which defaults to the user's company (Company A) rather than the report's target company (Company B).

## Fix

- Add optional `company_id` parameter to `_get_accounts_data()` in `abstract_report.py` and call `with_company()` when provided — this is the **root cause fix** at the data source
- All callers (General Ledger, Trial Balance, Aged Partner Balance, Open Items) already have `company_id` in scope and now pass it through
- Defensive `or ""` guards on 3 remaining direct ORM accesses in `trial_balance.py` and `journal_ledger.py` where `account.code` is used outside the centralized data dict

## Affected files

| File | Change |
|------|--------|
| `abstract_report.py` | `_get_accounts_data()` accepts `company_id`, calls `with_company()` |
| `general_ledger.py` | Pass `company_id` to `_get_accounts_data()` |
| `trial_balance.py` | Pass `company_id` + guard 2 direct ORM accesses |
| `aged_partner_balance.py` | Pass `company_id` |
| `open_items.py` | Pass `company_id` |
| `journal_ledger.py` | Guard `account.code` in data dict construction |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/account_financial_report/report/abstract_report.py
+++ b/account_financial_report/report/abstract_report.py
@@ -122,12 +122,11 @@ class AgedPartnerBalanceReport(models.AbstractModel):
                 move_line["amount_currency"] = 0
         return move_lines
 
-    def _get_accounts_data(self, accounts_ids):
-        accounts = (
-            self.env["account.account"]
-            .with_context(active_test=False)
-            .browse(accounts_ids)
-        )
+    def _get_accounts_data(self, accounts_ids, company_id=False):
+        AccountAccount = self.env["account.account"].with_context(active_test=False)
+        if company_id:
+            AccountAccount = AccountAccount.with_company(company_id)
+        accounts = AccountAccount.browse(accounts_ids)
         accounts_data = {}
         for account in accounts:
             accounts_data.update(
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -248,7 +248,7 @@ class AgedPartnerBalanceReport(models.AbstractModel):
                 date_at_object,
             )
         journals_data = self._get_journals_data(list(journals_ids))
-        accounts_data = self._get_accounts_data(ag_pb_data.keys())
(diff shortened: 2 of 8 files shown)


pedrobaeza:
Isn't this because the company for retrieving the code is not correct?

dnplkndll:
I am still checking. I think might be if you have >1 company checked? but this is wip needs more review.

pedrobaeza:
Yes, when you are looking several companies at once, the code is retrieved for the current company, not the rest, so a deep refactoring is needed for storing the account and the company from which the journal items were fetched I think, or do an initial grouping by company. The code of this module is hard though...

PR #1500, opened 2026-05-06
[18.0][FIX]account_financial_report: missing accounts

ChristophAbenthungCibex:
Fixes that some accounts are not found at the general ledger wizard.
Same fix as #1098 (that fix was for the trial balance wizard)

### DIFF
--- a/account_financial_report/tests/test_general_ledger.py
+++ b/account_financial_report/tests/test_general_ledger.py
@@ -11,6 +11,8 @@ from odoo.tests import tagged
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 
+from .helpers import check_all_accounts_loaded
+
 
 @tagged("post_install", "-at_install")
 class TestGeneralLedgerReport(AccountTestInvoicingCommon):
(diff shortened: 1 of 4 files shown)


BhaveshHeliconia (approved):
LGTM!

account_financial_report/tests/test_general_ledger.py
```diff
@@ -687,6 +695,36 @@ def test_04_unaffected_account_balance_2_years(self):
...
+                "fy_start_date": self.fy_date_start,
+                "account_code_from": self.account001.id,
+                "account_code_to": all_accounts[-1].id,
+            }
+        )
+        general_ledger.on_change_account_range()
+        # sets are needed because some codes are duplicated and
+        # thus the length of all_accounts would be higher
+        all_accounts_code_set = set()
+        trial_balance_code_set = set()
```
pedrobaeza: Incorrect variable name from copy/paste, but adding all of these lines for doing the same doesn't make sense. You can extract a generic method, just passing the model name or similar to do both checks.

ChristophAbenthungCibex: Thx for the input. Changed it. Is it ok now?

pedrobaeza (approved):

PR #1503, opened 2026-05-11
[17.0][FIX] account_invoice_report: Add init balance for grouped_by "none"

Andrii9090-tecnativa:
Without grouping by `partner_id` or `taxes`, we can’t see the initial balance. This PR fixes this bug and adds the balance for the non-grouped case.

@pedrobaeza can you review this?
@Tecnativa
TT62304


pedrobaeza:
Incorrect module name

PR #1504, opened 2026-05-11
[17.0][FIX] account_financial_report: Add init balance for grouped_by "none"

Andrii9090-tecnativa:
Without grouping by `partner_id` or `taxes`, we can’t see the initial balance. This PR fixes this bug and adds the balance for the non-grouped case.

@pedrobaeza can you review this?
@Tecnativa
TT62304

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -61,6 +61,9 @@ class GeneralLedgerReport(models.AbstractModel):
         ]
 
     def _get_acc_prt_accounts_ids(self, company_id, grouped_by):
+        # In non-grouped mode, lines must stay at account level.
+        if grouped_by == "none":
+            return []
         accounts_domain = [
             ("company_id", "=", company_id),
         ] + self._get_account_type_domain(grouped_by)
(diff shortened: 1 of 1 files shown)


account_financial_report/report/general_ledger.py
```diff
@@ -168,15 +171,35 @@ def _get_gl_initial_acc(
...
             account_ids, company_id, date_from, fy_start_date, base_domain
         )
-        return self._get_accounts_initial_balance(initial_domain_bs, initial_domain_pl)
+        gl_initial_acc = self._get_accounts_initial_balance(
+            initial_domain_bs, initial_domain_pl
+        )
+        if grouped_by != "none":
+            return gl_initial_acc
+        # If grouped_by none we add balance
+        domain = list(base_domain) + [("date", "<", date_from)]
```
pedrobaeza, line 180: `date_from` for the initial balance?

Andrii9090-tecnativa: I used `date_from` because we need to know the initial balance before the report start date.

The same approach is used in another part of the general ledger report ([L91](https://github.com/OCA/account-financial-reporting/blob/17.0/account_financial_report/report/general_ledger.py#L91))

pedrobaeza: Oh, my bad, I read `from` but understood `to`.

pedrobaeza (approved):

victoralmau (approved):

PR #1508, opened 2026-05-14
[18.0][FIX] account_financial_report: Trial balance: show correctly data for multiple currencies

pedrobaeza:
After #1387, a new grouping key was added (`currency_id`), but the code handling the resulting data was not prepared to find several groups for the same account/partner combination, which may happen if your book operations in several currencies for the same partner (it's a weird case, but following Murphy's law, it has happened in one of my customers).

This commit fixes the data preparation for having that into account.

@Tecnativa TT62400

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -265,10 +265,10 @@ class TrialBalanceReport(models.AbstractModel):
                     tb, foreign_currency
                 )
             else:
-                total_amount[acc_id]["initial_balance"] = tb["balance"]
+                total_amount[acc_id]["initial_balance"] += tb["balance"]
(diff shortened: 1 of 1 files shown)


victoralmau (approved):
That makes sense.

micheledic (approved):

Andrii9090-tecnativa (approved):

pedrobaeza:
Thanks for your quick reaction!

/ocabot merge patch

pedrobaeza:
Aborting because we may still have an incorrect result.

pedrobaeza:
Now the ending balance is also handled when grouping by partner, and I have included also sums for account computation just in case.

/ocabot merge patch

PR #1511, opened 2026-05-20
[19.0][FIX] account_financial_report: skip line_subsection in domains

micheledic:
Odoo 19 added `line_subsection` to `account.move.line.display_type`. Like `line_note` and `line_section`, that has not `account_id`.

  Trial Balance, General Ledger and Journal Ledger was fitlering only the two older display types, so a posted move with a subsection row in the reporting period reaches `formatted_read_group` as a `False` group and crashes `_compute_account_amount`.

  ## Error

  Printing the Trial Balance over a period that contains at least one
  posted move with a `line_subsection` row raises:

>   File ".../account_financial_report/report/trial_balance.py", line 922, in _get_report_values
>       total_amount, accounts_data, partners_data = self._get_data(
>   File ".../account_financial_report/report/trial_balance.py", line 583, in _get_data
>       total_amount = self._compute_account_amount(
>   File ".../account_financial_report/report/trial_balance.py", line 226, in _compute_account_amount
>       acc_id = tb["account_id"][0]
>                ~~~~~~~~~~~~~~~~^^^
>   TypeError: 'bool' object is not subscriptable

  
  ## Fix

  Add line_subsection to the domains so it is not included in `formatted_read_group`

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -415,7 +415,11 @@ class GeneralLedgerReport(models.AbstractModel):
         cost_center_ids,
     ):
         domain = [
-            ("display_type", "not in", ["line_note", "line_section"]),
+            (
+                "display_type",
+                "not in",
+                ["line_note", "line_section", "line_subsection"],
+            ),
             ("date", ">=", date_from),
             ("date", "<=", date_to),
         ]
--- a/account_financial_report/tests/test_general_ledger.py
+++ b/account_financial_report/tests/test_general_ledger.py
@@ -7,6 +7,7 @@ import time
 from datetime import date
 
 from odoo import api, fields
+from odoo.fields import Command
 from odoo.tests import tagged
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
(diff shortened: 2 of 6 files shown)


micheledic:
@pedrobaeza this bug is very blocking if somebody use subsections

account_financial_report/report/general_ledger.py
```diff
@@ -415,7 +415,11 @@ def _get_period_domain(
         cost_center_ids,
     ):
         domain = [
-            ("display_type", "not in", ["line_note", "line_section"]),
+            (
+                "display_type",
```
pedrobaeza, line 419: Isn't better to do `display_type = False/none` for avoiding this case in the future?

micheledic: it is not correct because the display_type can also be
```
('product',
'tax',
'payment_term',
'rounding',
'discount',
'cogs',
'epd',
etc )
```

pedrobaeza: OK then.

account_financial_report/tests/test_general_ledger.py
```diff
@@ -783,3 +783,56 @@ def test_05_onchange_account_range_no_typeerror(self):
...
+        journal = self.env["account.journal"].search(
+            [("company_id", "=", self.env.user.company_id.id)], limit=1
+        )
+        move = self.env["account.move"].create(
+            {
+                "journal_id": journal.id,
+                "date": self.fy_date_start,
+                "line_ids": [
+                    (
+                        0,
```
pedrobaeza: For taking less lines, please use `Command.add`

micheledic: done

micheledic: @pedrobaeza everything is green ;)

pedrobaeza (approved):
Fast-tracking for fixing the problem.

/ocabot merge patch

PR #1513, opened 2026-05-20
[19.0] [account_financial_report/report/general_ledger_xlsx.py]: : ValueError: Unknown format code 'd' for object of type 'float' on XLSX export

jonathandbdb:
Install account_financial_report on an Odoo 19 instance.
Ensure at least one journal entry has analytic distribution set (any percentage).
Open Accounting → Reporting → General Ledger.
Fill in the wizard and click Export to XLSX.
Expected behavior

The Excel file is generated successfully. Analytic distribution percentages are displayed correctly (e.g. Marketing 75%).

Additional context

Full traceback from the server:
File ".../account_financial_report/report/general_ledger_xlsx.py", line 210, in _generate_report_content
analytic_distribution += f"{value:d}%"
ValueError: Unknown format code 'd' for object of type 'float'

Odoo's field (JSON) stores percentage values as float. The :d format specifier is only valid for integers in Python. Replacing it with :g (or casting to int if whole-number percentages are guaranteed) resolves the issue.

Odoo: 19.0

### DIFF
--- a/account_financial_report/tests/test_general_ledger.py
+++ b/account_financial_report/tests/test_general_ledger.py
@@ -831,3 +831,104 @@ class TestGeneralLedgerReport(AccountTestInvoicingCommon):
                 entry.get("id"),
                 f"Report contains a line with falsy id: {entry}",
             )
+
+    def test_06_analytic_distribution_partial_percentage_xlsx(self):
+        """Regression test: XLSX export must not raise ValueError when
+        analytic distribution value is a float < 100.
+        Covers both the centralized (no-partner) and the partner-grouped paths
+        in general_ledger_xlsx._generate_report_content.
+        """
+        from io import BytesIO
+
+        import xlsxwriter
+
+        # Create analytic plan and account
+        analytic_plan = self.env["account.analytic.plan"].create(
+            {"name": "Test Plan GL XLSX"}
+        )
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Thanks for the patch. Instead of adding a full test redoing all the process, I think it should be enough to add an analytic distribution in the existing test setup for executing the same steps.

account_financial_report/tests/test_general_ledger.py
```diff
@@ -349,6 +355,42 @@ def test_01_account_balance(self):
         self.assertEqual(income_fin_balance["credit"], 0)
         self.assertEqual(income_fin_balance["balance"], 2000)
 
+        # Regression: XLSX export with partial analytic distribution (< 100 %)
```
pedrobaeza, line 358: I would say XLSX export is already present, isn't it? Can you reuse it?

account_financial_report/tests/test_general_ledger.py
```diff
@@ -784,50 +863,3 @@ def test_05_onchange_account_range_no_typeerror(self):
             wizard.account_ids,
             "Accounts out of the range should NOT be in the filter.",
         )
-
-    def test_06_line_subsection_excluded(self):
```
pedrobaeza, line 788: You have removed this recently added test.

PR #1516, opened 2026-05-22
[18.0][IMP] account_financial_report: improve floating point precision in XLSX amount cells + remove dead code

eduezerouali-tecnativa:
cc @Tecnativa TT62400
ping @pedrobaeza @victoralmau 

When generating XLSX reports, amount values resulting from arithmetic operations (e.g. summing negative floats) could produce IEEE 754 floating point artifacts like -1152.4299999999998 instead of -1152.43.

The PR applies currency.round() on all amount values before writing them to the sheet, using the company currency defined in the wizard.

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -185,54 +185,6 @@ class AbstractReportXslx(models.AbstractModel):
             )
         report_data["row_pos"] += 1
 
-    def write_line(self, line_object, report_data):
-        """Write a line on current line using all defined columns field name.
-        Columns are defined with `_get_report_columns` method.
-        """
-        for col_pos, column in report_data["columns"].items():
-            value = getattr(line_object, column["field"])
-            cell_type = column.get("type", "string")
-            if cell_type == "many2one":
-                report_data["sheet"].write_string(
-                    report_data["row_pos"],
-                    col_pos,
-                    value.name or "",
-                    report_data["formats"]["format_right"],
-                )
-            elif cell_type == "string":
-                if (
-                    hasattr(line_object, "account_group_id")
-                    and line_object.account_group_id
-                ):
-                    report_data["sheet"].write_string(
-                        report_data["row_pos"],
-                        col_pos,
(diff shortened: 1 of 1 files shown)


account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -10,6 +10,12 @@ class AbstractReportXslx(models.AbstractModel):
     _description = "Abstract XLSX Account Financial Report"
     _inherit = "report.report_xlsx.abstract"
 
+    def _get_currency_from_wizard(self):
+        model = self.env.context.get("active_model")
+        active_id = self.env.context.get("active_id")
+        wizard = self.env[model].browse(active_id)
+        return wizard.company_id.currency_id
```
pedrobaeza, line 17: This is not correct, as we need the currency of the cell, for example for amounts in currency.

pedrobaeza (changes requested):

eduezerouali-tecnativa:
Changes applied. The logic now uses the currency defined on the line to round the value; if no currency is set there, it falls back to the currency from the wizard’s company_id.

Also included an additional commit to remove unused dead code.

eduezerouali-tecnativa:
ping @pedrobaeza

account_financial_report/report/abstract_report_xlsx.py
```diff
@@ -10,6 +10,20 @@ class AbstractReportXslx(models.AbstractModel):
...
+
+    def _round_value_by_currency(self, value, currency=None):
+        if isinstance(currency, tuple):
+            currency = currency[0]
+        if currency:
+            return self.env["res.currency"].browse(currency).round(value)
```
pedrobaeza, line 23: Name correctly variables: if this is an id, you can't call it `currency`, but `currency_id`.

pedrobaeza (approved):
Checked and sounds OK, and the dead code removal is always good (how bloated this module is...).

Andrii9090-tecnativa (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to 19

PR #1518, opened 2026-05-25
[18.0][FIX] account_financial_report: avoid currency grouping without foreing currency

RabbitJon-S73:
## Summary

This PR fixes the Trial Balance computation in `account_financial_report` when partner details are enabled and foreign currency is disabled.

The change avoids grouping `account.move.line` by `currency_id` in that scenario, while preserving the current behavior when `foreign_currency=True`. Test coverage is added to validate consistency between account totals, partner detail lines, and the XLSX export.

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -206,7 +206,9 @@ class TrialBalanceReport(models.AbstractModel):
         initial_balances = self.env["account.move.line"].read_group(
             domain=domain,
             fields=["account_id", "balance", "amount_currency:sum"],
-            groupby=["account_id", "currency_id"],
+            groupby=(
+                ["account_id", "currency_id"] if foreign_currency else ["account_id"]
+            ),
         )
         pl_initial_balance = 0.0
         pl_initial_currency_balance = 0.0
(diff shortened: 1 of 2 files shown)


account_financial_report/report/trial_balance.py
```diff
@@ -357,8 +359,14 @@ def _compute_partner_amount(
...
                     tb, foreign_currency
                 )
+                if foreign_currency:
+                    total_amount[acc_id][prt_id]["initial_currency_balance"] = 0.0
             else:
                 total_amount[acc_id][prt_id]["ending_balance"] += tb["balance"]
+                if foreign_currency:
+                    total_amount[acc_id][prt_id]["ending_currency_balance"] += round(
+                        tb["amount_currency"], 2
+                    )
```
eduezerouali-tecnativa, line 369: Instead of rounding to 2 decimal why don´t  you use `res.currency`  round method that will round to decimal places set on the currency.
```suggestion
                    total_amount[acc_id][prt_id]["ending_currency_balance"] += foreign_currency.round(
                        tb["amount_currency"]
                    )
```

pedrobaeza:
AFAIK, I already fixed the possible problems in https://github.com/OCA/account-financial-reporting/commit/4d1840f0d235b88c96393e9f88fa8447652a1f1c

Is there any other problem?

RabbitJon-S73:
@pedrobaeza we tried with you changes and seems that they solve the problem. I proceed to close de PR

PR #1523, opened 2026-06-08
[17.0][ADD]partner_statement_periodic_mail_send

GuillermoNforgeflow:
This module automatically sends a monthly Activity Statement PDF to each customer on the last day of the month. The statement covers the full calendar month and includes all invoices issued, payments received, and an aging breakdown of the outstanding balance. Customers can be individually excluded via a flag on their partner record.


partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,133 @@
...
+            mail.write({"attachment_ids": [Command.set(attachment_ids)]})
+        except Exception:
+            _logger.exception(
+                "Failed to render periodic statement PDF for %s", self.display_name
+            )
+            mail.unlink()
+            return
+        mail.send()
+
+    def action_send_periodic_statement(self, date_start=None, date_end=None):
```
GuillemCForgeFlow:
```suggestion
    @api.model
    def action_send_periodic_statement(self, date_start=None, date_end=None):
```
you can add this to the methods where the content of self is not relevant

partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,133 @@
...
+                mail, date_start, date_end
+            )
+            mail.write({"attachment_ids": [Command.set(attachment_ids)]})
+        except Exception:
+            _logger.exception(
+                "Failed to render periodic statement PDF for %s", self.display_name
+            )
+            mail.unlink()
+            return
+        mail.send()
```
GuillemCForgeFlow: this solution can work, but what is more typical to use instead is:
```template.with_context(cleaned_context).send_mail(res_id=self.id, ..., email_values=email_values)```

you'll save quite some lines as most of the parts being added are done automatically in odoo's core

File: partner_statement_periodic_mail_send/models/res_partner.py [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+        partner = self.env["res.partner"].browse(invoice_contact)
+        if not partner.email:
+            _logger.warning(
+                "Skipping periodic statement for %s: no email address",
+                self.display_name,
+            )
+            return
+        mail_tpl = self.env.ref(
+            "partner_statement_periodic_mail_send.periodic_statement_mail_template"
+        )
```
GuillemCForgeFlow:
```suggestion
        mail_tpl = self.env.ref(
            "partner_statement_periodic_mail_send.periodic_statement_mail_template", raise_if_not_found=False
        )
        if not mail_tpl:
            return False
```
you should always cover the case where the mail template is not found, otherwise it may break the rest of the code

partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,133 @@
...
+            self.name.replace(" ", "_"),
+            date_end.strftime("%Y%m"),
+            report_format,
+        )
+        attach = iao.create(
+            {
+                "name": filename,
+                "datas": base64.b64encode(report_bin),
+                "res_model": "mail.message",
+                "res_id": mail.mail_message_id.id,
```
GuillemCForgeFlow: why is it necessary to link the attachment to the mail message?

partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,133 @@
...
+            ]
+        )
+        for customer in customers:
+            customer._send_periodic_statement_mail(date_start, date_end)
+
+    def _cron_send_periodic_statement(self):
+        today = date.today()
+        last_day = monthrange(today.year, today.month)[1]
+        if today.day != last_day:
+            return
```
GuillemCForgeFlow, line 131: I'd move this evaluation to a separate method, so that is easy to inherit in a custom module if anyone else wants to go with another approach (first day of month, ...)

partner_statement_periodic_mail_send/models/res_partner.py
GuillemCForgeFlow:
```suggestion
    def _send_periodic_statement(self, date_start=None, date_end=None):
```
`action` methods are more oriented to be used as per server actions, buttons in the UI, ...

partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,133 @@
...
+            date_start = today.replace(day=1)
+            last_day = monthrange(today.year, today.month)[1]
+            date_end = today.replace(day=last_day)
+        customers = self.search(
+            [
+                ("customer_rank", ">", 0),
+                ("parent_id", "=", False),
+                ("exclude_periodic_statement", "=", False),
+            ]
+        )
```
GuillemCForgeFlow: I'd move this to a method instead, and I'd also add a date field to record when it was last sent the statement to the customer, to be safe of not sending it again.
notice that if the method is executed twice in the last day of the month, it will be sent twice per customer

GuillemCForgeFlow (changes requested):

partner_statement_periodic_mail_send/models/res_partner.py
```diff
@@ -0,0 +1,134 @@
...
+                mail, date_start, date_end
+            )
+            mail.write({"attachment_ids": [Command.set(attachment_ids)]})
+        except Exception:
+            _logger.exception(
+                "Failed to render periodic statement PDF for %s", self.display_name
+            )
+            mail.unlink()
+            return
+        mail.send()
```
GuillemCForgeFlow:
```suggestion
     email_template.send_mail(self.id, force_send=True, email_values={'attachment_ids': [dte_attachment.id]})
```
instead of all this, you can do something like this

is there any specific part that needs to be done in the way you are doing?

GuillemCForgeFlow (approved):
code review lgtm 👍🏿

BhaveshHeliconia (approved):
Functional review LGTM!

PR #1525, opened 2026-06-09
[18.0][FIX] account_tax_balance: Compute financial_type correctly when the total is zero

carlos-lopez-tecnativa:
Before this commit, for records with total = 0, the `financial_type` was computed incorrectly:

`out_invoice` → `receivable_refund` (should be `receivable`) 
`in_invoice` → `payable_refund` (should be `payable`)

This commit fixes the computation of the `financial_type` field according to the move type.

TT62976
@Tecnativa @pedrobaeza Could you please review this?

### DIFF
--- a/account_tax_balance/models/account_move.py
+++ b/account_tax_balance/models/account_move.py
@@ -34,6 +34,15 @@ class AccountMove(models.Model):
                 ).mapped("balance")
             )
 
+        mapping = {
+            "entry": "other",
+            "out_invoice": "receivable",
+            "out_receipt": "receivable",
+            "out_refund": "receivable_refund",
+            "in_invoice": "payable",
+            "in_receipt": "payable",
+            "in_refund": "payable_refund",
+        }
         for move in self:
             account_types = move.line_ids.mapped("account_id.account_type")
             if (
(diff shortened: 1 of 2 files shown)


account_tax_balance/models/account_move.py
```diff
@@ -34,6 +34,15 @@ def _balance_get(line_ids, account_type):
                 ).mapped("balance")
             )
 
+        financial_type_map = {
```
pedrobaeza: For having later code in one line, and being a local variable:

```suggestion
        mapping = {
```

account_tax_balance/models/account_move.py
```diff
@@ -43,11 +52,21 @@ def _balance_get(line_ids, account_type):
                 move.financial_type = "liquidity"
             elif "liability_payable" in account_types:
                 balance = _balance_get(move.line_ids, "liability_payable")
-                move.financial_type = "payable" if balance < 0 else "payable_refund"
+                if move.company_currency_id.is_zero(balance):
+                    move.financial_type = (
+                        financial_type_map.get(move.move_type) or "other"
+                    )
```
pedrobaeza:
```suggestion
                    move.financial_type = mapping.get(move.move_type, "other")
```

account_tax_balance/models/account_move.py
```diff
@@ -43,11 +52,21 @@ def _balance_get(line_ids, account_type):
...
+                    move.financial_type = "payable" if balance < 0 else "payable_refund"
             elif "asset_receivable" in account_types:
                 balance = _balance_get(move.line_ids, "asset_receivable")
-                move.financial_type = (
-                    "receivable" if balance > 0 else "receivable_refund"
-                )
+                if move.company_currency_id.is_zero(balance):
+                    move.financial_type = (
+                        financial_type_map.get(move.move_type) or "other"
+                    )
```
pedrobaeza:
```suggestion
                    move.financial_type = mapping.get(move.move_type, "other")
```

pedrobaeza (approved):

victoralmau (approved):

PR #1526, opened 2026-06-12
[17.0][FIX] account_financial_report: split when account have multiple anaytics distributions

eduezerouali-tecnativa:
cc @Tecnativa TT63092
@pedrobaeza @victoralmau 

When a journal item is distributed among analytic accounts belonging to several analytic plans, the `analytic_distribution` keys are the analytic account ids joined by commas (e.g. `"337,357"`).

Solution is to split the distribution key on commas and resolve each analytic account
individually, consistently with the non-grouped XLSX branch and the QWeb
template, which already handle this case.

Includes a regression test.


pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
Please fw-port it to upper versions.

/ocabot merge patch

PR #1532, opened 2026-06-19
[19.0][FIX] account_financial_report: format analytic percentages in XLSX

### LINKED ISSUE #1530: 19.0 - General Ledger XLSX export crashes with ValueError on partial analytic distribution
### Module

  `account_financial_report` (19.0)

  ### Describe the bug

  Exporting the **General Ledger** report to XLSX raises a `ValueError` when at
  least one journal item has a **partial analytic distribution** (a percentage
  < 100). The percentage is formatted with the integer format code `:d`, but in
  Odoo the `analytic_distribution` values are **floats**, so formatting fails.

  ### Traceback

  File ".../account_financial_report/report/general_ledger_xlsx.py", line 210, in _generate_report_content
      analytic_distribution += f"{value:d}%"
                                 ^^^^^^^^^
  ValueError: Unknown format code 'd' for object of type 'float'

  ### Steps to reproduce

  1. Have at least one posted move line with an analytic distribution where a
     plan entry is assigned a percentage lower than 100 (e.g. `50%`).
  2. Open **Invoicing / Reporting / OCA accounting reports / General Ledger**.
  3. Enable **Show analytic distribution** (`show_cost_center`).
  4. Export the report to **XLSX**.

  The export crashes with the traceback above. The PDF/HTML export is not
  affected.

  ### Root cause

  `analytic_distribution` dict values are `float` (e.g. `50.0`), and the f-string
  uses the integer format code `:d`, which only accepts `int`.

  There are **two occurrences** of the same bug in
  `report/general_ledger_xlsx.py`:

  ```python
  if value < 100:
      analytic_distribution += f"{value:d}%"      # first occurrence
  ...
  if value < 100:
      analytic_distribution += f"{value:d}% "     # second occurrence

  Proposed fix

  Replace the integer format code :d with :g, which handles floats and also
  drops trailing zeros (so 50.0 → 50%, 33.33 → 33.33%):

  analytic_distribution += f"{value:g}%"

  (both occurrences).

  Versions

  - Odoo: 19.0
  - account_financial_report: 19.0 (latest)

gaoflow:
Fixes #1530.

The General Ledger XLSX renderer formatted analytic distribution percentages with the integer-only `:d` formatter. Odoo stores `analytic_distribution` percentages as floats, so partial values such as `50.0` crash the export with `ValueError: Unknown format code 'd' for object of type 'float'`.

This switches partial percentage formatting to `:g`, which accepts floats while preserving clean output such as `50%`, and extends the existing XLSX analytic distribution test with partial float percentages.

Checks run locally:

- `python -m py_compile account_financial_report/report/general_ledger_xlsx.py account_financial_report/tests/test_general_ledger.py`
- `uv run --python 3.11 --with ruff ruff check account_financial_report/report/general_ledger_xlsx.py account_financial_report/tests/test_general_ledger.py`
- `uv run --python 3.11 --with ruff ruff format --check account_financial_report/report/general_ledger_xlsx.py account_financial_report/tests/test_general_ledger.py`
- `git diff --check`

I could not run the Odoo integration test locally because this environment does not have the `odoo` Python package installed.

AI assistance was used under my direction.


account_financial_report/tests/test_general_ledger.py
```diff
@@ -756,6 +757,9 @@ def test_analytic_distribution_xlsx(self):
                             "credit": 1000,
                             "account_id": self.income_account.id,
                             "partner_id": self.partner.id,
+                            "analytic_distribution": {
```
pedrobaeza: Where is this checked?

pedrobaeza: I mean. Does this provoke a crash without your patch? Or what is the aim of this change?

binggao1230:
The added lines give the second move line (income account, not partner-grouped) an analytic distribution so the report hits the other branch in `general_ledger_xlsx.py`. Without the patch, that branch appends a tuple to the string (`"%s %d%% ", (names, value)`) and rendering fails with `TypeError`; the first branch raises `ValueError` on float values (`Unknown format code d`). The render completing (`assertTrue(content)`) is the check — both branches must succeed for the XLSX to render.

binggao1230:
Note for triage: 19.0 now carries the equivalent change in 97533329 ('support decimal analytic percentages in XLSX exports') — both `:g` formatting and the corrected second branch are in, and the test already renders both move lines with analytic distributions. A rebase leaves this PR empty, so I've left it as-is; happy to close if you prefer.

PR #1534, opened 2026-06-22
[19.0][FIX] account_financial_report: use report data company for XLS…

adrip-s73:
Fix General Ledger XLSX export when launched from a partner action. The XLSX helper was resolving the company currency from active_model / active_id, but in this flow the active record is the partner, not the transient wizard.
Justification
When exporting the General Ledger XLSX from a contact, the report context can contain active_model=res.partner and active_id=<partner_id>. The XLSX formatter then tried to browse that ID as general.ledger.report.wizard, causing a MissingError such as:
general.ledger.report.wizard(21906)
The report data already contains the correct company_id, so the currency should be resolved from data["company_id"] instead of relying on the active record context.

### DIFF
--- a/account_financial_report/report/abstract_report_xlsx.py
+++ b/account_financial_report/report/abstract_report_xlsx.py
@@ -10,19 +10,23 @@ class AbstractReportXslx(models.AbstractModel):
     _description = "Abstract XLSX Account Financial Report"
     _inherit = "report.report_xlsx.abstract"
 
-    def _get_currency_from_company(self):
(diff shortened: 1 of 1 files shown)


Reyes4711-S73 (approved):
LGTM

Raul-S73 (approved):

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Can you backport this to 18?

PR #1538, opened 2026-06-24
[19.0][FIX] account_financial_report: sum all currencies in Trial Balance period totals

### LINKED ISSUE #1537: [19.0] account_financial_report: Trial Balance drops all but the last currency for multi-currency accounts
module: account_financial_report
version: 19.0 (19.0.0.0.15 / current HEAD) — same code path in 18.0 / 17.0 / 16.0

**Steps to reproduce**

1. On a company in currency A (e.g. CAD), post entries so that, within the
   reporting period, a single balance-sheet account holds move lines in
   **more than one currency** — e.g. a bank "Outstanding Payments" / suspense
   account fed by both local and foreign-currency vendor payments
   (CAD + USD + EUR + GBP). This is normal for anyone paying foreign vendors.
2. Run **Invoicing → Reporting → OCA accounting reports → Trial Balance**
   for the period (e.g. 2025-01-01 → 2025-12-31), Target Moves = *All Posted
   Entries*, Show Partner Details = off.
3. Run the **General Ledger** for the same account / period / target as a
   cross-check.

**Current behavior**

For that account the Trial Balance shows the figures of **only one currency
sub-group** (the last one returned by the period read group) instead of the
total across all currencies. Debit, Credit, Period balance and Ending balance
are all wrong and do **not** match the General Ledger (which is correct).

Concrete example (real data): an account whose true posted period balance is
**-18,056.23** (CAD -2,900.23 + EUR 0 + GBP 0 + USD -15,156.00) is reported by
the Trial Balance as **-15,156.00** — exactly the USD sub-group, with CAD/EUR/GBP
dropped.

Two visible consequences:
- **Trial Balance ≠ General Ledger** for every account with multi-currency
  movements.
- **"Target Moves" appears to have no effect**: switching All Entries ↔ All
  Posted Entries does not change the displayed figure whenever the only
  draft/unposted entries fall in a dropped (non-last) currency sub-group — the
  two exports come out byte-for-byte identical.

Single-currency accounts are unaffected, so most of the report ties out and
only multi-currency accounts are wrong, which makes this easy to miss.

**Expected behavior**

The Trial Balance should sum all currency sub-groups per account and match the
General Ledger.

**Root cause**

`report/trial_balance.py`, `TrialBalanceReport._compute_account_amount`. The
period read group (`tb_period_acc`) is grouped by account **and** currency, so
an account with N currencies yields N rows. The loop assigns rather than
accumulates:

```python
for tb in tb_period_acc:
    acc_id = tb["account_id"][0]
    total_amount[acc_id] = self._prepare_total_amount(tb, foreign_currency)
    total_amount[acc_id]["credit"] = tb["credit:sum"]   # '=' clobbers prior currency
    total_amount[acc_id]["debit"] = tb["debit:sum"]
    total_amount[acc_id]["balance"] = tb["balance:sum"]
```

so each currency row overwrites the previous and only the last survives.

Commit af77270f6 ("[FIX] Trial balance: show correctly data for multiple
currencies", 2026-05-14) fixed the analogous `=`→`+=` problem in the
*initial-balance* loop and both *partner-detail* loops, but left this first
(period, non-partner-details) loop untouched and added no test, so the
regression persists for the default Trial Balance. Confirmed still present on
19.0 HEAD.

A pull request with the fix and a regression test follows.

kwoychesko:
## Problem
The Trial Balance reports wrong Debit/Credit/Balance for any account that has
period move lines in **more than one currency**: only the last currency
sub-group is kept. Side effects: the "Target Moves" (All vs Posted) toggle looks
ineffective when the dropped currencies carry the only unposted entries, and the
Trial Balance stops matching the General Ledger.

Fixes #1537.

## Root cause
`TrialBalanceReport._compute_account_amount` iterates `tb_period_acc`, grouped by
account **and** currency (one row per `(account, currency)`), and **assigns**
instead of **accumulating** the per-currency figures. Commit af77270f6 fixed the
same `=`→`+=` issue in the initial-balance and partner-detail loops but missed
this first (period, non-partner-details) loop and shipped no test, so the default
Trial Balance is still affected.

## Fix
- Accumulate the per-currency period rows (debit/credit/balance/ending balance,
  plus the currency balances when "Show foreign currency" is on) instead of
  overwriting them.
- Add a regression test that posts two currencies on one account and asserts the
  summed total (100 company-currency + 60 == 160), run both with and without
  "Show foreign currency".

## Notes
The analytic *group by* path (`group_by_data`) has the same last-currency-wins
behaviour and would need an analogous change, but it only triggers with "Group by
Analytic Account" **and** multi-currency; left out of scope here to keep the fix
focused and fully covered. Happy to address it in a follow-up.

## Backport
The same code path exists on 18.0 / 17.0 / 16.0 — happy to forward-/back-port.

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -228,13 +228,30 @@ class TrialBalanceReport(models.AbstractModel):
     ):
         for tb in tb_period_acc:
             acc_id = tb["account_id"][0]
-            total_amount[acc_id] = self._prepare_total_amount(tb, foreign_currency)
-            total_amount[acc_id]["credit"] = tb["credit:sum"]
-            total_amount[acc_id]["debit"] = tb["debit:sum"]
-            total_amount[acc_id]["balance"] = tb["balance:sum"]
-            total_amount[acc_id]["initial_balance"] = 0.0
-            if foreign_currency:
-                total_amount[acc_id]["initial_currency_balance"] = 0.0
+            # ``tb_period_acc`` is grouped by account *and* currency, so an
+            # account holding move lines in several currencies produces one
+            # row per currency here. Those rows must be accumulated, not
+            # overwritten, otherwise the account only keeps the figures of its
+            # last currency sub-group (wrong debit/credit/balance, and the
+            # "Target Moves" filter looks ineffective when the dropped
+            # currencies carry the only unposted entries).
+            if acc_id not in total_amount:
+                total_amount[acc_id] = self._prepare_total_amount(tb, foreign_currency)
+                total_amount[acc_id]["credit"] = tb["credit:sum"]
+                total_amount[acc_id]["debit"] = tb["debit:sum"]
+                total_amount[acc_id]["balance"] = tb["balance:sum"]
+                total_amount[acc_id]["initial_balance"] = 0.0
+                if foreign_currency:
+                    total_amount[acc_id]["initial_currency_balance"] = 0.0
+            else:
+                total_amount[acc_id]["credit"] += tb["credit:sum"]
+                total_amount[acc_id]["debit"] += tb["debit:sum"]
+                total_amount[acc_id]["balance"] += tb["balance:sum"]
+                total_amount[acc_id]["ending_balance"] += tb["balance:sum"]
+                if foreign_currency:
+                    total_amount[acc_id]["ending_currency_balance"] += round(
+                        tb["amount_currency:sum"], 2
+                    )
             if "__context" in tb and "group_by" in tb["__context"]:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Instead of the AI description, which is long, but not the most descriptive, please show visually with screenshots the problem.

kwoychesko:
The description is completely accurate and conveys the problem better than a picture of an incorrect balance and a correct balance side-by-side. It’s hard to show a picture of a code loop overwriting instead of accumulating.

pedrobaeza:
I was asking for the screenshot/PDF of an incorrect trial balance because of that.

PR #1539, opened 2026-06-26
[FIX] account_financial_report: restore open items xlsx ending balance

twalter-c2c:
The account ending balance line in Open Items XLSX became conditional after e14373d.
This diverges from the qweb-html and qweb-pdf Open Items outputs, where account ending balance is still shown.
This patch restores the previous XLSX behavior by always printing account ending balance for account sections.


TDu (approved):
Make sense.
@carolinafernandez-tecnativa Could you look at it please ?

grindtildeath (approved):

PR #1336, opened 2025-06-13
[IMP] account_financial_report: Hide accounts with no end balance

SirPyTech:
Add a new flag in the wizard that allows the user to hide from the report the accounts that have 0 ending balance.

PR #1541, opened 2026-06-30
[IMP] account_financial_report: Hide accounts with no end balance

SirPyTech:
Add a new flag in the General Ledger wizard that allows the user to hide from the report the accounts that have 0 ending balance.

Same as https://github.com/OCA/account-financial-reporting/pull/1336, but for the General Ledger.

I also rephrased the description of the existing `hide_account_at_0` field because it does not really "Hide account ending balance at 0", and this is probably the root cause of the false expectations.

### DIFF
--- a/account_financial_report/README.rst
+++ b/account_financial_report/README.rst
@@ -49,7 +49,8 @@ currency used in account move lines is properly shown.
 In case that in an account has not been configured a second currency foreign
 currency balances are not available.
 
-In the Trial Balance, the flag *Hide accounts with 0 end balance*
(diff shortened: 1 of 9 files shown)


Marco-ooops (approved):
test funzionale ok

GSLabIt (approved):
Code LGTM

francesco-ooops:
> Same as #1336, but for the General Ledger.

@OCA/accounting-maintainers @sergiocorato wdyt? good for merge? Thanks!

sergiocorato (approved):

PR #1542, opened 2026-07-03
[17.0][ADD] mis_builder_cash_flow_forecast_plan

cgarcia-solvos:
This module extends `mis_builder_cash_flow` so you do not have to create recurring cash flow forecast lines one by one.

It adds:

- Forecast categories: that can be assigned to any cash flow forecast line, so forecast lines can be grouped and filtered (e.g. by rent).
- Cash flow plans: a recurring template that generates a whole series of forecast lines with a single click.
- A per-company, configurable limit on the number of forecast lines a single plan can generate at once, to avoid runaway series over very long date ranges.


pedrobaeza:
Thanks for this, very interesting. I think in this case, you can include the feature in the main module, as if you want, you don't use it, but it's very logical to have it on the main module.

cgarcia-solvos:
> Thanks for this, very interesting. I think in this case, you can include the feature in the main module, as if you want, you don't use it, but it's very logical to have it on the main module.

Thanks!  I've implemented it as an improvement; I'm closing this PR:

https://github.com/OCA/account-financial-reporting/pull/1543

PR #1544, opened 2026-07-15
[19.0][FIX] account_financial_report: trial balance fc key error

Qlasta:
Fixes a KeyError in the Trial Balance report when Show Foreign Currency is enabled and an initial balance exists in a foreign currency. 

### Steps to reproduce

1. Configure a foreign currency.
2. Create a journal entry in that foreign currency with a date before the reporting period.
3. Open the Trial Balance report.
4. Select a reporting period that starts after the journal entry date, so the entry is included in the initial balance.
5. Enable Show Foreign Currency.
6. Generate the report.
7. Observe the following error:

`KeyError: 'amount_currency:sum'`

After this change, the missing key is handled safely and the Trial Balance report is generated successfully with the foreign currency initial balance.

### DIFF
--- a/account_financial_report/report/trial_balance.py
+++ b/account_financial_report/report/trial_balance.py
@@ -275,11 +275,16 @@ class TrialBalanceReport(models.AbstractModel):
                     tb["balance"] if "balance" in tb else tb["balance:sum"]
                 )
                 if foreign_currency:
+                    amount_currency = (
+                        tb["amount_currency"]
+                        if "amount_currency" in tb
+                        else tb["amount_currency:sum"]
+                    )
                     total_amount[acc_id]["initial_currency_balance"] += round(
-                        tb["amount_currency:sum"], 2
+                        amount_currency, 2
                     )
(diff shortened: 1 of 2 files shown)


account_financial_report/report/trial_balance.py
eduezerouali-tecnativa, line 362: Could you fix this too please?

eduezerouali-tecnativa (changes requested):
Hi, thanks for your PR. Tested and code review ok.

account_financial_report/report/trial_balance.py
Qlasta: @eduezerouali-tecnativa good catch, will update.

eduezerouali-tecnativa:
Please could you squash into one commit only as it is same fix and from commit remove [19.0] tag. Thank you. Everything else looks good.

eduezerouali-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1546, opened 2026-07-22
[18.0][FIX] account_financial_report: use 'in' on analytic_account_ids domain

micheledic:
Open the trial balance grouped by analytic account, click an account to see its move lines, and the log gets this on every click:

```
    The domain term "('analytic_account_ids', '=', [166])" should use the
    'in' or 'not in' operator.

```
The report builds that domain as ('analytic_account_ids', '=', id). The id by itself is fine, but the web client wraps it in a list before the search runs, and analytic_account_ids is a many2many, so '=' against a list is what the ORM complains about.

Changed it to 'in'. The many2many branch in expression.py wraps a bare id into a list itself, so the term is valid as-is. Same move lines, warning gone.

### DIFF
--- a/account_financial_report/report/templates/trial_balance.xml
+++ b/account_financial_report/report/templates/trial_balance.xml
@@ -52,7 +52,7 @@
                         <t
                             t-set="aml_domain_extra"
                             t-if="grouped_item['id'] > 0"
-                            t-value="[('analytic_account_ids', '=', grouped_item['id'])]"
+                            t-value="[('analytic_account_ids', 'in', grouped_item['id'])]"
                         />
                         <t
                             t-set="aml_domain_extra"


account_financial_report/report/templates/trial_balance.xml
```diff
@@ -52,7 +52,7 @@
                         <t
                             t-set="aml_domain_extra"
                             t-if="grouped_item['id'] > 0"
-                            t-value="[('analytic_account_ids', '=', grouped_item['id'])]"
+                            t-value="[('analytic_account_ids', 'in', [grouped_item['id']])]"
```
pedrobaeza: Is the wrapping between brackets needed?

micheledic: i checked. it is not strictly needed because using "in" the python code adds the brackets if they are missing . i removed them

pedrobaeza (approved):

ryanduguid (commented):
Thanks, this matches what I see in the ORM. `normalize_leaf` only rewrites `=` when
the value is a list, which is why the client wrapping the id triggers the warning, and
the many2many branch wraps a bare int itself, so the term is valid with or without the
client's list.

Code review only.

One small thing: the description says the id is now passed in a list, but the diff
passes it bare. The code is fine either way, so this is only about the description
matching what merged.

Worth noting for anyone porting: the `t-else` a few lines down keeps
`('analytic_account_ids', '=', False)` and should stay that way, since the warning does
not apply to a bool. 17.0 and 19.0 both still have the same two occurrences, so they
will want this too.

micheledic:
> Thanks, this matches what I see in the ORM. `normalize_leaf` only rewrites `=` when the value is a list, which is why the client wrapping the id triggers the warning, and the many2many branch wraps a bare int itself, so the term is valid with or without the client's list.
> 
> Code review only.
> 
> One small thing: the description says the id is now passed in a list, but the diff passes it bare. The code is fine either way, so this is only about the description matching what merged.
> 
> Worth noting for anyone porting: the `t-else` a few lines down keeps `('analytic_account_ids', '=', False)` and should stay that way, since the warning does not apply to a bool. 17.0 and 19.0 both still have the same two occurrences, so they will want this too.

Description changed.
After this PR is merged i will port to 17.0 and 19.0 .
@pedrobaeza merge ?

PR #1547, opened 2026-07-23
[18.0][FWP] account_finacial_report: PR #1483

Andrii9090-tecnativa:
Add the option to group by partner for accounts (account.account) where reconcile=True

FWP #1483 

@Tecantiva
@pedrobaeza @eduezerouali-tecnativa can you review please?

### DIFF
--- a/account_financial_report/report/general_ledger.py
+++ b/account_financial_report/report/general_ledger.py
@@ -61,13 +61,16 @@ class GeneralLedgerReport(models.AbstractModel):
             )
         return taxes_data
 
-    def _get_account_type_domain(self, grouped_by):
+    def _get_account_type_domain(self, grouped_by, model_name="account.move.line"):
         """To avoid set all possible types, set in or not in as operator of the types
         we are interested in. In v15 we used the internal_type field (type of
         account.account.type)."""
-        at_op = "in" if grouped_by != "taxes" else "not in"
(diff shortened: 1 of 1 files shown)


account_financial_report/report/general_ledger.py
```diff
@@ -94,8 +97,8 @@ def _get_initial_balances_bs_ml_domain(
         domain += [("date", "<", date_from)]
         accounts = self.env["account.account"].search(accounts_domain)
         domain += [("account_id", "in", accounts.ids)]
-        if acc_prt and grouped_by != "none":
-            domain += self._get_account_type_domain(grouped_by)
+        if acc_prt:
```
pedrobaeza, line 100: Why the part of `grouped_by != "none"` is removed?

Andrii9090-tecnativa: It is not affected because the `_prepare_gen_ld_data_group_none` method does not exist [(Line 228).](https://github.com/OCA/account-financial-reporting/blob/18.0/account_financial_report/report/general_ledger.py#L228) Therefore, when `grouped_by` is set to "none", the code returns the data directly without calling read_group. As a result, the condition is redundant now that accounts with reconcile=True can be grouped.

pedrobaeza (approved):

eduezerouali-tecnativa (approved):

PR #1556, opened 2026-08-10
[19.0][FWP] account_financial_report: PR #1434

Ricardoalso:
FWP https://github.com/OCA/account-financial-reporting/pull/1434

**EDIT**
Was already fixed by https://github.com/OCA/account-financial-reporting/commit/0b390bd002baaef84f8eb2196443add63e47c47a. FWP the code design elements remaining

### DIFF
--- a/account_financial_report/wizard/aged_partner_balance_wizard.py
+++ b/account_financial_report/wizard/aged_partner_balance_wizard.py
@@ -60,7 +60,7 @@ class AgedPartnerBalanceWizard(models.TransientModel):
                 ("reconcile", "=", True),
             ]
             if self.company_id:
-                domain.append(("company_ids", "in", [self.company_id.id]))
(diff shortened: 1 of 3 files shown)


pedrobaeza:
It seems the patch is already here, isn't it?

Ricardoalso:
> It seems the patch is already here, isn't it?

Indeed, I updated the PR description accordingly

pedrobaeza:
@adrip-s73 do you remember why changing the domain from `=` to `in`. Is it because the domain optimizer will change it anyway?

adrip-s73:
@pedrobaeza Yes, exactly. I used "in"  because it is a many2many but with a single ID the domain optimizer makes it equivalent to =

pedrobaeza:
Thanks for the confirmation. Then this is not needed, so closing. @Ricardoalso thanks for the fw-port.

PR #1558, opened 2026-08-14
[19.0][MIG] account_sale_stock_report_non_billed: Migration to 19.0

CarlosRoca13:
cc @Tecnativa TT64177

ping @carlos-lopez-tecnativa @christian-ramos-tecnativa


account_sale_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
```diff
@@ -0,0 +1,150 @@
...
+            "search_view_id": search_view_id,
+            # 'from' is a reserved keyword, so the placeholders can't be passed as
+            # plain keyword arguments without changing the translated term.
+            "name": self.env._(
+                "Non billed moves (%(from)s -> %(to)s)",
+                **{
+                    "from": self.stock_move_non_billed_threshold,
+                    "to": self.date_check,
+                },
+            ),
```
carlos-lopez-tecnativa, line 141:
```suggestion
            "name": self.env._(
                "Non billed moves (%(from)s -> %(to)s)",
                 from=self.stock_move_non_billed_threshold,
                 to=self.date_check,
                },
            ),
```

account_sale_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz_views.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2022 Tecnativa - Carlos Roca
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="account_sale_stock_report_non_billed_wiz_form" model="ir.ui.view">
+        <field name="model">account.sale.stock.report.non.billed.wiz</field>
+        <field name="arch" type="xml">
+            <form string="Non Billed Stock Moves">
```
carlos-lopez-tecnativa:
```suggestion
            <form>
```

account_sale_stock_report_non_billed/models/stock_move.py
```diff
@@ -0,0 +1,255 @@
...
+    date_done = fields.Date(
+        string="Effective Date", compute="_compute_date_done", store=True
+    )
+
+    @api.depends("picking_id.date_done")
+    def _compute_date_done(self):
+        self.date_done = False
+        for move in self:
+            if move.picking_id.date_done:
+                move.date_done = move.picking_id.date_done.date()
```
carlos-lopez-tecnativa: I have some doubts about these fields.

Why are they necessary? `stock.move` already has a `date` field, which is set in [_action_done](https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/addons/stock/models/stock_move.py#L2285) at the same point in time when the picking is validated in [_action_done](https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/addons/stock/models/stock_picking.py#L1274).

Additionally, in a large database, this field could take a significant amount of time to recompute. If you keep this field as a computed field, please consider adding a hook to pre-create and populate this field before Odoo computes it.

The `date_done` field is a `datetime`, so it is stored in UTC. When you take only the date part, imagine a picking is validated at night (for example, at 23:00:00). Depending on the user's timezone, the date could be shifted to the next day. This is a problem in countries with timezones that are behind UTC, for example, in LATAM.


```suggestion
    date_done = fields.Date(
        string="Effective Date", compute="_compute_date_done", store=True
    )

    @api.depends("picking_id.date_done")
    def _compute_date_done(self):
        self.date_done = False
        for move in self:
            if move.picking_id.date_done:
                move.date_done = fields.Datetime.context_timestamp(picking, move.picking_id.date_done).date()
```

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2022 Tecnativa - Carlos Roca
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_move_list" model="ir.ui.view">
```
carlos-lopez-tecnativa: I don't think it is necessary to change the ID.

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
```diff
@@ -0,0 +1,82 @@
...
+        <field name="inherit_id" ref="stock.view_move_tree" />
+        <field name="priority">9999</field>
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='date']" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='location_id']" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </xpath>
```
carlos-lopez-tecnativa, line 16: I suggest not displaying these fields either, as they don't add any value in this context. WDYT?
Image

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
```diff
@@ -0,0 +1,82 @@
...
+                />
+            </xpath>
+            <xpath expr="//field[@name='product_uom_qty']" position="after">
+                <field name="currency_id" column_invisible="1" />
+                <field name="quantity_not_invoiced" />
+                <field
+                    name="price_not_invoiced"
+                    widget='monetary'
+                    options="{'currency_field': 'currency_id'}"
+                />
```
carlos-lopez-tecnativa: Could you add the `sum` attribute?
```suggestion
                <field name="quantity_not_invoiced" sum="Qty to invoice" />
                <field
                    name="price_not_invoiced"
                    widget='monetary'
                    options="{'currency_field': 'currency_id'}"
                    sum="Amount to invoice"
                />
```

carlos-lopez-tecnativa (commented):
LGTM, just a few doubts. Please let me know what you think about my comments.

account_sale_stock_report_non_billed/models/stock_move.py
CarlosRoca13: Good catch on the timezone, fixed. Let me explain why the field is still there.

**Why not just use date?** Because the report reasons in dates, not datetimes:

- The wizard domain filters date_done >= threshold / date_done <= date_check with fields.Date values. Against date (a UTC datetime), date <= '2026-08-14' means <= 2026-08-14 00:00:00, so everything validated during that same day would be dropped. We'd have to build timezone-aware bounds in every place that compares against those dates: the search domain, the filtered() in get_quantity_invoiced(), _compute_not_invoiced_values() and the group by.
- date holds the scheduled date while the move isn't done and changes on every rescheduling, whereas depending on picking_id.date_done restricts the recomputation to the moment the picking is validated.

So the field is really "the validation day, in the user's timezone, as a Date" — which is what the report filters and groups by.

**Timezone**: applied, now using fields.Datetime.context_timestamp(move, move.picking_id.date_done).date() (your snippet referenced picking, which isn't defined in that scope). I also changed the wizard's date_check default from fields.Date.today to fields.Date.context_today, so the default day matches the localized date_done.

**Recomputation on large databases**: added a pre_init_hook that creates and fills the column with plain SQL.

account_sale_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz.py
CarlosRoca13: Changed

account_sale_stock_report_non_billed/wizard/account_sale_stock_report_non_billed_wiz_views.xml
CarlosRoca13: Done

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
CarlosRoca13: Done!

account_sale_stock_report_non_billed/models/account_move_line.py
```diff
@@ -0,0 +1,27 @@
...
+            ) >= date_start
+        return (
+            start
+            and (
+                self.move_id.invoice_date
+                or self.move_id.date
+                or self.create_date.date()
+            )
+            <= date_check
+        )
```
carlos-lopez-tecnativa, line 27: I think you need to use `datetime.context_timestamp` too, as mentioned in https://github.com/OCA/account-financial-reporting/pull/1558#discussion_r3783831471

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
```diff
@@ -0,0 +1,93 @@
...
+            <!-- Stock valuation fields don't add any value in this report -->
+            <xpath expr="//field[@name='value']" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='remaining_qty']" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='remaining_value']" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </xpath>
```
carlos-lopez-tecnativa, line 38: Note that these fields are added by the `stock_account` module, not in this view. I'm not sure whether an explicit dependency on `stock_account` is necessary, or whether you should inherit from the `stock_account.stock_move_view_list` view instead.

account_sale_stock_report_non_billed/models/stock_move.py
carlos-lopez-tecnativa: > Good catch on the timezone, fixed. Let me explain why the field is still there.
> 
> **Why not just use date?** Because the report reasons in dates, not datetimes:
> 
> * The wizard domain filters date_done >= threshold / date_done <= date_check with fields.Date values. Against date (a UTC datetime), date <= '2026-08-14' means <= 2026-08-14 00:00:00, so everything validated during that same day would be dropped. We'd have to build timezone-aware bounds in every place that compares against those dates: the search domain, the filtered() in get_quantity_invoiced(), _compute_not_invoiced_values() and the group by.
> * date holds the scheduled date while the move isn't done and changes on every rescheduling, whereas depending on picking_id.date_done restricts the recomputation to the moment the picking is validated.
> 
> So the field is really "the validation day, in the user's timezone, as a Date" — which is what the report filters and groups by.
> 

OK, I understand the purpose of the new field. However, I don't think this field is strictly necessary if you make the necessary adaptations, such as searching from 00:00 to 23:59, using the minimum and maximum times. But if this approach works for you and behaves as expected, let's go with it.

carlos-lopez-tecnativa (changes requested):

account_sale_stock_report_non_billed/models/account_move_line.py
CarlosRoca13: Done, now using context_timestamp for the create_date fallback.

account_sale_stock_report_non_billed/views/stock_move_non_billed_views.xml
CarlosRoca13: Those fields come from stock_account.stock_move_view_list, which extends stock.view_move_tree, so our primary view already gets them. It was already a transitive dependency via sale_stock; I've added stock_account explicitly to the manifest.

account_sale_stock_report_non_billed/models/stock_move.py
CarlosRoca13: I'd rather keep the field: besides the search domain, the date is compared in the filtered() calls and the wizard check, and it's also used for the "Effective date" group by. Grouping on the raw datetime would split days in UTC rather than the user's timezone, so we'd need a stored date field for that anyway.

carlos-lopez-tecnativa (approved):

PR #1562, opened 2026-08-20
[19.0][FIX] account_financial_report: Don't re-read P&L accounts

pedrobaeza:
Forward-port of #1561

The PR #1504 was adding this part of code as part of the fix, while it's enough with just the other part. This code is provoking that for P&L accounts, the initial balance is computed, while they shouldn't, as they by definition should be 0.

Maybe the misunderstanding was to not find that accounts and think they should be there.

@Tecnativa TT64215


pedrobaeza:
Merging for having the patch in the whole series:

/ocabot merge patch

PR #1564, opened 2026-08-21
[18.0][IMP] account_financial_report: hide initial balance row if not include_initial_balance

eduezerouali-tecnativa:
cc @Tecnativa TT64268
ping @pedrobaeza @Andrii9090-tecnativa 

Accounts with `include_initial_balance = False `have an initial balance of 0 by definition, as P&L accounts are reset automatically each year. However, the 0.0 value is still clickable and applies the expected domain, returning entries that do not match the displayed balance. Let's hide the zero initial balance row if account `include_initial_balance=False`


eduezerouali-tecnativa:
From runboat:
imagen

pedrobaeza (changes requested):
You only have to disable it on accounts with `include_initial_balance = True', not on all!

eduezerouali-tecnativa:
> You only have to disable it on accounts with `include_initial_balance = True', not on all!

Done!. I miss understood it. Now it depends on include_initial_balance.

eduezerouali-tecnativa:
ping @pedrobaeza

pedrobaeza:
Thanks for the changes. They are correct, but now I'm thinking that why showing a row that it's going to always be 0? Let's hide the whole row if the account is include_initial_balance = False

eduezerouali-tecnativa:
@pedrobaeza Done. Now if account is `include_initial_balance = False` that line will not show up.
image

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to 19

PR #1565, opened 2026-08-28
[19.0][FIX] account_financial_report: Guard report enrichment

### LINKED ISSUE #1524: [19.0][account_financial_report] TypeError: Cannot read properties of null (reading 'querySelectorAll') when opening reports
## Module
account_financial_report

## Describe the bug
When opening any financial report (e.g. General Ledger), the report fails to render and the following JavaScript error is thrown:

```
TypeError: Cannot read properties of null (reading 'querySelectorAll')
    at enrich (web.assets_web.min.js)
    at HTMLIFrameElement.<anonymous> (web.assets_web.min.js)
```

The root cause is in `static/src/js/report_action.esm.js`: `useEnrichWithActionLinks(this.iframe)` is called unconditionally for **all** reports in `setup()`, not just for account financial reports. When the iframe's `contentDocument` is still `null` (not yet loaded), `report.esm.js` tries to call `contentDocument.querySelectorAll()` and crashes.

## To Reproduce

**Affected versions**: 19.0

1. Install `account_financial_report` on Odoo 19.0
2. Go to Accounting → Reporting → General Ledger (or any other financial report)
3. Fill in the wizard and click Print / Preview

**Expected behavior**
The report renders correctly without JavaScript errors.

**Additional context**

The fix requires two small changes:

**`static/src/js/report_action.esm.js`** — only register the enrich hook for account financial reports:
```javascript
setup() {
    super.setup(...arguments);
    this.isAccountFinancialReport = this.props.report_name.startsWith(`${MODULE_NAME}.`);
    if (this.isAccountFinancialReport) {
        useEnrichWithActionLinks(this.iframe);
    }
},
```

**`static/src/js/report.esm.js`** — add a null guard before accessing `contentDocument`:
```javascript
if (isIFrame) {
    contentDocument = targetElement.contentDocument;
    if (!contentDocument) return;
    doc = contentDocument;
}
```

ryanduguid:
Fixes #1524

## Summary

- scope the custom action-link enrichment hook to `account_financial_report` reports
- ignore iframe load events until `contentDocument` is available

## Verification

- `SKIP=oca-gen-addon-readme pre-commit run --all-files --show-diff-on-failure --color=always` (passes)
- GitHub CI and Runboat remain the integration checks for the Odoo browser suite / `oca_run_tests`

## Ownership
I reviewed, understand, and take responsibility for every line in this contribution.

### DIFF
--- a/account_financial_report/static/src/js/report.esm.js
+++ b/account_financial_report/static/src/js/report.esm.js
@@ -17,6 +17,9 @@ function enrich(component, targetElement, selector, isIFrame = false) {
     // both for the element and the doc
     if (isIFrame) {
         contentDocument = targetElement.contentDocument;
+        if (!contentDocument) {
+            return;
+        }
         doc = contentDocument;
     }
 
--- a/account_financial_report/static/src/js/report_action.esm.js
+++ b/account_financial_report/static/src/js/report_action.esm.js
@@ -10,7 +10,9 @@ patch(ReportAction.prototype, {
         this.isAccountFinancialReport = this.props.report_name.startsWith(
             `${MODULE_NAME}.`
         );
-        useEnrichWithActionLinks(this.iframe);
+        if (this.isAccountFinancialReport) {
+            useEnrichWithActionLinks(this.iframe);
+        }
     },
 
     export() {


account_financial_report/static/tests/report.test.js
```diff
@@ -0,0 +1,52 @@
+import {expect, test} from "@odoo/hoot";
```
pedrobaeza: Please don't add test that doesn't bring value for the flows.

ryanduguid: Removed the Hoot file and the `web.assets_unit_tests` bundle. The iframe `contentDocument` guard and report-name scoping in the production JS are unchanged.

pedrobaeza:
Please reduce it to one commit.

CarlosRoca13 (approved):
It makes sense

ryanduguid:
@pedrobaeza squashed to a single commit on 5 September and rebased on the current 19.0 tip. All checks are green.

pedrobaeza (approved):
/ocabot merge patch

PR #1569, opened 2026-09-09
[18.0][IMP] account_financial_report: Add column options to the report wizards

carlosdauden:
15.0 and 16.0 have the option to choose which columns show on the reports and to limit the width of some texts, but it never reached 17.0 or 18.0: the forward-port to 18.0 only carried the initial `label_text_limit` commit and left out the ones that added the columns model.

Restores the `account.financial.report.column` model, its view and its data, and turns the wizard's single `label_text_limit` field into a per-column configurable limit. The General Ledger, Open Items and Journal Ledger wizards gain a "Columns" page. Compared to the original, fixes the `foreing_currency_visible` typo in the Journal Ledger header, which misaligned the currency columns, and omits the analytic-tags column since it's not supported in 18.0.

@Tecnativa
@sergio-teruel @carlos-lopez-tecnativa


carlos-lopez-tecnativa (changes requested):
Sorry, but I couldn't test these changes because I got this error when I tried to open the wizard.
Image

carlosdauden:
Changes done

carlos-lopez-tecnativa (commented):
I think we should remove the option to add a new line because it displays columns that are not part of the report model. Alternatively, we could add a domain based on the report model. What do you think?

Image

carlos-lopez-tecnativa (commented):
Additionally, I noticed that these options are applied to the `view` and `PDF`, but they don't work in `Excel`. Is this the expected behavior, or should they also work in `Excel`?

carlosdauden:
Changes done

account_financial_report/view/account_financial_report_column_view.xml
```diff
@@ -0,0 +1,43 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+    <record model="ir.ui.view" id="view_account_financial_report_column_tree">
+        <field name="name">account.financial.report.column.list</field>
+        <field name="model">account.financial.report.column</field>
+        <field name="arch" type="xml">
+            <list editable="bottom" create="false" delete="false">
+                <field name="sequence" widget="handle" />
```
carlos-lopez-tecnativa, line 30: What is the purpose of the sequence field?

Personally, I would remove this field. As a user, I might think that the order in which I arrange the fields determines the order in which they are rendered in the report, but this doesn't seem to be the case. Therefore, I don't see any utility in having this field.

Image

PR #1570, opened 2026-09-09
[16.0][IMP] account_financial_report : read wizard data directly

clb-openfire:
All the wizard data was sent to the user's web browser and then sent back to odoo to generate the report

That is useless transit, complicates the code and can raise and error in some extreme cases, when selecting thousands of accounts and using the html export (URI too large, set to 8K bytes by default on nginx)

@mmequignon

### DIFF
--- a/account_financial_report/report/aged_partner_balance.py
+++ b/account_financial_report/report/aged_partner_balance.py
@@ -3,10 +3,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import operator
-from datetime import date, datetime, timedelta
+from datetime import date, timedelta
 
 from odoo import api, models
-from odoo.tools import float_is_zero
+from odoo.tools import float_is_zero, format_date
 
 
 class AgedPartnerBalanceReport(models.AbstractModel):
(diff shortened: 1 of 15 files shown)


account_financial_report/report/aged_partner_balance.py
```diff
@@ -405,56 +405,45 @@ def _calculate_percent(self, aged_partner_data):
...
-        date_at = data["date_at"]
-        date_at_object = datetime.strptime(date_at, "%Y-%m-%d").date()
-        date_from = data["date_from"]
-        only_posted_moves = data["only_posted_moves"]
-        show_move_line_details = data["show_move_line_details"]
-        aged_partner_configuration = self.env[
-            "account.age.report.configuration"
-        ].browse(data["age_partner_config_id"])
+        wizard = self.env["aged.partner.balance.report.wizard"].browse(wizard_id)
+        only_posted_moves = wizard.target_move == "posted"
```
mmequignon, line 409: This can probably move in the `_get_move_lines_data` with the rest of the arguments, or is there a reason to instantiate it before ?

account_financial_report/wizard/aged_partner_balance_wizard.py
```diff
@@ -131,23 +131,6 @@ def _print_report(self, report_type):
...
-                "company_id": self.company_id.id,
-                "account_ids": self.account_ids.ids,
-                "partner_ids": self.partner_ids.ids,
-                "show_move_line_details": self.show_move_line_details,
-                "account_financial_report_lang": self.env.lang,
-                "age_partner_config_id": self.age_partner_config_id.id,
-            }
-        )
-        return res
-
```
mmequignon, line 150: As discussed yesterday, we cannot delete methods, we don't know what module depends on `account_financial_report`, some module might depend on this.
What if some other module uses one of those values?

mmequignon (changes requested):
I guess a few tests would be nice.

@OCA/accounting-maintainers any insight ?

account_financial_report/report/aged_partner_balance.py
clb-openfire: `only_posted_moves` is also included in the return dict of the function

account_financial_report/wizard/aged_partner_balance_wizard.py
clb-openfire: It's not deleting a method, but removing keys in the returned dict.
I couldn't find modules that would use these data.

PR #1571, opened 2026-09-11
[19.0][FIX]: account_financial_report

adrip-s73:
### Problem

The Trial Balance report could overwrite an account’s amounts when its posted entries were split across multiple currencies.

As a result, debit, credit, balance, ending balance, and foreign-currency totals could reflect only the last processed currency group instead of the complete account activity.

### Solution

The report now accumulates values for each account instead of reinitializing them when multiple currency groups are processed.

Currency-specific data remains separated by group, while the account totals are correctly accumulated in the company currency.

### Changes proposed in this PR

- Accumulate debit, credit, balance, and ending balance values per account.
- Preserve foreign-currency amounts for each currency group.
- Prevent account totals and grouped data from being overwritten during iteration.
- Add a regression test for the same account containing entries in company currency and a foreign currency.
- Verify that total debit and credit amounts remain balanced.

### Technical changes

- Updated `_compute_account_amount` in `account_financial_report/report/trial_balance.py`.
- Added foreign-currency test helpers in `account_financial_report/tests/test_trial_balance.py`.
- Added regression coverage for multi-currency account aggregation.
- Added assertions for company-currency totals, foreign-currency amounts, and debit/credit reconciliation.


Reyes4711-S73 (approved):
LGTM

sergioM-S73 (approved):

pedrobaeza (changes requested):
You should act over `_prepare_total_amount` to not create a new one if exist instead of incorporating that `if`. Anyway, please check previous branches if they have that problem, as it sounds to me that we already tackled this.

adrip-s73:
@pedrobaeza done

pedrobaeza (commented):
Please name the comment correctly, and the test generated by AI is too synthetic. This should check real flows, not direct calls or patches over single methods.

adrip-s73:
@pedrobaeza done

account_financial_report/tests/test_trial_balance.py
```diff
@@ -841,3 +841,41 @@ def test_07_foreign_currency_initial_balance(self):
         total = res_data["total_amount"][self.account100.id]
         self.assertEqual(total["initial_currency_balance"], 2000)
         self.assertEqual(total["ending_currency_balance"], 3000)
+
+    def test_08_account_multicurrency_period_totals(self):
+        """Period totals must include every currency group of an account."""
+        company_currency = self.env.company.currency_id
+
```
pedrobaeza: Please don't add empty lines inside methods.

adrip-s73: done

pedrobaeza (approved):
The test is still calling internal methods like `_get_report_lines`, but I can't blame you, as other tests already do that.

Let's move on and thanks for the fix.

/ocabot merge patch

PR #1573, opened 2026-09-15
[FIX] account_financial_report: repair corrupted Turkish translation

volkantasci:
**Description of the issue**

The Turkish translation of `account_financial_report` is corrupted: a non-ASCII character has been replaced with `??`:

- `KDV Raporu Se??enekleri` instead of `KDV Raporu Seçenekleri` (VAT Report wizard)

As a consequence, the string is displayed with broken characters in the Turkish user interface.

**What this PR does**

Restores the missing Turkish character in `i18n/tr.po` (also in an obsolete entry). It only changes translation strings; no code or functional behavior is modified.

**Possible root cause**

This looks like an encoding issue in the tooling that generated the file (Weblate). It might be worth fixing the corresponding entry on Weblate as well, so the correction is not lost on the next automatic synchronization.

Thanks in advance for your review.


volkantasci:
Closing this in favour of fixing the strings on Weblate (OCA's translation workflow).

I checked where the corruption actually comes from: the Turkish file downloaded from Weblate itself for this component still contains `??` lines. The corrupted text is stored in Weblate and pushed into this repository, so a direct `.po` change here would be overwritten by the next Weblate sync.

The 19.0 component to fix is https://translation.odoo-community.org/projects/account-financial-reporting-19-0/account-financial-reporting-19-0-account_financial_report/ — I will correct the Turkish strings there so the fix lands through Weblate.

PR #1574, opened 2026-09-17
[19.0][IMP] account_financial_report: daterange widget on wizards.

adasatorres-tecnativa:
@Tecnativa TT64402
@pedrobaeza 

Date range widgets are added to all wizards that use the `date_from` and `date_to` fields.


pedrobaeza (commented):
Tested and working as expected. You can put everything into a single commit, as it belongs to the same module and even to the same logical block.

